PR #424, opened 2018-03-23
NSSL mp updates to interface with Thompson's climatological aerosol set-up

kkeene44:
TYPE: enhancement

KEYWORDS: nssl, aerosol, aero_opt, thompson

SOURCE: Ted Mansell, National Severe Storms Laboratory

DESCRIPTION OF CHANGES: Modifications to code for interfacing the NSSL microphysics to Greg Thompson’s climatological aerosol setup. The aerosols are turned on with a namelist parameter “aero_opt” which could be used for other schemes, as well.  Scaling for NSSL is introduced because it treats nucleation differently from Thompson (activates them more easily). The microphysics code itself primarily has changes for aero aware (basically a new init subroutine called by physics_init). README.namelist is also updated to explain how this works. There is also commentary at the top of the microphysics code.

LIST OF MODIFIED FILES: 
M    dyn_em/module_initialize_real.F
M    main/real_em.F
M    phys/module_microphysics_driver.F
M    phys/module_mp_nssl_2mom.F
M    phys/module_physics_init.F
M    run/README.namelist

TESTS CONDUCTED: Verified code compiles. WTF pending.
**had to make a couple small mods to module_mp_nssl_2mom.F due to compiling errors. Waiting on confirmation from Ted that those are okay.

### DIFF
--- a/dyn_em/module_initialize_real.F
+++ b/dyn_em/module_initialize_real.F
@@ -166,7 +166,10 @@ integer::oops1,oops2
 
       INTEGER :: j_save
 
+
       LOGICAL :: wif_upside_down = .FALSE.
+
+      REAL :: qnwfa_scale_fac  ! scale factor for water aerosols when using NSSL scheme. Set to value of nssl_aero_scale
 
       !  Dimension information stored in grid data structure.
 
@@ -181,6 +184,8 @@ integer::oops1,oops2
                                 ipsy, ipey, jpsy, jpey, kpsy, kpey )
       its = ips ; ite = ipe ; jts = jps ; jte = jpe ; kts = kps ; kte = kpe
 
+      
+      qnwfa_scale_fac = config_flags%nssl_aero_scale
       CALL model_to_grid_config_rec ( grid%id , model_config_rec , config_flags )
 
       !  Would the user prefer to forego the use of the level of max winds, or the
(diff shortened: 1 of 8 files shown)


dyn_em/module_initialize_real.F
```diff
@@ -181,6 +184,8 @@ SUBROUTINE init_domain_rk ( grid &
                                 ipsy, ipey, jpsy, jpey, kpsy, kpey )
       its = ips ; ite = ipe ; jts = jps ; jte = jpe ; kts = kps ; kte = kpe
 
+      
+      qnwfa_scale_fac = config_flags%nssl_aero_scale
```
davegill: @dudhia @weiwangncar @kkeene44 @MicroTed 
Folks,
Again, the scenario that someone runs the real program with Thompson and then later changes the namelist.input file to NSSL scheme when running the model, which is discussed in a previous comment. Maybe we add in a input_wrf.F and just stop if there is a problem?

We would need to have the qn[iw]fa_scale_fac be metadata from real to WRF. Then we compare the MP scheme (namelist) and the qn[iw]fa_scale_fac (metadata) to see if they are consistent. This is similar to what we do with grid distance, for example: you can't change it between real and WRF.

MicroTed: Oh, but I think I didn't quite understand your point at first, i.e., if somebody runs real.exe with Thompson aero and then changes mp_physics to NSSL but doesn't rerun real.exe. In my current code, that would result in qnn being uninitialized, so that field could be tested for being all zero. If that test fails then one could abort?

davegill: @MicroTed 
A cleaner way that looking at a bunch of zeros is for us to have some metadata that goes between the real program and the WRF model, in just the same way that we ensure that the grid distance has not changed. Here's an idea for input_wrf.F (this is inside of an ARW only ifdef):
```
 453        !  Test to make sure that the MP aerosol option is consistent.
 454 
 455        IF ( switch .EQ. input_only ) THEN
 456           CALL wrf_get_dom_ti_integer ( fid , 'AERO_OPT' ,  aero_opt_compare , 1 , icnt , ierr )
 457           IF ( aero_opt_compare .NE. config_flags%aero_opt ) THEN
 458              WRITE(wrf_err_message,*)'aero_opt from input file  ',aero_opt_compare
 459              CALL wrf_message(wrf_err_message)
 460              WRITE(wrf_err_message,*)'aero_opt from namelist    ',config_flags%aero_opt
 461              CALL wrf_message(wrf_err_message)
 462              CALL wrf_debug ( 0, '---- ERROR: aero_opt does not match, comparing namelist to the input file' )
 463              count_fatal_error = count_fatal_error + 1
 464           END IF 
 465        END IF
```

Some code would need to get into output_wrf.F to actually write out the metadata. This is just a code snippet.

davegill:

PR #489, opened 2018-05-04
New soil table 

weiwangncar:
TYPE: no impact

KEYWORDS: soil parameter table

SOURCE: Patrick Campbell (EPA)

DESCRIPTION OF CHANGES: 
The proposed table value changes are based recent work of Kishne et al. 2017. EPA shows that the change helps to improve their CMAQ results. Full explanation can be found in the attached pdf file.

Kishne, A. Sz., Y. T. Yimam, C. L.S. Morgan, and B. C. Dornblaser, 2017: Evaluation and improvement of the default soil hydraulic parameters for the Noah Land Surface Model. J. Geoderma, 247 - 259. 
http://dx.doi.org/10.1016/j.geoderma.2016.09.022

[Noah_SOILPARM_Kishne_et_al.2017.pdf](https://github.com/wrf-model/WRF/files/1973030/Noah_SOILPARM_Kishne_et_al.2017.pdf)

LIST OF MODIFIED FILES: 
A       run/SOILPARM.TBL_Kishne_2017

TESTS CONDUCTED: 
No regtest needed.


davegill:
@weiwangncar @dudhia 
Folks,
What are we doing with this soil table file? Are we expecting users to rename the file?

dudhia:
Maybe we should consider Mike's suggestion of appending it under another
name in the
existing file. This avoids proliferation of files. Currently we have a
special section for RUC already.
Jimy

davegill:
@dudhia @barlage @weiwangncar 
Folks,
We are not doing structural changes now for v4.1. So, does this file then go in "as-is"? We already have quite a number of these files that users need to rename themselves, so this is really no different.

dudhia:
I think just adding that file is fine and having users rename it to use it.

PR #528, opened 2018-05-16
Example of how to add 4d arrays into ARW

davegill:
## DO NOT COMMIT, EXAMPLE ONLY ##

TYPE: no impact

KEYWORDS: 4d arrays

SOURCE: internal

DESCRIPTION OF CHANGES:
These mods are directed towards a user who would like to have a new 4d array that has the following qualifications:
1. Not in the lateral BC file
2. No tendency arrays auto-created
3. Not required for output
4. Physics only, column oriented, no comms, no advection or diffusion
5. Initialized either in real or WPS
6. No nesting requirement with the field
7. No restart requirements

The new registry file has the following:
1. The 4d array container and the new constituent 3d arrays listed.
2. Namelist option to activate the fields
3. Package for the allocation of the new 4d arrays

At the tail end of the real program, is an example location of where the initialization code could be located (if the information is not coming in from WPS).

Commented out is the included new registry file in the EM shared collection.

LIST OF MODIFIED FILES:
M	   Registry/registry.em_shared_collection
A	   Registry/registry.test_4d
M	   dyn_em/module_initialize_real.F

TESTS CONDUCTED:
1. Code builds
2. New fields are in the WRF input file.

New 3d constituent SINX of TEST4D
screen shot 2018-05-16 at 11 30 13 am

New 3d constituent COSY of TEST4D
screen shot 2018-05-16 at 11 29 58 am

New 3d constituent TANXY of TEST4D
screen shot 2018-05-16 at 11 29 43 am

### DIFF
--- a/Registry/registry.em_shared_collection
+++ b/Registry/registry.em_shared_collection
@@ -25,3 +25,4 @@ include registry.bdy_perturb
 include registry.hyb_coord
 include registry.new3d_wif
 include registry.trad_fields
+#include registry.test_4d
--- a/Registry/registry.test_4d
+++ b/Registry/registry.test_4d
@@ -5,7 +5,6 @@ state  real    -     ikjf   test4d   1   -  -  -
 
 #	Here are the multiple 3d arrays that will be part of the 4d array
 
-#ikjftb works
 state  real  sinx    ikjf   test4d   1   -  i "TESTSINX"   "Test pattern sin(x)"   "unitless"
 state  real  cosy    ikjf   test4d   1   -  i "TESTCOSY"   "Test pattern cos(y)"   "unitless"
 state  real  tanxy   ikjf   test4d   1   -  i "TESTTANXY"  "Test pattern tan(xy)"  "unitless"
--- a/Registry/registry.test_4d
+++ b/Registry/registry.test_4d
@@ -1,0 +1,20 @@
+#	Test 4d array, no boundary conditions, not time tendencies, not output, physics so no nesting, no restart
+
+#	Always start with an empty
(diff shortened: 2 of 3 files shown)


kkeene44:
@davegill @smileMchen 
Does this need to stay in here as a PR, or was the intention to eventually move it elsewhere?

davegill:
@kkeene44 @smileMchen 
Kelly,
I am not sure where to put this. The github diffing makes it easy for users to see what they are supposed to do. We can close it, and just come back to it when we want. I'd be OK with that.

kkeene44:
@davegill @smileMchen 
Is this something we could put on the forum, as an FAQ-type thing? I can attach files, etc. there if needed.

weiwangncar:
That (FAQ on Forum) seems to be a good place for something like this.

On Dec 12, 2018, at 9:07 AM, Kelly Werner <notifications@github.com> wrote:

@davegill <https://github.com/davegill> @smileMchen
<https://github.com/smileMchen>
Is this something we could put on the forum, as an FAQ-type thing? I can
attach files, etc. there if needed.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/wrf-model/WRF/pull/528#issuecomment-446642758>, or mute
the thread

davegill:
Folks,
Maybe on the forum just have a link to the closed PR? I am OK with this or
putting all of the info on the forum.
Dave

smileMchen:
I think it is better that the forum has a link to the PR. In this way people will get a full picture what they should do no matter what version of WRF they are working with.

PR #605, opened 2018-07-31
bug fix: incorrect PSISAT sign change introduced with groundwater fixes

barlage:
bug fix: incorrect PSISAT sign change introduced with groundwater fixes

TYPE: bug fix

KEYWORDS: Noah-MP

DESCRIPTION OF CHANGES:

PSISAT sign change incorrectly introduced in Noah-MP subroutine noahmp_init (PR #287; commit 45d1945c). This sign change is already taken into account in the equation where it is used and should not have been added. The sign change added to subroutine groundwater_init in PR #287 is correct.

Only affects the initialization of frozen soil but can have large effect on soil temperature and moisture simulations when soil temperature is below freezing.
    
LIST OF MODIFIED FILES: 

TESTS CONDUCTED:


weiwangncar:
I'm ok with this PR.

PR #607, opened 2018-08-09
updated thompson scheme to unify with mpas

kkeene44:
TYPE: enhancement

KEYWORDS: thompson, mp_physics, wrf, mpas, unification, shared, physics

SOURCE: internal

DESCRIPTION OF CHANGES: Made modifications to the file module_mp_thompson.F to ensure that it can be used for both WRF and MPAS. The MPAS file was last modified to V3.8 of WRF, so I brought everything up to date with the top of the WRF repository. 

LIST OF MODIFIED FILES: 
M    phys/module_mp_thompson.F

TESTS CONDUCTED: Verified that WRF builds correctly and that output is identical before and after the modifications. Regression test pending.


kkeene44:
waiting on Greg to make modifications to file, determine what should be used for both wrf/mpas, and to add updates he has made since V4.0.

PR #613, opened 2018-09-05
Add a flag that turns on/off soil moisture adjustment

tanyasmirnova:
TYPE: enhancement

KEYWORDS: soil moisture, initialization, RUC LSM

SOURCE: Tanya Smirnova (GSD/ESRL/NOAA)

DESCRIPTION OF CHANGES: 
Issue: This enhancement was motivated by the WRF user's request who initializes RUC soil moisture from the ECMWF soil moisture. The soil moisture adjustment code uses specific Noah's vertical structure, and therefore does not work for ECMWF initialization. Plus, the user cycles soil moisture and does not need any adjustment.

Solution: Flag_sm_adj is added to turn on/off soil moisture adjustment when RUC LSM. The default value is 1 = do soil moisture adjustment.

LIST OF MODIFIED FILES:
M       share/module_soil_pre.F

TESTS CONDUCTED: 
Tested in RAP model at GSD/ESRL

### DIFF
--- a/Registry/Registry.EM_COMMON
+++ b/Registry/Registry.EM_COMMON
@@ -2178,7 +2178,7 @@ rconfig   integer mosaic_cat              namelist,physics	1             3      
 rconfig   integer mosaic_cat_soil         derived         	1            12       rh     "mosaic_cat_soil"          "should be the number of soil layers times the mosaic_cat"      ""  
 rconfig   integer mosaic_lu               namelist,physics      1             0       irh    "mosaic_lu"             ""      ""
 rconfig   integer mosaic_soil             namelist,physics      1             0       irh    "mosaic_soil"           ""      ""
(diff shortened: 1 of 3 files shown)


davegill:
@tanyasmirnova @weiwangncar @dudhia 
Tanya,
Your commit message says "Flag_sm_adj is added to turn on/off soil moisture adjustment when RUC LSM is initialized from the Noah LSM."
Is this switch for use only when RUC is initialized by Noah LSM? The issue you raised is when a user was initializing RUC with ECMWF soil moisture.

tanyasmirnova:
Dave,
It can accommodate both goals. It gives the opportunity to turn off soil moisture adjustment when RUC is initialized not from Noah LSM, but also when the user wants to turn it off completely even when initialized from the Noah. My personal experience with RAP/HRRR physics package shows that soil moisture adjustment reduces cold/moist bias in the cold-start runs initialized from GFS. But I did not test all possible physics options, and can't guarantee that, for example, with the use of YSU PBL the result will be the same.
Thanks,
Tanya

weiwangncar:
Just to be clearer, when Tanya says Noah LSM, it generally means for data coming from NAM or GFS, which use a LSM very much like Noah.

PR #621, opened 2018-09-10
Remove repeated factor of (1+Rv/Rd Qv) in start_em, p computation

### LINKED ISSUE #608: incorrect pressure with use_theta_m=1 with v4.0
Several users have pointed out an error in the wrf.exe program, noticeable in the pressure field, related to the usage of the moist potential temperature option (use_theta_m=1). In the WRF/dyn_em/start_em.F file, near line 782, the original code ALWAYS includes the scale factor of 1+Rv/Rd*Qv for the potential temperature. In the case that the incoming potential temperature variable is already moist theta (i.e., the user has set use_theta_m=1, which is now the default), then this is a repeated scaling.

Here is the original code, which is incorrect for use_theta_m=1.
```
 781 #else
 782        qvf = 1.+rvovrd*moist(i,k,j,P_QV)
 783        grid%p(i,k,j)=p1000mb*( (r_d*(t0+grid%t_1(i,k,j))*qvf)/                     &
 784                   (p1000mb*(grid%al(i,k,j)+grid%alb(i,k,j))) )**cpovcv  &
 785                   -grid%pb(i,k,j)
 786 #endif
```

The code should _functionally_ be:
```
 781 #else
            IF(config_flags%use_theta_m == 0 ) THEN
 782           qvf = 1.+rvovrd*moist(i,k,j,P_QV)
            ELSE
               qvf = 1
            END IF
 783        grid%p(i,k,j)=p1000mb*( (r_d*(t0+grid%t_1(i,k,j))*qvf)/                     &
 784                   (p1000mb*(grid%al(i,k,j)+grid%alb(i,k,j))) )**cpovcv  &
 785                   -grid%pb(i,k,j)
 786 #endif
```

davegill:
TYPE: bug fix

KEYWORDS: use_theta_m, moist theta

SOURCE: Several users pointed out problem, solution from Wei Wang

DESCRIPTION OF CHANGES:
In the porting of the theta_m mods from being ONLY inside of the WRF model
to being part of the initialization step in the real and ideal programs,
the start_em file was modified. Part of the computation of the pressure
involves dry_theta * ( 1 + rv/Rd Qv ).

When the incoming data from the real or ideal program has the namelist
option use_theta_m==1, then the factor ( 1 + Rv/Rd Qv ) is already included
in the theta field. An IF test based on the namelist option use_theta_m
tests whether or not to assign the factor qvf as either 1 or ( 1 + Rv/Rd Qv ).

ISSUE:
Fixes #608

LIST OF MODIFIED FILES:
M	dyn_em/start_em.F

TESTS CONDUCTED:

1. Differences at 0h (since the mod is in start_em.F). Compare original (busted)
code with use_theta_m=1 to fixed code. Differences shown are: 

LEFT: t=0, use_theta_m=0 MINUS  use_theta_m=1_FIXED.

RIGHT: t=0, use_theta_m=0 MINUS  use_theta_m=1_BUSTED.

The differences are supposed to be very small, for example a few hundredths
of a Pa.
screen shot 2018-09-10 at 9 27 04 am

2. The error is introduced in the initialization routine in WRF. After that initial
p' error (3000 Pa), the imbalance quickly decays. However, the perturbations
are enough to change the larger scale solution, which leads to pressure
differences around 300 Pa. Shown is the differences of the p' field at 3h and 12h,
where the differences are wrfout_FIXED - wrfout_BUSTED.
screen shot 2018-09-10 at 10 05 36 am

3. Regression test on cheyenne.
- [x] Reggie looks OK 

RELEASE NOTE:
This bug was introduced in v4.0. This is a critical modification because 
the option use_theta_m=1 is the default, and the results are unphysical.

### DIFF
--- a/dyn_em/start_em.F
+++ b/dyn_em/start_em.F
@@ -773,13 +773,21 @@
     DO i=its,min(ite,ide-1)
 #if ( WRFPLUS == 1 )
        IF ( .NOT. config_flags%var4d_run ) THEN
-          qvf = 1.+rvovrd*moist(i,k,j,P_QV)
+          IF ( config_flags%use_theta_m == 0 ) THEN
+             qvf = 1.+rvovrd*moist(i,k,j,P_QV)
+          ELSE
+             qvf = 1
+          END IF
           grid%p(i,k,j)=p1000mb*( (r_d*(t0+grid%t_1(i,k,j))*qvf)/                     &
                      (p1000mb*(grid%al(i,k,j)+grid%alb(i,k,j))) )**cpovcv  &
                      -grid%pb(i,k,j)
        END IF
 #else
-       qvf = 1.+rvovrd*moist(i,k,j,P_QV)
+       IF ( config_flags%use_theta_m == 0 ) THEN
+          qvf = 1.+rvovrd*moist(i,k,j,P_QV)
+       ELSE
+          qvf = 1
+       END IF
        grid%p(i,k,j)=p1000mb*( (r_d*(t0+grid%t_1(i,k,j))*qvf)/                     &
                   (p1000mb*(grid%al(i,k,j)+grid%alb(i,k,j))) )**cpovcv  &
                   -grid%pb(i,k,j)


davegill:
@liujake @jjguerrette 
I asked for your review because I modified lines inside of a WRFPLUS ifdef.

PR #626, opened 2018-09-12
Correct a few wrong namelist options in namelist.input.fire

smileMchen:
TYPE:  Text only

KEYWORDS: WRF-FIRE, namelist

SOURCE: internal

DESCRIPTION OF CHANGES: In the sample namelist.input.fire that comes with the released WRF.tar file, the syntax for some variables is different than the ones in the registry.fire. 

A few namelist options were spelled incorrectly in the sample namelist.input.fire. This PR corrects the spelling. I compared the namelist options with those in registry.fire to make sure they are correctly spelled in the sample namelist.

LIST OF MODIFIED FILES: 
M test/em_real/namelist.input.fire

TESTS CONDUCTED: Regression test is not necessary

### DIFF
--- a/test/em_real/namelist.input.fire
+++ b/test/em_real/namelist.input.fire
@@ -118,12 +118,12 @@
 !
 ! ignition
  fire_num_ignitions = 3,        ! integer, only the first fire_num_ignition used, up to 5 allowed
- fire_ignition_start_long1=-107.293664, ! start points of ignition lines, longitude 
+ fire_ignition_start_lon1=-107.293664, ! start points of ignition lines, longitude 
  fire_ignition_start_lat1 =  39.698696, ! start points of ignition lines,latitude 
- fire_ignition_end_long1 = -107.293664, ! end points of ignition lines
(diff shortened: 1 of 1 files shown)


davegill:
@smileMchen 
Ming,
If these are legitimate bugs, would you put the "bug" label on these? 

1. Go to https://github.com/wrf-model/WRF/pull/626
2. Click on "Labels" (in the right column of the page), and select "bug"

davegill:
@smileMchen 
Ming,
In the testing, would you state that you ran the WRF Fire code with these new options, and that the old options cause the code to fail.

smileMchen:
@davegill  
Dave, a few  namelist options are spelled wrong in the sample namelist.input.fire. This PR corrected the spelling. I didn't run the fire test, but I compared the namelist options with those in registry.fire to make sure they are correctly spelled in the sample namelist.

smileMchen:
@davegill 
Dave, I have no permission to add labels to PR. What changes I should make so that I can label the PR?

PR #633, opened 2018-09-25
bugfix: fix potential divide by zero in stomatal resistance diagnostic

barlage:
bugfix: fix potential divide by zero in stomatal resistance diagnostic

TYPE: bugfix

KEYWORDS: Noah-MP

DESCRIPTION OF CHANGES:

Bug fix introduced in #399 where divide by zero can occur if LAI is zero. Added check in IF statement. This calculation is only for diagnostics so shouldn't change answers, except to prevent model crash.
    
LIST OF MODIFIED FILES: 

TESTS CONDUCTED:


davegill:
This PR was based on develop, not release-v4.0.1. Replaced by PR #646

PR #634, opened 2018-09-25
bugfix: add limit to Noah-MP leaf aerodynamic resistance

barlage:
bugfix: add limit to Noah-MP leaf aerodynamic resistance

TYPE: bugfix

KEYWORDS: Noah-MP

DESCRIPTION OF CHANGES:

Change in Noah-MP subroutine RAGRB to limit leaf aerodynamic resistance (RB). Currently there is no lower limit. This can result is very large canopy exchange coefficients and subsequent canopy fluxes. Can result in model blow-up with high wind speed.
    
LIST OF MODIFIED FILES: 

M       phys/module_sf_noahmplsm.F

TESTS CONDUCTED:


davegill:
This PR is based from develop, not release-v4.0.1. Now using PR #647

PR #636, opened 2018-09-25
identically matching MM5 code with MM5 MPAS PR 

kkeene44:
TYPE: no impact

KEYWORDS: wrf, mpas, unification, mm5

SOURCE: internal

DESCRIPTION OF CHANGES: There were very minor modifications made to this file to make sure it's identical to the modified file submitted as a pull request to the MPAS repository. Mods consist of spacing and including an MPAS comment.

LIST OF MODIFIED FILES: 
M   phys/module_sf_sfclay.F

TESTS CONDUCTED: code compiles and runs, with bit-for-bit results to unmodified code.

### DIFF
--- a/phys/module_sf_sfclay.F
+++ b/phys/module_sf_sfclay.F
@@ -24,9 +24,8 @@ CONTAINS
                      ids,ide, jds,jde, kds,kde,                    &
                      ims,ime, jms,jme, kms,kme,                    &
                      its,ite, jts,jte, kts,kte,                    &
-                     ustm,ck,cka,cd,cda,isftcflx,iz0tlnd,scm_force_flux &
-                    ,dxCell                                        &
-                         )
+                     ustm,ck,cka,cd,cda,                           &
+                     isftcflx,iz0tlnd,scm_force_flux)
 !-------------------------------------------------------------------
       IMPLICIT NONE
 !-------------------------------------------------------------------
@@ -174,8 +173,14 @@ CONTAINS
       REAL,     DIMENSION( ims:ime, jms:jme )                    , &
                 INTENT(INOUT)   ::                                 &
                                                               QGH
+#if defined(mpas)
+      REAL,     INTENT(IN   )               ::   CP,G,ROVCP,R,XLV
 
+      REAL,     DIMENSION( ims:ime, jms:jme )                    , &
+                INTENT(IN   )               ::                 DX
+#else
       REAL,     INTENT(IN   )               ::   CP,G,ROVCP,R,XLV,DX
+#endif
  
       REAL, OPTIONAL, DIMENSION( ims:ime, jms:jme )              , &
                 INTENT(OUT)     ::                  ck,cka,cd,cda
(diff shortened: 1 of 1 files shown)


phys/module_sf_sfclay.F
```diff
@@ -241,14 +241,14 @@ SUBROUTINE SFCLAY(U3D,V3D,T3D,QV3D,P3D,dz8w,                    &
...
-#if defined(mpas) 
+#if defined(mpas)
                 ,isftcflx,iz0tlnd,scm_force_flux,                  &
                 USTM(ims,j),CK(ims,j),CKA(ims,j),                  &
                 CD(ims,j),CDA(ims,j),dxCell(ims,j)                 &
 #elif ( EM_CORE == 1)
                 ,isftcflx,iz0tlnd,scm_force_flux,                  &
                 USTM(ims,j),CK(ims,j),CKA(ims,j),                  &
-                CD(ims,j),CDA(ims,j)                               & 
+                CD(ims,j),CDA(ims,j)                               &
```
weiwangncar: It looks like this could be changed so that the only difference between mpas and EM_CORE is dxCell.

kkeene44: @weiwangncar 
I think that EM_CORE (1) does not equate wrf_model. If that's true, then there are 3 categories to deal with here, and since MPAS and EM_CORE both need all of these arguments, I think I have to do it this way so that they each get all their necessary arguments, unless you'd rather have something like:

#if (defined(mpas) || (EM_CORE == 1))
,isftcflx,iz0tlnd,scm_force_flux,                  &
 USTM(ims,j),CK(ims,j),CKA(ims,j),                  &
 CD(ims,j),CDA(ims,j)                          &
#endif
#if defined(MPAS)
,dxCell(ims, j)
#endif

weiwangncar: @kkeene44 You are right about this. The other way you proposed to do this is more understandable, but it's not necessarily cleaner. Does anyone else have an opinion on this?

kkeene44: @weiwangncar 
The code has now been updated to get around this problem. DX is now a 2d array in the subroutine sfclay1D for both WRF/MPAS (which is what we want in the long run anyway). This allows us to only need a 'defined(mpas)' at the top of the subroutine sfclay. All mods compile and run fine for WRF and MPAS, and I have updated the code in this PR with the additional changes.

phys/module_sf_sfclay.F
```diff
@@ -521,11 +532,8 @@ SUBROUTINE SFCLAY1D(J,UX,VX,T1D,QV1D,P1D,dz8w1d,                &
         VCONV = SQRT(DTHVM)
         endif
 ! Mahrt and Sun low-res correction
-        if(present(dxCell)) then
-           vsgd = 0.32 * (max(dxCell(i)/5000.-1.,0.))**.33
-        else
-           VSGD = 0.32 * (max(dx/5000.-1.,0.))**.33
-        endif
+        VSGD = 0.32 * (max(dx(i)/5000.-1.,0.))**.33
+!MPAS specific end.
```
weiwangncar: @kkeene44 The comment can be removed now.

kkeene44: @weiwangncar 
I removed and updated the code.

PR #637, opened 2018-09-25
WSM6 change of refl_10cm to "optional" to work with mpas

kkeene44:
TYPE: no impact

KEYWORDS: wsm6, mp_physics, wrf, mpas, unification

SOURCE: internal

DESCRIPTION OF CHANGES: Added 'optional' to the variable refl_10cm inside subroutine wsm6, so that it would work with MPAS.

LIST OF MODIFIED FILES: 
M    phys/module_mp_wsm6.F

TESTS CONDUCTED: code compiles and gives bit-for-bit results before and after modifications.

### DIFF
--- a/phys/module_mp_wsm6.F
+++ b/phys/module_mp_wsm6.F
@@ -136,8 +136,8 @@ CONTAINS
                                                           re_ice, &
                                                          re_snow
 !+---+-----------------------------------------------------------------+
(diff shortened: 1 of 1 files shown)


weiwangncar:
I'm ok with this update.

kkeene44:
@wrf-model/shared_physics 
If this is okay to be approved, will someone approve it? Thanks!

PR #632, opened 2018-09-25
bugfix: prevent excessive snow liquid water movement

barlage:
bugfix: prevent excessive snow liquid water movement

TYPE: bugfix

KEYWORDS: Noah-MP

SOURCE: Michael Barlage (NCAR)

DESCRIPTION OF CHANGES:

Change in Noah-MP subroutine SNOWH2O to update snow liquid water as water passes to layers below. Without this fix, more water is allowed to move to layer than pore space available. This results in unrealistic snow density values during melt season.
    
LIST OF MODIFIED FILES: 

M       phys/module_sf_noahmplsm.F

TESTS CONDUCTED:
1. Summer and winter 24-hr case

PR #645, opened 2018-10-01
bugfix: prevent excessive snow liquid water movement 

davegill:
TYPE: bug fix 

KEYWORDS: Noah-MP

DESCRIPTION OF CHANGES:

Change in Noah-MP subroutine SNOWH2O to update snow liquid water as water passes to layers below. Without this fix, more water is allowed to move to layer than pore space available. This results in unrealistic snow density values during melt season.

Mods to release-v4.0.1, not develop. Replaces PR #632 

LIST OF MODIFIED FILES:

M phys/module_sf_noahmplsm.F

TESTS CONDUCTED:

Summer and winter 24-hr case

### DIFF
--- a/phys/module_sf_noahmplsm.F
+++ b/phys/module_sf_noahmplsm.F
@@ -6809,38 +6809,21 @@ ENDIF   ! CROPTYPE == 0
 
 ! Porosity and partial volume
 
-   !KWM Looks to me like loop index / IF test can be simplified.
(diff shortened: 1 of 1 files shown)


weiwangncar:
I'm ok with this PR.

dudhia:
More substantive changes, but we can assume it is OK.

PR #646, opened 2018-10-01
bugfix: fix potential divide by zero in stomatal resistance diagnostic

davegill:
TYPE: bug fix

KEYWORDS: Noah-MP

DESCRIPTION OF CHANGES:

Bug fix introduced in #399 where divide by zero can occur if LAI is zero. Added check in IF statement. This calculation is only for diagnostics so shouldn't change answers, except to prevent model crash.

Mods to release-v4.0.1, not develop. Replaces PR #633 

LIST OF MODIFIED FILES:

M phys/module_sf_noahmpdrv.F

TESTS CONDUCTED:

Summer and winter 24-hr case


dudhia:
Looks OK. Prevents divide by zero.

weiwangncar:
I'm ok with this PR.

PR #653, opened 2018-10-10
Change index 2 to p_qv

smileMchen:
TYPE: bug fix

KEYWORDS: stochastic perturbation, moist tendency

SOURCE: internal (Dave Gill) 

DESCRIPTION OF CHANGES: in the code "module_first_rk_step_part2.F" for stochastic perturbation, the moist tendency term should be "moist_tend(ims,kms,jms,p_qv)" instead of "moist_tend(ims,kms,jms,2)". 

LIST OF MODIFIED FILES: 
M       dyn_em/module_first_rk_step_part2.F

TESTS CONDUCTED: regression test is not done yet.


davegill:
@smileMchen 
A full regression test is not needed.
1. Compile the code.
2. Run the code and verify that the logic is the same as before, and the fields are the same as before.

PR #654, opened 2018-10-10
Change the number of "28" to 'thompsonaero' in solve_em.F

smileMchen:
TYPE: no impact

KEYWORDS:  dust emission and Thompson scheme  

SOURCE:  internal (Dave Gill)

DESCRIPTION OF CHANGES: when  surface dust emissions scheme is  linked to the ice-friend aerosol variable of Thompson  microphysics, in solve_em.F, the line is changed from
```
       IF (config_flags%dust_emis.eq.1 .AND. config_flags%mp_physics.eq.28)  then
```
to
```
       IF (config_flags%dust_emis.eq.1 .AND. config_flags%mp_physics.eq.thompsonaero)  then
```
LIST OF MODIFIED FILES:
M       dyn_em/solve_em.F

TESTS CONDUCTED: regression test is not done

### DIFF
--- a/dyn_em/solve_em.F
+++ b/dyn_em/solve_em.F
@@ -3548,7 +3548,7 @@ BENCH_START(moist_physics_prep_tim)
                                    its, ite, jts, jte,               &
                                    k_start    , k_end               )
 
-       IF (config_flags%dust_emis.eq.1 .AND. config_flags%mp_physics.eq.28)  then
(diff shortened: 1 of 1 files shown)


davegill:
@smileMchen 
Ming,
A full regression test is not required for this.
1. Build the code
2. Put in some prints that the expected values are actually what get printed out.

PR #657, opened 2018-10-10
Bug fix in Noah LSM - divide by zero

smileMchen:
TYPE: bug fix

KEYWORDS: NOAH LSM, divided by zero

SOURCE: internal

DESCRIPTION OF CHANGES: when the leaf area index is derived from static input data (rdlai2d=true), the value could be zero at certain areas. This leads to extremely large, unrealistic HFX and the model eventually crashed.  

LIST OF MODIFIED FILES: 

TESTS CONDUCTED: 
One case has been tested. The case crashed immediately without the fix. The same case can be done successfully  with the fix. 

Reggie passed.

### DIFF
--- a/phys/module_sf_noahdrv.F
+++ b/phys/module_sf_noahdrv.F
@@ -977,7 +977,7 @@ CONTAINS
 #endif
 
           IF (rdlai2d) THEN
-             if(lai(i,j).eq.0.0) lai(i,j) = 0.01 
+             IF (SHDFAC > 0.0 .AND. XLAI <= 0.0) XLAI = 0.0
              xlai = lai(i,j)
           endif
 
@@ -3306,7 +3306,7 @@ SUBROUTINE lsm_mosaic(DZ8W,QV3D,P8W3D,T3D,TSK,                      &
 #endif
       
                 IF (rdlai2d) THEN
-                   if(lai(i,j).eq.0.0) lai(i,j) = 0.01 
+                   IF (SHDFAC > 0.0 .AND. XLAI <= 0.0) XLAI = 0.01
                    xlai = lai(i,j)
                 endif
       
--- a/phys/module_sf_noahdrv.F
+++ b/phys/module_sf_noahdrv.F
@@ -977,7 +977,7 @@ CONTAINS
 #endif
 
           IF (rdlai2d) THEN
-             IF (SHDFAC > 0.0 .AND. XLAI <= 0.0) XLAI = 0.0
+             IF (SHDFAC(I,J) > 0.0 .AND. LAI(I,J) <= 0.0) LAI(I,J) = 0.01
              xlai = lai(i,j)
           endif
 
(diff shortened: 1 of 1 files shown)


weiwangncar:
There should be a similar fix needed for lsm_mosaic around line 3308.

phys/module_sf_noahdrv.F
```diff
@@ -977,6 +977,7 @@ SUBROUTINE lsm(DZ8W,QV3D,P8W3D,T3D,TSK,                      &
 #endif
 
           IF (rdlai2d) THEN
+             if(lai(i,j).eq.0.0) lai(i,j) = 0.01
```
barlage: This is essentially a consistency check on input datasets and could be made more general:

`if(lai(i,j).eq.0.0) lai(i,j) = 0.01`

replaced with

`
if(shdfac == 0.0) lai(i,j) = 0.0
if(shdfac > 0.0 .and. lai(i,j) == 0.0) lai(i,j) = 0.01
`

smileMchen:
@weiwangncar  The code related to mosaic option is also changed. Thanks.
@barlage  LAI cannot be zero, otherwise it will trigger segmentation fault.

barlage:
Requiring LAI to be > 0 when there is no vegetation seems like an indication a different problem/inconsistency in my opinion. But I guess VEGPARM.TBL has a minimum LAI of 0.01 for all land types so having a global min of 0.01 is consistent for rdlai2d=true.

phys/module_sf_noahdrv.F
```diff
@@ -3305,6 +3306,7 @@ SUBROUTINE lsm_mosaic(DZ8W,QV3D,P8W3D,T3D,TSK,                      &
 #endif
       
                 IF (rdlai2d) THEN
+                   if(lai(i,j).eq.0.0) lai(i,j) = 0.01
```
davegill: @barlage @smileMchen 
Mike,
Should your proposed mod
```
if(shdfac == 0.0) lai(i,j) = 0.0
if(shdfac > 0.0 .and. lai(i,j) == 0.0) lai(i,j) = 0.01
```
also be substituted in this second location?

smileMchen: @davegill 
Dave,
LAI cannot be zero, otherwise it will trigger segmentation fault.

weiwangncar:
Ming, was the seg fault problem related to rdlai2d = true?

weiwangncar:
OK, I see that it is. Many years ago, we put in a similar fix in module_sf_noahlsm.F (#836240bc0):

         IF(RDLAI2D) THEN
          IF (SHDFAC > 0.0 .AND. XLAI <= 0.0) XLAI = 0.01
         END IF

Unfortunately it was accidentally removed a few commits later. This seems to suggest that the check is only needed when SHDFAC is greater than 0. If we have a case that fails because of xlai = 0, a slight different fix can be tested easily.

phys/module_sf_noahdrv.F
barlage: My modifications do not need to be included in either place because they are not addressing/solving the current issue of the pull request, apparently. I have no answer for why the model seg faults (in Noah) for LAI=0.

weiwangncar:
In the final PR, we should add a sentence for release purpose.

PR #661, opened 2018-10-11
Modification of README.namelist

smileMchen:
TYPE:  text only

KEYWORDS: README.namelist

SOURCE: internal

DESCRIPTION OF CHANGES: 
Modify README.namelist so that users can copy options directly. Previously the ";" character was used to denote an inline comment. Now, the "!" character is used, which is the default comment character for Fortran namelist files.

LIST OF MODIFIED FILES:
M       run/README.namelist

TESTS CONDUCTED:  Not necessary

### DIFF
--- a/run/README.namelist
+++ b/run/README.namelist
@@ -10,28 +10,28 @@ information on NMM specific settings (http://www.dtcenter.org/wrf-nmm/users)
        be defined for the nests when max_dom > 1.
 
  &time_control
- run_days                            = 0,	; run time in days
- run_hours                           = 0,	; run time in hours
+ run_days                            = 0,	! run time in days
+ run_hours                           = 0,	! run time in hours
                                                   Note: if it is more than 1 day, one may use both run_days and run_hours
                                                   or just run_hours. e.g. if the total run length is 36 hrs, you may
                                                   set run_days = 1, and run_hours = 12, or run_days = 0, and run_hours = 36
- run_minutes                         = 0,	; run time in minutes
- run_seconds                         = 0,	; run time in seconds
- start_year (max_dom)                = 2001,	; four digit year of starting time
- start_month (max_dom)               = 06,	; two digit month of starting time
- start_day (max_dom)                 = 11,	; two digit day of starting time
- start_hour (max_dom)                = 12,	; two digit hour of starting time
- start_minute (max_dom)              = 00,	; two digit minute of starting time
- start_second (max_dom)              = 00,	; two digit second of starting time
+ run_minutes                         = 0,	! run time in minutes
+ run_seconds                         = 0,	! run time in seconds
+ start_year (max_dom)                = 2001,	! four digit year of starting time
+ start_month (max_dom)               = 06,	! two digit month of starting time
+ start_day (max_dom)                 = 11,	! two digit day of starting time
+ start_hour (max_dom)                = 12,	! two digit hour of starting time
+ start_minute (max_dom)              = 00,	! two digit minute of starting time
+ start_second (max_dom)              = 00,	! two digit second of starting time
                                                   Note: the start time is used to name the first wrfout file.
                                                   It also controls the start time for nest domains, and the time to restart
- tstart (max_dom)                    = 00,	; FOR NMM: starting hour of the forecast
- end_year (max_dom)                  = 2001,	; four digit year of ending time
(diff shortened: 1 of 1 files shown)


run/README.namelist
```diff
@@ -161,117 +161,117 @@ For automatic moving nests: requires special input data, and environment variabl
...
-                                                ; ARW real only.
- use_maxw_level                                 ; 0=do not use max wind speed level in vertical interpolation inside
-                                                ; of the ARW real program, 1 = use level
- use_trop_level                                 ; as above, with tropopause level data
- sfcp_to_sfcp                        = .false.  ; Optional method to compute model's surface pressure when incoming
-                                                ; data only has surface pressure and terrain, but not SLP
- smooth_cg_topo	                     = .false.  ; Smooth the outer rows and columns of domain 1's topography w.r.t.
-                                                ; the input data
- use_tavg_for_tsk                    = .false.  ; whether to use diurnally averaged surface temp as skin temp. The 
+ num_metgrid_soil_levels             = 4        ! number of vertical soil levels or layers input
```
davegill: @smileMchen 
Ming,
In this section, would you add the missing "!" characters back in? The original README.namelist had the ";" character.

run/README.namelist
```diff
@@ -161,117 +161,117 @@ For automatic moving nests: requires special input data, and environment variabl
...
- p_top_requested                     = 5000     ; p_top (Pa) to use in the model
- ptsgm                               = 42000.   ; FOR NMM:  defines the pressure interface dividing
-                                                ;           the terrain following portion of the hybrid vertical
-                                                ;           coordinate (p > ptsgm) and the purely
-                                                ;           isobaric portion of the vertical coordinate (p < ptsgm)
- vert_refine_fact                    = 1        ; vertical refinement factor for ndown, not used for concurrent vertical grid refinement
- vert_refine_method (max_dom)        = 0        ; vertical refinement method (new in 3.7)
+ p_top_requested                     = 5000     ! p_top (Pa) to use in the model
+ ptsgm                               = 42000.   ! FOR NMM:  defines the pressure interface dividing
+                                                            the terrain following portion of the hybrid vertical
```
davegill: @smileMchen 
Ming,
Add in the "!" that are missing here, also.

run/README.namelist
```diff
@@ -334,11 +334,11 @@ lateral BC computations.
  interp_method_type                  = 1 ! bi-linear interpolation
                                      = 2 ! SINT, (default)
                                      = 3 ! nearest neighbor - only to be used for 
-                                         ! testing purposes
+                                           testing purposes
```
davegill: @smileMchen 
Ming,
Here and a few line below, it looks like the original file had the correct "!", and now it has been removed.

run/README.namelist
```diff
@@ -375,102 +375,102 @@ Namelist variables for controlling the specified moving nest:
...
- dfi_bckstop_minute                  = 00       ; two-digit minute of stop time for backward DFI integration
- dfi_bckstop_second                  = 00       ; two-digit second of stop time for backward DFI integration
- dfi_fwdstop_year                    = 2004     ; four-digit year of stop time for forward DFI integration
- dfi_fwdstop_month                   = 03       ; two-digit month of stop time for forward DFI integration
- dfi_fwdstop_day                     = 13       ; two-digit month of stop time for forward DFI integration
- dfi_fwdstop_hour                    = 12       ; two-digit month of stop time for forward DFI integration
- dfi_fwdstop_minute                  = 00       ; two-digit month of stop time for forward DFI integration
- dfi_fwdstop_second                  = 00       ; two-digit month of stop time for forward DFI integration
- dfi_radar                           = 0        ; DFI radar da switch
+ dfi_opt                             = 0        ! which DFI option to use (3 is recommended)
```
davegill: @smileMchen 
Ming,
This will be the last comment like this. The original code had ";" and we want that replaced by "!", not a blank

smileMchen:
@davegill 
Dave, I intentionally replace ";" by blank in some places to make the file look neat. And in other places where there are multiple options like mp_physics, I didn't replace ";" by "!" except the first line. This is because one always needs to copy the first line, then  manually change the option. 
Replacing all ":"  by "!" makes the file look full of 'surprise', which I don't like.
Let's talk tomorrow.

davegill:
@smileMchen 
Ming,
Understood. The mods are approved.

PR #663, opened 2018-10-12
Fix DA compilation when RTTOV-v11 is built with HDF5-enabled

jamiebresch:
TYPE: bug fix

KEYWORDS: WRFDA, compile, RTTOV-v11, HDF5

SOURCE: Jamie Bresch (NCAR)

DESCRIPTION OF CHANGES:
1. Add two more libs (hdf5hl_fortran and hdf5_hl) to CONFIGURE_HDF5_LIB_PATH
2. Add one more lib (rttov11_hdf) to RTTOV_LIB when RTTOV-v11 is built with HDF5-enabled.
**Note that this fix is specific to WRFDA with RTTOV-v11 and
should NOT be merged to the develop branch that has different RTTOV-v12 implementation.**

LIST OF MODIFIED FILES:
M       arch/Config.pl
M       compile

TESTS CONDUCTED:
1. DA compiles after the fix.
2. Without the fix, users have to manually edit configure.wrf to add missing libs to LIB_EXTERNAL in order to compile.

### DIFF
--- a/arch/Config.pl
+++ b/arch/Config.pl
@@ -510,7 +510,7 @@ while ( <CONFIGURE_DEFAULTS> )
 	 }
 
     if ( $sw_hdf5_path ) 
-      { $_ =~ s:CONFIGURE_HDF5_LIB_PATH:-L$sw_hdf5_path/lib -lhdf5_fortran -lhdf5 -lm -lz: ;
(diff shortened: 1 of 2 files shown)


davegill:
@jamiebresch 
Jamie,
How do you suggest that we handle a case where part of this PR should not be merged to develop? Should the PR be split? Should we have a special method for handling this scenario?

jamiebresch:
@davegill Sorry that my commit message may be a bit confusing. The two items are a bundle to fix the compilation issue. I was just describing the change made to each file in separate bullet point.

PR #664, opened 2018-10-12
Fix a few 'typos' for 4DVAR JcDF

jamiebresch:
TYPE: bug fix

KEYWORDS: WRFDA, 4DVAR, jcdfi

SOURCE: Jamie Bresch (NCAR)

DESCRIPTION OF CHANGES:
1. When jcdfi_diag==1, one of the packaged variable names should be jcdfi_v, not dfi_v.
This fix does not affect results.

2. Correct the error message in check-a-mundo when jcdfi_use=1 and jcdfi_diag=0.
The message before the fix:
--- ERROR: If jcdfi_use = 1, then jcdfi_io must also = 1 for that domain
FATAL CALLED FROM FILE:  <stdin>  LINE:    1541
--- Change jcdfi_io in namelist.input
(1) jcdfi_io should be jcdfi_diag. There is no jcdfi_io anywhere else in the code.
(2) 'for that domain' should be removed because jcdfi_diag is a single value.

LIST OF MODIFIED FILES:
M       Registry/registry.wrfplus

TESTS CONDUCTED:
1. a 4dvar surface DA case
2. DA regtest 4dvar_small and radar_4dvar_cv7 cases that have jcdfi_use=true produced
identical results.

### DIFF
--- a/Registry/registry.wrfplus
+++ b/Registry/registry.wrfplus
@@ -626,7 +626,7 @@ package   dyn_em_tl     dyn_opt==202                 -             -
 package   dyn_em_ad     dyn_opt==302                 -             -
 
 package   jcdfi_off     jcdfi_diag==0                  -             -
-package   jcdfi_on      jcdfi_diag==1                  -             state:jcdfi_u,dfi_v,jcdfi_t,jcdfi_p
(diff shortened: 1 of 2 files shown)


davegill:
@jamiebresch 
Jamie,
These look good to me

@weiwangncar @liujake @jjguerrette 
Would you please review this PR

weiwangncar:
This looks good to me.

PR #665, opened 2018-10-15
Bug fix in RUC LSM with MYJ PBL. Bug fix in RUC LSM for seasonal variation of LAI with .rdlai2d=.false.

tanyasmirnova:
TYPE: bug fix

KEYWORDS: MYJ PBL with RUC LSM, Leaf Area Index

SOURCE: Tanya Smirnova, NOAA/ESRL/GSD

DESCRIPTION OF CHANGES: 

1. The MYJ and QNSE PBLs use surface water vapor mixing ratio as low boundary condition for moisture and compute it from the latent heat flux, while other PBL schemes use directly latent heat flux as low boundary condition. Therefore, there is a special computation of latent heat flux inside RUC LSM to get correct surface Qv. In original code the check in the surface driver is made on the option of surface layer scheme, and not PBL scheme. Therefore, if MYJ PBL is not used together with MYJ surface layer scheme, there will be inaccuracy in low boundary condition for moisture.

2. The second bug is in the computation of LAI for particular day when LAI parameter is read in from the VEGPARM.TBL. The bug makes the value of LAI smaller than it should be. The situation when .rdlai2d.=T is not affected by the bug.

3. Removed several commented out pieces of code inside RUC LSM.

LIST OF MODIFIED FILES: 
M       phys/module_sf_ruclsm.F
M       dyn_nmm/module_PHYSICS_CALLS.F

TESTS CONDUCTED: 
Test cases and retrospective runs  are performed at GSD in RAP and HRRR.

RELEASE NOTE: Optional, as appropriate. Delete if not used. Included only once for new features requiring several merge cycles. Changes to default behavior are also note worthy.

### DIFF
--- a/dyn_em/module_first_rk_step_part1.F
+++ b/dyn_em/module_first_rk_step_part1.F
@@ -485,6 +485,7 @@ BENCH_START(surf_driver_tim)
      &        ,SFCRUNOFF=grid%sfcrunoff,ACRUNOFF=grid%ACRUNOFF                           &
      &        ,opt_thcnd=config_flags%opt_thcnd                                          &
      &        ,SF_SFCLAY_PHYSICS=config_flags%sf_sfclay_physics                        &
+     &        ,BL_PBL_PHYSICS=config_flags%bl_pbl_physics                              &
      &        ,SF_SURFACE_PHYSICS=config_flags%sf_surface_physics  ,SH2O=grid%sh2o          &
      &        ,SHDMAX=grid%shdmax      ,SHDMIN=grid%shdmin      ,SMOIS=grid%smois        &
      &        ,SMSTAV=grid%smstav      ,SMSTOT=grid%smstot      ,SNOALB=grid%snoalb      &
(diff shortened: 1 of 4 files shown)


smileMchen:
@tanyasmirnova 
Tanya, 
(1) We want the three bug fixes separated. Either there should be three commits in one PR, or three separate PRs.
(2) You should request the pull request from the release-v4.0.2 for bug fixes, not from master.

davegill:
@tanyasmirnova @smileMchen 
Tanya,

1. We are trying to be very insistent on the bug-fix pull requests (PRs). We are relying on people who are not experts in a particular area to be able to read a description and then compare that description to the actual code changes (and importantly, do that comparison as quickly and accurately as possible). When there are multiple bug fixes per commit, it is not easy to understand which PR comment refers to which code change. Single bug-fix PRs also make commenting on particular lines much easier for reviewers.

2. If there is not need to mention these in the release notes, you may delete that section of the code. However, if someone is not using MYJ PBL with the MYJ surface layer, should they know that there is an impact on low-level moisture? Same goes for the LAI mod - if there is a noteworthy reason for a user to update their code base to include your fixes, please mention that in the release notes.

smileMchen:
@tanyasmirnova @davegill 
Tanya
I modified the code dyn_nmm/module_PHYSICS_CALLS.F and phys/module_surface_driver.F to make it work for NMM. Dave has modified the base branch. Please take a look and let us know if you have any concerns.

PR #669, opened 2018-10-18
Bug fix: cray, nonrepresentable value, module_mp_morr_two_moment_aero.F

davegill:
TYPE: bug fix

KEYWORDS: cray, 1.e-39, small, tiny

SOURCE: Alexander Davies (US Naval Academy)

DESCRIPTION OF CHANGES:
In place of a constant that is too small to represent with single precision
on a Cray compiler, explicitly name a small value as a PARAMETER, and
use that value in the comparison. That small value is the smallest representable
value.

Here is a sample test code
```
program foo
  real, parameter :: real_tiny = tiny(1.0)
  print *,'real tiny = ',real_tiny
end
```

Here is the single and double precision build:
```
> gfortran foo.f90
> a.out
 real tiny =    1.17549435E-38
```
```
> gfortran -fdefault-real-8 foo.f90
> a.out
 real tiny =    2.2250738585072014E-308
```

LIST OF MODIFIED FILES: 
M   module_mp_morr_two_moment_aero.F

TESTS CONDUCTED:
1. Without this mod, the code does not compile.
2. With this mod, the code compiles and runs.


cmattocks:
Maybe use built-in intrinsic function tiny() for this?

program test_huge_tiny
   implicit none

   integer     , parameter :: zeroInt = 0
   real        , parameter :: zeroFlt = 0.0
   real(kind=8), parameter :: zeroDbl = 0.0d0

   print *, "Huge (i,f,d): ", huge(zeroInt), huge(zeroFlt), huge(zeroDbl)
   print *, "Tiny (  f,d): ", tiny(zeroFlt), tiny(zeroDbl)
end program test_huge_tiny

davegill:
@cmattocks 
Craig,
Good idea, thanks for the suggestion

cmattocks:
Glad to help, Dave!

You can also do something like this:

real, parameter :: small = tiny(small)

Maybe add this to the global constants module/class?

Then import as needed:

!---------------
! Import modules
!---------------
use :: Constants, only: small

Craig

Issue #631, opened 2018-09-24
Bounds checking: string length shorter than declared size in ideal initialization code


letmaik:
(Originally reported at https://github.com/WRF-CMake/WRF/issues/4)

When compiling WRF with `-fcheck=bounds` then running `ideal.exe` fails with:

```
At line 19288 of file .../inc/nl_config.inc
Fortran runtime error: Actual string length is shorter than the declared one for dummy argument 'mminlu' (4/256)
```

It seems that lots of code is correctly handling strings, e.g.:
https://github.com/wrf-model/WRF/blob/3ced231c6785c6d26615972a955634365e582db4/dyn_em/module_initialize_fire.F#L207-L209

However, there are cases which would trigger the runtime error:
https://github.com/wrf-model/WRF/blob/3ced231c6785c6d26615972a955634365e582db4/dyn_em/module_initialize_ideal.F#L346

A grep for `nl_set_mminlu` reveals a few more.

I would highly recommend to turn all relevant runtime checks on at least during testing so that those issues can be caught early and definitely before new releases.

cmattocks:
Perhaps this is a more robust way to zap (clear) a character string:

string = repeat(' ', len(string))

https://software.intel.com/en-us/fortran-compiler-18.0-developer-guide-and-reference-repeat

Craig

letmaik:
Here are the others I found:

https://github.com/wrf-model/WRF/blob/3ced231c6785c6d26615972a955634365e582db4/dyn_em/module_wps_io_arw.F#L359-L371

https://github.com/wrf-model/WRF/blob/3ced231c6785c6d26615972a955634365e582db4/dyn_nmm/module_si_io_nmm.F#L613-L614

https://github.com/wrf-model/WRF/blob/3ced231c6785c6d26615972a955634365e582db4/dyn_em/module_initialize_heldsuarez.F#L228

stacywalters:
I believe there is, a simple

string = ' '

should do the job.

cmattocks:
Yes, good point - the two forms work the same way.

To prevent bounds overflows, one can do this (truncate input if necessary):

program testStrings
   implicit none
   character(len =  8) :: string
   character(len =  6) :: tooSmall
   character(len = 10) :: tooLarge
   character(len =  8) :: justRight

   tooSmall = "123456"
   call setString(string, tooSmall)
   print *, "String too small: ", string

   tooLarge = "1234567890"
   call setString(string, tooLarge)
[...]

PR #672, opened 2018-10-19
bounds check bug regarding  mminlu string length

kkeene44:
TYPE: bug fix

KEYWORDS: mminlu, bounds check, character, string, length

SOURCE: internal

DESCRIPTION OF CHANGES: 
Problem:
User complained that when they turned on bounds checking, they received runtime errors: 
```
Fortran runtime error: Actual string length is shorter than the declared one for dummy argument 'mminlu'
```
This call to nl_{set,get}_mminlu was treated incorrectly in a few different files - specifically some ideal initialization files. Modified the files in which the user found the issue to correct the problem.

Due to those modifications, for all files that had logic asking whether mminlu* was equal/not-equal to '  ' (i.e., nothing), it was necessary to add to this logic for when sf_surface_physics was equal/not-equal to nolsmscheme (no LSM scheme set in the namelist). Otherwise, the model assumed one was set, and executed calls, etc. that were not supposed to be used.

ISSUE: 
This issue was referenced in and fixes wrf-model/WRF#631 "Bounds checking: string length shorter than declared size in ideal initialization code"

LIST OF MODIFIED FILES: 
M    dyn_em/module_initialize_heldsuarez.F
M    dyn_em/module_initialize_ideal.F
M    dyn_em/module_wps_io_arw.F
M    dyn_nmm/module_si_io_nmm.F
M    phys/module_physics_init.F
M    share/input_wrf.F
M    share/output_wrf.F

TESTS CONDUCTED: 
1. Verified that all real and idealized cases are able to be built and run with modifications (with a configure -D compile).

### DIFF
--- a/dyn_em/module_initialize_ideal.F
+++ b/dyn_em/module_initialize_ideal.F
@@ -30,6 +30,7 @@ MODULE module_initialize_ideal
 #endif
 
 
+
 CONTAINS
 
 
@@ -111,7 +112,7 @@ CONTAINS
 
    LOGICAL :: moisture_init
    LOGICAL :: stretch_grid, dry_sounding
-   character (len=256) :: mminlu2
+   CHARACTER (len=256) :: mminlu_temp 
 
    REAL    :: xa1, xal1,pii,hm1  !  data for intercomparison setup from dale
 !  space for initial jet in b_wave
@@ -308,7 +309,8 @@ CONTAINS
 
    CASE(hill2d_x, quarter_ss, squall2d_x, squall2d_y, grav2d_x)
 
-    CALL nl_set_mminlu(1,'    ')
+    mminlu_temp(1:10) = 'never_used'
+    CALL nl_set_mminlu(1, mminlu_temp)
     CALL nl_set_iswater(1,0)
     CALL nl_set_cen_lat(1,40.)
     CALL nl_set_cen_lon(1,-105.)
@@ -343,7 +345,8 @@ CONTAINS
 
    CASE(b_wave)
 
-    CALL nl_set_mminlu(1,'    ')
+    mminlu_temp(1:10) = 'never_used'
+    CALL nl_set_mminlu(1, mminlu_temp)
     CALL nl_set_iswater(1,0)
     CALL nl_set_cen_lat(1,40.)
     CALL nl_set_cen_lon(1,-105.)
@@ -377,10 +380,9 @@ CONTAINS
    END DO
 
    CASE(convrad)
-    mminlu2 = ' '
-    mminlu2(1:4) = 'USGS'
-    CALL nl_set_mminlu(1, mminlu2)
-!   CALL nl_set_mminlu(1, 'USGS')
+
+    mminlu_temp(1:10) = 'never_used'
+    CALL nl_set_mminlu(1, mminlu_temp)
     CALL nl_set_iswater(1,16)
     CALL nl_set_isice(1,3)
     CALL nl_set_cen_lat(1,20.)
@@ -469,10 +471,8 @@ CONTAINS
                                    its,ite, jts,jte, kts,kte )
    CASE(seabreeze2d_x)
 
-    mminlu2 = ' '
-    mminlu2(1:4) = 'USGS'
-    CALL nl_set_mminlu(1, mminlu2)
-!   CALL nl_set_mminlu(1, 'USGS')
+    mminlu_temp(1:10) = 'never_used'
+    CALL nl_set_mminlu(1, mminlu_temp)
     CALL nl_set_iswater(1,16)
     CALL nl_set_isice(1,3)
     CALL nl_set_cen_lat(1,20.)
@@ -566,7 +566,8 @@ CONTAINS
 
    CASE (les)
 
-    CALL nl_set_mminlu(1, '    ')
+    mminlu_temp(1:10) = 'never_used'
+    CALL nl_set_mminlu(1, mminlu_temp)
     CALL nl_set_iswater(1,0)
     CALL nl_set_cen_lat(1,40.)
     CALL nl_set_cen_lon(1,-105.)
--- a/dyn_em/module_initialize_ideal.F
+++ b/dyn_em/module_initialize_ideal.F
@@ -30,7 +30,6 @@ MODULE module_initialize_ideal
 #endif
 
 
-
 CONTAINS
 
 
--- a/phys/module_physics_init.F
+++ b/phys/module_physics_init.F
@@ -880,8 +880,9 @@ integer myproc
 !jm   CALL nl_get_mminlu( 1, mminlu_loc )
    CALL nl_get_mminlu( id, mminlu_loc )
 #if (NMM_CORE == 1 && HWRF == 1 )
-   IF ( trim(mminlu_loc) .eq. "" ) THEN
+   IF ( ( trim(mminlu_loc) .eq. "" ) .OR. (config_flags%sf_surface_physics .EQ. nolsmscheme) ) THEN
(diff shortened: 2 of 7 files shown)


davegill:
@mgduda 
Michael,
Would you please ask the group with the CMAKE build capability to try out Kelly's PR?

dyn_em/module_initialize_ideal.F
```diff
@@ -377,10 +380,9 @@ SUBROUTINE init_domain_rk ( grid &
    END DO
 
    CASE(convrad)
-    mminlu2 = ' '
-    mminlu2(1:4) = 'USGS'
-    CALL nl_set_mminlu(1, mminlu2)
-!   CALL nl_set_mminlu(1, 'USGS')
+
+    mminlu_temp(1:10) = 'never_used'
+    CALL nl_set_mminlu(1, mminlu_temp)
```
letmaik: Doesn't this change the semantics? The PR description should explain this change.

dyn_em/module_initialize_ideal.F
```diff
@@ -377,10 +380,9 @@ SUBROUTINE init_domain_rk ( grid &
    END DO
 
    CASE(convrad)
-    mminlu2 = ' '
-    mminlu2(1:4) = 'USGS'
-    CALL nl_set_mminlu(1, mminlu2)
-!   CALL nl_set_mminlu(1, 'USGS')
+
+    mminlu_temp(1:10) = 'never_used'
```
davegill: @kkeene44 
Kelly,
This is where the original value was "USGS". Let's change your mod from "never used" to "USGS" for this line.

dyn_em/module_initialize_ideal.F
```diff
@@ -469,10 +471,8 @@ SUBROUTINE init_domain_rk ( grid &
                                    its,ite, jts,jte, kts,kte )
    CASE(seabreeze2d_x)
 
-    mminlu2 = ' '
-    mminlu2(1:4) = 'USGS'
-    CALL nl_set_mminlu(1, mminlu2)
-!   CALL nl_set_mminlu(1, 'USGS')
+    mminlu_temp(1:10) = 'never_used'
```
davegill: @kkeene44 
Kelly,
same here

dyn_em/module_initialize_ideal.F
kkeene44: @davegill 
I modified this, and the one below that you requested.

davegill:
@kkeene44 
Kelly,
Since this impacts quite a few idealized cases, and since you have a script that builds ALL of the idealized cases for the WRF Tutorial, could you plow through those tests and verify that we are getting bit-wise identical results, before vs after?

kkeene44:
@davegill @dudhia 
All idealized cases have been compiled and tested with the modified code. All run correctly now, with some additional mods to files in phys/ and share/. See the modified PR explanation for details.

davegill:
@kkeene44 @mgduda 
Folks,
We can go two ways with Michael's issue of having the end of the 256 length character string uninitialized.

1. We can initialize the entire string to blank:
```
mminlu_temp = ' '
```
before filling the first few characters through something like
```
mminlu_temp(1:4) = 'USGS'
```
or
```
CALL nl_get_mminlu ( 1 , mmminlu_temp )
```
This would be a few more lines of changes to the existing mods.

2. We can make the source code (automatically generated) deal with an input character string of any length. Below is effectively the before vs after:
```
SUBROUTINE nl_set_mminlu ( id_id , mminlu )
  USE module_configure, ONLY : model_config_rec 
  character*256 , INTENT(IN) :: mminlu
  INTEGER id_id
  model_config_rec%mminlu(id_id) = mminlu
  RETURN
END SUBROUTINE nl_set_mminlu
```
SUBROUTINE nl_set_mminlu ( id_id , mminlu )
  USE module_configure, ONLY : model_config_rec 
  character(len=*) , INTENT(IN) :: mminlu
  INTEGER id_id, loop
  model_config_rec%mminlu(id_id)(1:len(mminlu)) = mminlu
  do loop = len(mminlu)+1, len(model_config_rec%mminlu(id_id))
    model_config_rec%mminlu(id_id)(loop:loop) = ' '
  end do
  RETURN
END SUBROUTINE nl_set_mminlu
```
Looking at the source code there are over 200 subroutines to consider that use this same construct (file names mostly).

Opinions?

**Comment by cmattocks** [comment]
Another variation? Warning: untested.

pure subroutine nl_set_mminlu(id_id, mminlu)
   use :: module_configure, only: model_config_rec
   implicit none
   integer           , intent(in) :: id_id
   character(len = *), intent(in) :: mminlu
   integer :: nStr, nSub
   nStr = len(model_config_rec % mminlu(id_id))    ! Length of string
   nSub = len_trim(mminlu)                         ! Length of sub-string
   if (nSub <= nStr) then
      model_config_rec % mminlu(id_id) = ' '                   ! Clear string
      model_config_rec % mminlu(id_id)(1:nSub) = trim(mminlu)  ! Set 1st part
   else
      model_config_rec % mminlu(id_id) = mminlu(1:nStr)        ! Truncate, prevent overflow
   end if
end subroutine nl_set_mminlu

Craig

**Comment by davegill** [comment]
@cmattocks @mgduda @kkeene44 
Craig,
It looks like we are generally suggesting the same thing: initialize the entire module_config_rec% character string to blanks, and then fill in the beginning part of the module_config_rec% character string with the string input to the function, up to the LEN of that string. 
Do you agree?

**Comment by cmattocks** [comment]
Agreed, Dave,

You might want to guard against overflows and refactor the code to make it reusable.

Below is an example (testStrings.f90) that includes trimming (appropriate for the holiday season).

Craig

program testStrings
   implicit none
   character(len = 8)            :: string
   character(len = *), parameter :: tooSmall  = " 1234 "
   character(len = *), parameter :: tooLarge  = " 1234567890 "
   character(len = *), parameter :: justRight = "12345678"

   print *, "String too small: ", tooSmall
   call setString(string, tooSmall)
   print *, "String set to: ", string

   print *, "String too large: ", tooLarge
   call setString(string, tooLarge)
   print *, "String set to: ", string

   print *, "String just right: ", justRight
   call setString(string, justRight)
   print *, "String set to: ", string

contains

   pure subroutine setString(string, value)
      implicit none
      character(len = *), intent(in out) :: string
      character(len = *), intent(in)     :: value
      character(len = :), allocatable    :: temp
      integer :: ns, nv

      ns = len(string)
      nv = len_trim(adjustL(value))

      allocate (character(len = nv) :: temp)
      temp = trim(adjustL(value))

      string = ''
      string = temp(1:min(nv,ns))

      deallocate(temp)
   end subroutine setString

end program testStrings

File: phys/module_physics_init.F [hunk 4]
```diff
@@ -880,8 +880,9 @@ SUBROUTINE phy_init ( id, config_flags, DT, restart, zfull, zhalf,     &
 !jm   CALL nl_get_mminlu( 1, mminlu_loc )
    CALL nl_get_mminlu( id, mminlu_loc )
 #if (NMM_CORE == 1 && HWRF == 1 )
-   IF ( trim(mminlu_loc) .eq. "" ) THEN
+   IF ( ( trim(mminlu_loc) .eq. "" ) .OR. (config_flags%sf_surface_physics .EQ. nolsmscheme) ) THEN
      CALL wrf_message('WARNING:   MMINLU NOT SET, USING USGS')
+module_surface_driver.F:                     ra_sw_physics, mminlu                                        &
```
davegill: This line is an accidental cut and paste?

davegill:
@kkeene44 
Kelly,
Could you recap the status please. If we have run all idealized cases, then I will approve.

mgduda:
@davegill I don't think all of the comments on the PR have been addressed.

davegill:

Issue #617, opened 2018-09-07
MYNN PBL can't be used with SHCU


davegill:
The MYNN PNL has the EDMF (bl_mynn_edmf=1) option on by default. In check_a_mundo when the bl_mynn_edmf=1, even if the user specifically asks for the shallow cumulus option to be activated (sh_cu=1), the  shallow cumulus scheme is turned off.
```
IF ( model_config_rec%bl_mynn_edmf(i) .EQ. MYNN_STEM_EDMF .OR. &
   model_config_rec%bl_mynn_edmf(i) .EQ. MYNN_TEMF_EDMF) THEN
   model_config_rec%shcu_physics(i) = 0  ! maxdom 
   model_config_rec%ishallow = 0         ! not maxdom
END IF
```

This should be FATAL.

davegill:
fixed by "MYNN PBL shcu combo bug #680"

PR #680, opened 2018-10-25
MYNN PBL shcu combo bug

kkeene44:
TYPE: bug fix

KEYWORDS: mynn, bug, shcu, ishallow, bl_mynn_edmf, check_a_mundo, fatal

SOURCE: internal

DESCRIPTION OF CHANGES: 
Problem: 
The MYNN PBL has the EDMF (bl_mynn_edmf=1) option turned on by default. In check_a_mundo when the bl_mynn_edmf=1, (even if the user specifically asks for the shallow cumulus option to be activated (sh_cu=1)), the shallow cumulus scheme is simply turned off. 

Solution:
The check_a_mundo test is modified to give a fatal error when bl_mynn_edmf=1 AND  sh_cu=1 so that a user will not overlook this message in the log file.

ISSUE: 
This addresses the issue referenced in wrf-model/WRF#617 "MYNN PBL can't be used with SHCU"
closes:#617

LIST OF MODIFIED FILES:
M     share/module_check_a_mundo.F

TESTS CONDUCTED: Verified that code builds with modification. Verified that when using the MYNN PBL, and turning on ishallow, shcu_physics, or both, the real.exe or wrf.exe program stops with the intended fatal error message.

RELEASE NOTE: A check was implemented to stop the real or wrf program if bl_mynn_edmf (which is set to 1 by default in the MYNN scheme) is being used in combination with either the ishallow or shcu_physics options turned on. A message informs the user that bl_mynn_edmf cannot be used with the other 2 options and the namelist must be modified.

### DIFF
--- a/share/module_check_a_mundo.F
+++ b/share/module_check_a_mundo.F
@@ -52,7 +52,7 @@
 
       LOGICAL :: exists, vnest
       LOGICAL , EXTERNAL :: wrf_dm_on_monitor
-      INTEGER :: i, j, oops, d1_value
+      INTEGER :: i, j, oops, d1_value, EDMFMAX, SCHUMAX
       LOGICAL :: km_opt_already_done , diff_opt_already_done
       INTEGER :: count_opt
       LOGICAL :: lon_extent_is_global , lat_extent_is_global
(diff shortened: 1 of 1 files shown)


weiwangncar:
Perhaps we could add a few sentences for release purpose.

kkeene44:
@weiwangncar I added some content.

kkeene44:
@davegill @weiwangncar 
I've updated the code, and tested the various combinations again to ensure that it fails when it should, and does not fail when it shouldn't.

share/module_check_a_mundo.F
```diff
@@ -994,18 +994,18 @@ SUBROUTINE check_nml_consistency
...
+         IF ( ( ( EDMFMAX .GT. 0 ) .AND. ( SCHUMAX .GT. 0 ) ) .OR. &
+              ( ( EDMFMAX .GT. 0 ) .AND. ( model_config_rec%ishallow .GT. 0 ) ) ) THEN
+            wrf_err_message = '--- ERROR: bl_mynn_edmf > 0 requires both shcu_physics=0 & ishallow=0' 
+            CALL wrf_message(wrf_err_message)
+            wrf_err_message = 'when using MYNN PBL, by default bl_mynn_edmf is turned on'
+            CALL wrf_message(wrf_err_message)
+            wrf_err_message = 'Modify namelist.input so that shcu_physics nor ishallow is used when bl_mynn_edmf is turned on'
+            CALL wrf_debug ( 0, TRIM( wrf_err_message ) )
+            count_fatal_error = count_fatal_error + 1
          END IF
```
weiwangncar: This will not allow the domain that doesn't use MYNN/EDMF to use another shallow scheme. Is this more restrictive than what we want?

dudhia:
I don't think anyone would want to do that, so it should be OK.

davegill:
@kkeene44 @weiwangncar 
Kelly,
Yep, this logic seems correct. Thanks for the test results, too.

Wei,
I am now OK with this PR, and Jimy seems to be also OK with the IF tests that are in place. You may officially review this.

davegill:
@kkeene44 
Kelly,
Since this is a bug fix, with one commit with a nice commit message, a squash and merge is OK (when approved)

weiwangncar:
The release note says this stops 'real' program. It should be wrf model, correct?

kkeene44:
@weiwangncar 
No, it actually does stop when the real program is run.

dudhia:
Hopefully it also stops when WRF is run in case someone changed their
namelist after real.
Jimy

weiwangncar:
Should we limit it to stop for WRF only? These options do not require rerunning real program.

dudhia:
I think it is fair to stop it at either stage if the namelist is wrong.

weiwangncar:
In that case, I'm ok. The message could be modified to say it would stop in both real and wrf programs.

kkeene44:
@weiwangncar 
Agreed. I modified the message.

PR #682, opened 2018-10-29
Wet scavenging optional for MP schemes when using WRF Chem

stacywalters:
TYPE: enhancement

KEYWORDS: RAINPROD, EVAPPROD, wetscav

SOURCE: internal

DESCRIPTION OF CHANGES:

Package the state variables RAINPROD and EVAPPROD into the 4D variable
wetscav_frcing.  This allows for a memory footprint reduction when chemistry
is enabled but wet scavenging is disabled.

Add logic to control setting rainprod and evapprod variables only when
the chem group namelist variable wetscav_onoff is set to 1; wetscav_onoff
defaults to 0.  This logical "gate" has been added to the following files:

phys/module_mp_morr_two_moment.F
phys/module_mp_morr_two_moment_aero.F
phys/module_mp_nssl_2mom.F
phys/module_mp_thompson.F
phys/module_mp_wsm6.F

This change affects the following microphysics schemes:

MORR_TWO_MOMENT, MORR_TM_AERO
NSSL_2MOM, NSSL_2MOMG, NSSL_2MOMCCN
THOMPSON, THOMPSON_AERO
WSM6SCHEME

LIST OF MODIFIED FILES:

M       Registry/registry.chem
M       share/module_trajectory.F
M       chem/chem_driver.F
M       phys/module_microphysics_driver.F
M       phys/module_mp_morr_two_moment.F
M       phys/module_mp_morr_two_moment_aero.F
M       phys/module_mp_nssl_2mom.F
M       phys/module_mp_wsm6.F
M       dyn_em/em_solve.F

TESTS CONDUCTED:

Stacy Walters compared results from one hour simulations of the modified
and original code.  There was bit for bit agreement for the RAINPROD and
EVAPRPOD variables as well as BFB agreement for key washout chemical species
such as H2O2 and HNO3. The comparison was done for the following microphyscis
schemes: 

MORR_TWO_MOMENT, NSSL_2MOM, THOMPSON, WSM6SCHEME

### DIFF
--- a/Registry/registry.chem
+++ b/Registry/registry.chem
@@ -6,8 +6,6 @@ rconfig   integer    io_form_auxinput12  namelist,time_control   1  2
 
 
 state    real    CLDFRA2         ikj    misc        1         -      h        "CLDFRA2"             "CLOUD FRACTION"
-state    real    RAINPROD        ikj    misc        1         -      h        "RAINPROD"            "TOTAL RAIN PRODUCTION RATE"       "s-1"
-state    real    EVAPPROD        ikj    misc        1         -      h        "EVAPPROD"            "RAIN EVAPORATION RATE"            "s-1"
 state    real    QV_B4MP         ikj    misc        1         -      -        "QV_B4MP"             "WATER VAPOR BEFORE MICROPHYSICS"  "kg kg-1"
 state    real    QC_B4MP         ikj    misc        1         -      -        "QC_B4MP"             "CLOUD WATER BEFORE MICROPHYSICS"  "kg kg-1"
 state    real    QI_B4MP         ikj    misc        1         -      -        "QI_B4MP"             "CLOUD ICE BEFORE MICROPHYSICS"    "kg kg-1"
@@ -42,6 +40,12 @@ state   integer last_chem_time_second    -  misc    -         -      r       "la
 #
 state   integer emissframes      -          -          -         -      -       "emissframes"           ""         ""
 state   integer fireemissframes  -        -          -        -        -      "fireemissframes"       ""         ""
+#
+# Rain, evaporation rates
+#
+state   real     -              ikjf     wetscav_frcing        -         -      -        -                     "Rain,evap rates"                  "s-1"
+state    real    RAINPROD       ikjf     wetscav_frcing        1         -      h        "RAINPROD"            "TOTAL RAIN PRODUCTION RATE"       "s-1"
+state    real    EVAPPROD       ikjf     wetscav_frcing        1         -      h        "EVAPPROD"            "RAIN EVAPORATION RATE"            "s-1"
 #
 # Anthropogenic emissions
 #
@@ -4141,6 +4145,9 @@ package   usechemdiag    chemdiag==1         -                conv_ct:conv_co,co
 
 # package for aerosol srf area diagnostic
 package   has_aero_srf_area_diag    aero_srf_area_diag==1         -                aero_srf_area:sulf_srf_area,oc_srf_area,bc_srf_area
+#
+# package for wetscavenging
+package   wetscav_pl                wetscav_onoff==1              -                wetscav_frcing:rainprod,evapprod
 
 # Lightning NOx tracers
 state   real    lnox_total   ikjf    tracer        1         -     i8rhusdf=(bdy_interp:dt)    "lnox_total"      "Total LNOx tracer"          "ppmv"
--- a/chem/chem_driver.F
+++ b/chem/chem_driver.F
@@ -1606,7 +1606,8 @@ if ( cam_mam_aerosols ) &
               rri,t_phy,moist,p8w,t8w,                                                &
               grid%dx, grid%dy,                                                       &
               p_phy,chem,rho,grid%cldfra,grid%cldfra2,                                &
-              grid%rainprod,grid%evapprod,grid%hno3_col_mdel,                         &
+              wetscav_frcing(ims,kms,jms,p_rainprod),wetscav_frcing(ims,kms,jms,p_evapprod), &
+              grid%hno3_col_mdel,                                                     &
               grid%qlsink,grid%precr,grid%preci,grid%precs,grid%precg,                &
               grid%wdflx, &
               gas_aqfrac, numgas_mam,dz8w,                                            &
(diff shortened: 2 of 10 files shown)


davegill:
@stacywalters 
Stacy,
I checked a few items.
 - [x] This is ARW only, since it it Chem. Only ARW files for the registry and solver are touched.
 - [x] The new arguments for the module_microphysics_driver.F are all inside of an ifdef for WRF Chem.
 - [x] I traced down mods that you put in the Morrison scheme. The original vars + the new flag are optional.
 - [x] The use of the optional flag is setup via local logical that is on the stack, so always _present_.
 - [x] Automatically, the 4d arrays are passed into the ARW solver, so the wetscav_frcing is OK.
 - [x] This is a change to "develop", so we are good there.

I would like an additional test.
Build the code with "configure -D", and run two simulations for just a few time steps each, using a nest. Turn the namelist flag wetscav_onoff = 0  and then =1. Let's be sure that there are no locations where the arrays are not full sized.

In your commit message:
1. Add in a sentence or two about what you would like to see in the release notes about this item. If this is a memory footprint savings, we may bundle that up comment with other efforts in this area.
2. Explicitly mention that the purpose is memory savings for WRF Chem.

Lastly, maybe let's see how folks feel about the names. 
I can live with wetscav_frcing, though you are only saving an "o".

stacywalters:
Dave,

The test with "configure -D" is a no go.  The repo code without any
modifications
is throwing a floating point exception, divide by zero.  I'm very
suspicious about
the exception as the traceback has it happening in the ext_ncd_write_field
routine
(externals/io_netcdf/wrf_io.f line #7489).

Now what?

Stacy

davegill:
@stacywalters @kayeekayee @ravanah 
Stacy,

Excluding the automatically manufactured Fortran files in the frame directory, most of the WRF modeling system includes the options implied by the "configure -D": bounds checks, various floating point checks, uninitialized variables / pointers, etc. The WRF model is able to run through a short simulation with these checks in place (we have to exclude ever using the old Fortran77-esque FFT stuff, which monkeys around with work space, and relies on pushing 3d arrays into 1d arrays through subroutine calls).

This is where the GitHub teams actually have teeth. My guess is that there are packages that are not correctly set up for WRF Chem. [Editor's note:  I have heard from several people independently that recent WRF versions with WRF Chem are significantly slower than older versions, anywhere between 40% to several times. We do not see this in WRF without chemistry.]

Test your namelist (turn off chem) with a "configure -D" build of WRF and see if it can get through a few time steps. You should rebuild WRF. If you are using GNU, use version 6.2 or earlier as the later versions cause rrtmg sw fast to not compile.

stacywalters:
Dave,

I'm working on pinning down the floating point exception.

I would need more details regarding the "significantly slower" WRF with
chemistry.  If these users
are comparing apples to apples in that they are running the same chemistry
with the same
options then this is something that merits further investigation.  If the
user has changed the photolysis
option to TUV then the 40% is entirely possible.

Regardless, if you get performance feedback from users of this nature I
would hope
you would pass that on with some contact information if possible.

Stacy

stacywalters:
Dave,

I just finished configuring with -D, compiling, and running WRF with no
chemistry at all.
And I still get the same floating point exception, divide by zero.
The run directory is:

/gpfs/fs1/scratch/stacy/WRF_tst

the source directory is:

/gpfs/fs1/scratch/stacy/WRF_REPO/WRFV4.0.2

You can see the exception in rsl.error.0000.

I compiled with the "standard" Intel compiler.

Strange thing about this is that the last traceable statement in the trace
is a call to define a variable in wrfout_d01*.  How in the world does a
call to define a variable produce a divide by zero???

Stacy

davegill:
@stacywalters 
Stacy,
Let's start simplifying your system.
Grab a namelist.input, wrfinput_d01, wrfbdy_d01 from
```
/gpfs/fs1/scratch/gill/WTF/Runs/wrf_combo.32/em_real/wrf_regression.namelist.input.01
```
1. Can you get that to work with your executable? If it works, then there is a run-time setting that you have that is causing troubles. If it does not work, then intel is catching a problem that GNU is not.
2. If so, then grab the metgrid files in that dir (Jan 2000, d01 only), add in your physics, and re-run real. If that is OK, then it is still a run-time problem, just not in physics. If it is not OK, then physics is the likely culprit.
3. Since there is a flagged problem in the output, maybe a good place to consider would be all of the I/O that you are including in the &time_control namelist record.

stacywalters:
Dave,

Will give it a try.  May need to come back to you for specifics regarding
item 2.

We have run hundreds of WRF-Chem simulations with multiple inputs specified
in the time_control group of the NML input.  Without the -D option we have
never
run into a problem.  I doubt that this is the cause of the problem.

For a short test I could remove those inputs for the pure WRF case.  What
do
you say?

Stacy

davegill:
@stacywalters 
Stacy,
Maybe add a "/" after your io_form for the boundary file to close off the remaining part of the time_control nml record.

stacywalters:
Dave,

Good idea, but the result is the same, exact same.  I'm going for a swim
and will
attempt some old fashion debugging to zero in a little closer to what is
happening.

Stacy

davegill:
@stacywalters 
Stacy,
This is a longer-term project, since it is a PR for develop (not a bug fix release). We do not have much time pressure on this.

davegill:
@stacywalters 
Stacy,
Yes, modify the commit message (do this on the github page directly, just click the "Edit" button. When you are ready, ask me to do a regression test.

stacywalters:
Dave,

If you want me to do the reg test and can give me a set of short directions
that is fine with me.

Stacy

davegill:
@stacywalters 
Stacy,
Given the way that cheyenne uses the share queue, given the disk system saying that a file does not exist until a minute or so after the program finishes on queued jobs - I will do this myself.

stacywalters:
Dave,

I've added a sentence regarding "memory footprint" at the end of the first
paragraph
below the "DESCRIPTION OF CHANGES" section.  The code is ready for reg
testing.

On another subject, are bug fixes still going into v4.0.2?

Thanks, Stacy.

ravanah:
I approve these changes.

kayeekayee:
Approved.

PR #686, opened 2018-10-30
Fix DA formatted print of random_seed array when put_rand_seed=false

jamiebresch:
TYPE: bug fix

KEYWORDS: WRFDA, random seed array size, da_message

SOURCE: Jamie Bresch (NCAR)

DESCRIPTION OF CHANGES:
When running WRFDA in RANDOMCV mode and put_rand_seed=.false.,
the formatted write of the random number seed array has a preceding count of 10.
Depending on compiler versions, the seed array size can be larger than 10.
In that case, the error message is
```
At line 2938 of file da_define_structures.f
Fortran runtime error: End of file
```
The solution is to build a run-time format string based on the seed_size.

LIST OF MODIFIED FILES:
M       var/da/da_define_structures/da_random_seed.inc

TESTS CONDUCTED:
With gnu/7.2.0, the size of the random_seed array is 33, RANDOMCV ran after the format fix.

RELEASE NOTE: Bug fix for WRFDA RANDOMCV when put_rand_seed=.false. and the compiler has a minimum size of random seed array larger than 10.

### DIFF
--- a/var/da/da_define_structures/da_random_seed.inc
+++ b/var/da/da_define_structures/da_random_seed.inc
@@ -28,6 +28,8 @@ subroutine da_random_seed
 #endif
 
    call random_seed(size=seed_size)              ! Get size of seed array.
+   write(unit=message(1),fmt='(a,i6)') 'Size of the random_seed array is ', seed_size
+   call da_message(message(1:1))
    allocate(seed_array(1:seed_size))
    seed_array(1:seed_size) = 1
 
@@ -77,7 +79,7 @@ subroutine da_random_seed
    else                                 ! Random seed set "randomly"
       call random_seed
       call random_seed(get=seed_array(1:seed_size))
-      write(unit=message(1),fmt='(a,10i16)') 'Random number seed array = ', seed_array
+      write(unit=message(1),fmt='(a,100i16)') 'Random number seed array = ', seed_array
       call da_message(message(1:1))
    end if
    
--- a/var/da/da_define_structures/da_random_seed.inc
+++ b/var/da/da_define_structures/da_random_seed.inc
@@ -14,6 +14,7 @@ subroutine da_random_seed
    integer, allocatable :: seed_array(:)
 
    integer              :: myproc,ierr,i
+   character(len=32)    :: fmtstring
 
    if (trace_use) call da_trace_entry("da_random_seed")
 
@@ -79,7 +80,8 @@ subroutine da_random_seed
    else                                 ! Random seed set "randomly"
       call random_seed
       call random_seed(get=seed_array(1:seed_size))
-      write(unit=message(1),fmt='(a,100i16)') 'Random number seed array = ', seed_array
+      write(fmtstring, '(a,i3,a)') '(a27,', seed_size, '(i16))'
+      write(unit=message(1),fmt=trim(fmtstring)) 'Random number seed array = ', (seed_array(i), i=1, seed_size)
       call da_message(message(1:1))
    end if
    


var/da/da_define_structures/da_random_seed.inc
```diff
@@ -77,7 +79,7 @@ subroutine da_random_seed
    else                                 ! Random seed set "randomly"
       call random_seed
       call random_seed(get=seed_array(1:seed_size))
```
davegill: @jamiebresch 
Jamie,
I agree that 100 should last for a while. However, maybe an if test just before the write statement?
```
IF ( seed_size .LE. 100 ) THEN
   write(unit=message(1),fmt='(a,100i16)') 'Random number seed array = ', seed_array
ELSE
    write(unit=message(1),fmt='(a,1000i16)') 'Random number seed array = ', seed_array
END IF
```
When we started WRF, most of the seed sizes were 2 or 4, now they are 33. No telling where these are headed.

jamiebresch: @davegill Dave, thanks for the idea. I actually wanted to use unlimited repeated format. However, it does not work when writing to unit=message(1).
The following code works though.
`write(unit=stdout,fmt='(a,(10i16))') 'Random number seed array = ', (seed_array(i), i=1, seed_size) `

davegill: @jamiebresch 
Jamie,
I did not see that you were writing to a character string. I assume that is this a "one time" write of the seeds, so performance is not a _huge_ issue. How about a loop with a single seed per line?
```
program foo
   implicit none
   integer :: i, seed_size
   integer, allocatable, dimension(:) :: seed_array
   character(len=256), dimension(3) :: message

   seed_size = 25
   allocate(seed_array(seed_size))
   do i = 1 , seed_size
      seed_array(i) = i
   end do

   write (unit=message(1),fmt='(a)') 'List of Seeds'
   print *,trim(message(1))
   do i = 1 , seed_size
      write (unit=message(1),fmt='(a,i3,a,i16)') 'seed count = ',i,', seed value = ',seed_array(i)
      print *,trim(message(1))
   end do
end
```
This gives
```
 List of Seeds
 seed count =   1, seed value =                1
 seed count =   2, seed value =                2
 seed count =   3, seed value =                3
 seed count =   4, seed value =                4
 seed count =   5, seed value =                5
 seed count =   6, seed value =                6
 seed count =   7, seed value =                7
 seed count =   8, seed value =                8
 seed count =   9, seed value =                9
 seed count =  10, seed value =               10
 seed count =  11, seed value =               11
 seed count =  12, seed value =               12
 seed count =  13, seed value =               13
 seed count =  14, seed value =               14
 seed count =  15, seed value =               15
 seed count =  16, seed value =               16
 seed count =  17, seed value =               17
 seed count =  18, seed value =               18
 seed count =  19, seed value =               19
 seed count =  20, seed value =               20
 seed count =  21, seed value =               21
 seed count =  22, seed value =               22
 seed count =  23, seed value =               23
 seed count =  24, seed value =               24
 seed count =  25, seed value =               25
```

var/da/da_define_structures/da_random_seed.inc
```diff
@@ -77,7 +80,8 @@ subroutine da_random_seed
...
-      write(unit=message(1),fmt='(a,10i16)') 'Random number seed array = ', seed_array
+      write(fmtstring, '(a,i3,a)') '(a27,', seed_size, '(i16))'
+      write(unit=message(1),fmt=trim(fmtstring)) 'Random number seed array = ', (seed_array(i), i=1, seed_size)
       call da_message(message(1:1))
```
jamiebresch: @davegill Dave, as Michael Duda did in setup_physics_suite in module_check_a_mundo.F, building a run-time format string allows writing formatted text for variable array size.

davegill: @jamiebresch @mgduda 
Jamie,
Yep, I like it.

PR #687, opened 2018-10-30
Add time slot information in DA gts_omb_oma diagnostics

jamiebresch:
TYPE: enhancement

KEYWORDS: WRFDA, gts_omb_oma, time_slot

SOURCE: Jamie Bresch (NCAR)

DESCRIPTION OF CHANGES:
For diagnostic purpose, write out additional time slot information in the gts_omb_oma output
so that innovation and residual information can be distinguished and evaluated for each time slot.
Since the additional i8-formatted integer is written out in the end of an existing line,
it should not have impact on users' existing programs that read the ascii gts_omb_oma.

LIST OF MODIFIED FILES:
M       var/da/da_obs_io/da_read_omb_tmp.inc
M       var/da/da_obs_io/da_write_obs.inc

TESTS CONDUCTED:
1. The same NCL script can process both old and new gts_omb_oma files.
2. WRFDA regtests on Cheyenne with intel 17.0.1.

RELEASE NOTE: Enhance the gts_omb_oma diagnostics by appending time slot information after the level information.

### DIFF
--- a/var/da/da_obs_io/da_read_omb_tmp.inc
+++ b/var/da/da_obs_io/da_read_omb_tmp.inc
@@ -33,6 +33,7 @@ subroutine da_read_omb_tmp(filename,unit_in,num,obs_type_in,nc,if_wind_sd)
                    rain_obs, rain_inv, rain_error, rain_inc, zk
    integer     :: u_qc, v_qc, t_qc, p_qc, q_qc, tpw_qc, spd_qc, ref_qc, rain_qc
    integer     :: eph_qc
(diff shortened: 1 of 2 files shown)


davegill:
@jamiebresch @jjguerrette @liujake 
Jamie,
This is diagnostic only. You mention that the location of the new print statements allows existing programs to process both old and new files. And this is helpful to distinguish innovation and residual info. I am OK with this PR. While this is an enhancement, it is a single modification and simple to understand.

I'll let the DA folks do a review.

PR #690, opened 2018-11-06
Provide brief description / units of shared model constants

smileMchen:
TYPE: enhancement

KEYWORDS: constants

SOURCE: internal

DESCRIPTION OF CHANGES: explicitly describe constants in the code module_model_constants.F

LIST OF MODIFIED FILES: 
M         share/module_model_constants.F
TESTS CONDUCTED: No test is necessary. 

RELEASE NOTE: Don't need to mention in release.


davegill:
@smileMchen @kkeene44 
Ming and Kelly,
This is a good starting point for us to begin the process of documenting all of these physical constants. 

This PR would merge into develop, so we can incrementally iterate on this several more times before the v4.1 release. 

All of the mods are comments, so there are no changes to model results. These modifications are safe and we need to eventually include a simple description of each constant.

I am OK with this PR.

PR #692, opened 2018-11-07
Add the option mp_physics=40 to README.namelist

smileMchen:
TYPE:  text only

KEYWORDS: README.namelist

SOURCE: internal

DESCRIPTION OF CHANGES: The option mp_physics=40, which is missing previously, is added to README.namelist.  

LIST OF MODIFIED FILES

M      run/README.namelist

TESTS CONDUCTED: not needed   

RELEASE NOTE:  Don't need to include

### DIFF
--- a/run/README.namelist
+++ b/run/README.namelist
@@ -514,6 +514,7 @@ Namelist variables for controlling the adaptive time step option:
                                      = 30, HUJI (Hebrew University of Jerusalem, Israel) spectral bin microphysics,
                                            fast version
                                      = 32, HUJI spectral bin microphysics, full version
+                                     = 10, Morrison (2 moments) with consideration of of CESM-NCSU RCP4.5 climatological aerosol 
                                      = 50, P3 1-category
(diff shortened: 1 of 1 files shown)


run/README.namelist
```diff
@@ -514,6 +514,7 @@ Namelist variables for controlling the adaptive time step option:
                                      = 30, HUJI (Hebrew University of Jerusalem, Israel) spectral bin microphysics,
                                            fast version
                                      = 32, HUJI spectral bin microphysics, full version
+                                     = 10, Morrison (2 moments) with consideration of of CESM-NCSU RCP4.5 climatological aerosol
```
davegill: @smileMchen 
Ming,
This is supposed to be mp_physics=40, right? You have "10" as the option. Also, there is a repeated "of"

smileMchen: @Dave, yes it should be mp_physics =40. Thanks for pointing out the mistake!

PR #693, opened 2018-11-08
Goddard contributions for Radiation and MP schemes

davegill:
TYPE:  new feature

KEYWORDS: goddard, radiation, microphysics

SOURCE: Toshi Matsui, Carlos Cruz (NASA-GSFC)

DESCRIPTION OF CHANGES:
The modified files, Makefile and depend.common, are modified to handle the new files for the Goddard 3ice and 4ice MP and radiation schemes. The added files contain the Goddard contributions for the radiation and microphysics schemes. Additional information about the Goddard MP and radiation schemes may be found in these recent articles:

Goddard Radiation:

Matsui, T., S. Q. Zhang, W.-K. Tao, S. Lang, C. Ichoku, and C. Peters-Lidard (2018), Impact of Radiation Frequency, Precipitation Radiative Forcing, and Radiation Column Aggregation on Convection-Permitting West African Monsoon Simulations, Climate Dynamics, 1-21, https://doi.org/10.1007/s00382-018-4187-2

4ICE microphysics:

Lang, S., W.-K. Tao, J.-D. Chern, D. Wu, and X. Li, 2014:  Benefits of a 4th ice class in the simulated radar reflectivities of convective systems using a bulk microphysics scheme.  J. Atmos. Sci., 71, 3583-3612. doi: http://dx.doi.org/10.1175/JAS-D-13-0330.1

Tao, W.-K., D. Wu, S. Lang, J.-D. Chern, C. Peters-Lidard, A. Fridlind, and T. Matsui (2016), High-resolution NU-WRF simulations of a deep convective-precipitation system during MC3E: Further improvements and comparisons between Goddard microphysics schemes and observations, J. Geophys. Res. Atmos., 121, 1278-1305, https://doi.org/10.1002/2015JD023986.

This PR replaces "Goddard contributions for radiation and microphysics schemes. #340". That PR was updated to merge into develop (instead of the v3.9 master branch). Additionally from that PR:
1. the nonstandard _isnan()_ function was replaced in several locations with the standard _ieee_is_nan()_ function
2. the new variables are packaged for the mp_physics=nuwrf3icescheme and nuwrf4icescheme
3. new variables are packaged with ra_lw_physics=goddardlwscheme and  ra_sw_physics=goddardswscheme
4. the binary lookup table is pushed around in the top-level Makefile, and removed with clean
5. the "use netcdf" and associated netcdf functionality in the Goddard radiation scheme are ifdef'ed out
6. code builds and runs with NMM, but safely does not use the new Goddard options
7. the 4ice MP scheme takes over the "7" option, 3ice is "27", and the old MP scheme is relegated to "97" 
8. the DFI option is modified for ARW to accommodate the additional effective radii from the Goddard MP schemes: rain, graupel, hail
9. the DFI option for NMM for new Goddard is sufficient to let the NMM code compile
10. configure -r8 builds OK with Goddard

LIST OF MODIFIED FILES:
M       Makefile
M       Registry/Registry.NMM
M       Registry/registry.chem
M       clean
M       dyn_em/depend.dyn_em
M       dyn_em/start_em.F
M       dyn_em/solve_em.F
M       main/depend.common
M       phys/Makefile
M       phys/module_microphysics_driver.F
M       phys/module_ra_goddard.F
A       phys/module_checkerror.F
A       phys/module_gocart_coupling.F
A       phys/module_mp_gsfcgce_3ice_nuwrf.F
A       phys/module_mp_gsfcgce_4ice_nuwrf.F
A       run/BROADBAND_CLOUD_GODDARD.bin
M      run/README.namelist
M      share/dfi.F
M      share/module_check_a_mundo.F

TESTS CONDUCTED:
1. Original version
   * Ran NU-WRF regression tests.
   * Tested 3-ice and 4-ice schemes with and without GOCART chemistry.
   * Regression test OK.
2. New version
   * Compiles ARW
   * Run Jan 2000 case, hail is shown in the 12-h movie loop, data every 30 min
4ice
    * ARW DFI (2 hours back and forth) with 12-h simulation, with configure -D build
    * Builds a and runs non-Goddard WRF Chem
    * Need tests with DA
    * Compiles with NMM, but attempting to use new MP options is caught. This test was conducted before the Goddard MP schemes were re-ordered. Here "55" refers to the 3ice scheme.
```
> grep mp_physics namelist.input
 mp_physics                          = 55,	5,	5,
```
```
> real_nmm.exe
--- ERROR: Chosen microphysics scheme cannot run with WRF-NMM
  --- Fix mp_physics in namelist.input
```

3. Comparison of standard MP + radiation with Goddard MP and radiation, 8 total tests for a 12-h Jan 2000 simulation, single domain. These Goddard MP options refer to the numbering prior to the re-organization of the scheme numbers for Goddard. 
    * MP = 7 (orig Goddard), 8, 55 (3ice), 56 (4ice)
    * RA = 4, 5
screen shot 2018-12-17 at 2 24 28 pm
4. Bit-for-bit identical results are expected and obtained when using the new code vs the original code for both the MP=7+RA=4 and MP=8+RA=4.
5. Approximate timings in seconds for each of the MP schemes, at a non-radiation step, and at the radiation time step for RA=4 and RA=5.  RA=5 is cheaper than RA=4. MP=4ice is more expensive than the others.

|               |  MP=Orig Goddard  |  MP=08  |  MP=3 ice  |  MP=4 ice  |  
| -----      | ------------------- | -------- | ----------- | ----------  |
| Non RA  | 0.026                       | 0.030     | 0.028          | 0.041          |
| RA=4     | 0.54                          | 0.59       | 0.63             | 0.66           |
| RA=5     | 0.45                          | 0.47       | 0.48             | 0.51            |
6. Results from Wei Wang from running mp_physics = 7 + ra_*_physics = 5 over US on a 4 km grid. Plot on the left is from this combination of new physics, and on the right is from mp_physics = 6 and RRTMG radiation. The top two are RAINNC from 12 h forecast, and bottom two are SWDOWN at 12 h forecast time. The microphysics appears to have stronger impact on cloud and hence downward shortwave radiation.
rainnc_12h
swdown_12h
7. Results from Wei Wang: Here are a few plots showing the radiative impact from 4ice (left) and 3ice (right) schemes, both run with Goddard radiation. The top panel is for SWDOWN, and bottom panel for GLW. The effect on radiation appears to be significantly different between the two microphysics options.
swdown_12h_3ice
glw_12h_3ice

RELEASE NOTE:
The default Goddard Radiation and Goddard Microphysics schemes have been updated to utilize a 4-ice class.


Registry/Registry.EM_COMMON
```diff
@@ -2645,6 +2681,8 @@ package   nssl_1mom       mp_physics==19               -             moist:qv,qc
 package   nssl_1momlfo    mp_physics==21               -             moist:qv,qc,qr,qi,qs,qg
 package   nssl_2momg      mp_physics==22               -             moist:qv,qc,qr,qi,qs,qg;scalar:qndrop,qnr,qni,qns,qng,qvolg;state:re_cloud,re_ice,re_snow
 package   thompsonaero    mp_physics==28               -             moist:qv,qc,qr,qi,qs,qg;scalar:qni,qnr,qnc,qnwfa,qnifa;state:re_cloud,re_ice,re_snow,qnwfa2d,qnifa2d,taod5503d,taod5502d
+package   nuwrf3icescheme mp_physics==55             -               moist:qv,qc,qr,qi,qs,qg
```
dudhia: as far as I can tell these need the physc, etc. arrays and not the old option 7

davegill:
@cacruz @dudhia @weiwangncar 
Carlos,
I am able to get results that do not appear to be unphysical. Would you please clone my fork and checkout this branch, then do one of your known cases.

davegill:
@cacruz 
Carlos,
Is there an expectation that the Goddard MP and radiation must be used together? For example, I get a segmentation fault when running RRTMG with mp_physics=56. This is likely a package issue if these should be permitted to run together.

davegill:
@cacruz 
Carlos,
Would you be able to construct a double precision lookup table? If not, then we'll just put an if test in check_a_mundo that says we can't do double precision with the new Goddard schemes. We can add this data set in during the next month or so, for example.

davegill:
@dudhia @weiwangncar @cacruz @smileMchen 
Folks,
I think that we are there. The code is running as expected, and producing reasonable looking fields (and all of the expected fields, too).

Carlos,
1. Please test this code on one of your known cases.
2. Let us know what you want for a default MP scheme: the old code or one of the new versions.
3. Is there an assumption that the Goddard MP and RA need to be run together.
4. Can we get a double precision version of the look up table

Ming,
When you run your month-long simulations for the release testing, we should include a couple of combinations of the new Goddard options. It is possible that we will re-number the settings. This is just an FYI.  The code ran at 30 km, but the 4ice scheme has hail, so if you have a higher-resolution summer case, that might be appropriate.

davegill:
@dudhia @weiwangncar @liujake @jjguerrette
Wei and Jimy, 
Coding still to do (that we know about):
 - [ ] Probably need to put in a test in check_a_mundo to make sure that new Goddard MP and radiation only run with each other due to the effective radii: extra variables and different units.

Jake and JJ,
Would you check that this builds OK with DA?

Registry/registry.chem
```diff
@@ -1,3 +1,10 @@
+rconfig  integer gsfcrad_gocart_coupling namelist,chem   1  2
```
davegill: @ravanah @kayeekayee @stacywalters @dudhia @weiwangncar 
Folks,
I needed to add these fields for the Goddard code to build with WRF Chem. I don't know what the fields are, if that are in or out, need nesting or I/O.

weiwangncar:
Here is some results from running mp_physics = 7 + ra_*_physics = 5 over US on a 4 km grid. Plot on the left is from this combination of new physics, and on the right is from mp_physics = 6 and RRTMG radiation. The top two are RAINNC from 12 h forecast, and bottom two are SWDOWN at 12 h forecast time. The microphysics appears to have stronger impact on cloud and hence downward shortwave radiation.
rainnc_12h
swdown_12h

weiwangncar:
@dudhia Combinations of new Goddard 4ice scheme plus RRTMG, and WSM6 plus updated Goddard radiation ran. But it is not clear that when WSM6 is used with Goddard radiation, there is little or no microphysics effect on long and shortwave radiation. The radiation fields looked different than other runs. When RRTMG is used with Goddard 4ice scheme, it ran, but it is possible that there is no effective radii available to the radiation.

davegill:
@weiwangncar @dudhia 
Folks,
It looks like part of the SW optimization is to compute overcast only, and then use a lookup table for the clear sky work.

weiwangncar:
Here are a few plots showing the radiative impact from 4ice (left) and 3ice (right) schemes, both run with Goddard radiation. The top panel is for SWDOWN, and bottom panel for GLW. The effect on radiation appears to be significantly different between the two microphysics options.
swdown_12h_3ice
glw_12h_3ice

weiwangncar:
@davegill I'm ok with the code now. We can put in the code to make it more general and bullet-proof after the initial merge. Two other comments: 1. Do we want to remove NUWRF, and other name references in the code? 2. Do we want to change the module module_checkerror.F to something like module_checkerror_goddard.F?

dudhia:
I don't think we should remove NUWRF since it is their code.
Name references can be removed from general code, like dynamics and drivers,
Jimy

kkeene44:
@davegill @weiwangncar @dudhia
All regression tests pass for this branch.

davegill:
@kkeene44 
Kelly,
Thanks!
Dave

phys/module_physics_init.F
```diff
@@ -1472,20 +1472,27 @@ SUBROUTINE phy_init ( id, config_flags, DT, restart, zfull, zhalf,     &
...
-             re_rain_gsfc(i,k,j) = 0.
-             re_snow_gsfc(i,k,j) = 0.
-             re_ice_gsfc(i,k,j) = 0.
-             re_graupel_gsfc(i,k,j) = 0.
-             re_hail_gsfc(i,k,j) = 0.
-          end do
-       end do
-     end do
+    
+   if ( ( ( config_flags%ra_lw_physics .EQ. GODDARDLWSCHEME )   .AND. &
```
davegill: @weiwangncar @dudhia @powersjg 
Folks,
I think this commit fixes the Goddard memory leak I reported in the WRF Release meeting yesterday. On a nested case that used to fail immediately, this new version of the code now works. I just needed to put an IF test around initializing these six 3d arrays, due to packaging they are only allocated full size when running Goddard MP + Goddard radiation. I just started a big regtest with this. Fingers crossed.

dudhia:
OK, they are associated with only the LW scheme in the Registry, but we
need to make sure other parts of the
code like in the microphysics don't use these unless the LW scheme is used
too.

davegill:
@weiwangncar @dudhia @powersjg 
Jimy,
The six 3d re_*gsfc fields are in the call to the radiation driver and in the call to the MP driver, so the _EM_ Registry has these specific variables packaged for both Goddard MP and Goddard radiation. In the _NMM_ Registry, I was more relaxed about packaging (because we trap against using these schemes in check_a_mundo).

kkeene44:
@davegill @weiwangncar 
I have run some basic restart tests with this branch (the newly modified one from earlier today). The results are as follows:
1) A single domain run is okay for restarts (i.e., there are no differences between a standard run and a restart run at the ending time) for both the 3ice and 4ice schemes
2) Using 3ice or 4ice with an adaptive time-step run, only on d02 there are small diffs for the max map factor vars (MAX_MSTFX, MAX_MSTFY)
3) Using 3ice or 4ice with DFI or sst_update, there are many differences in files when doing a restart (with or without a nest).
4) Using 3ice or 4ice with a basic nest runs produces many differences in both domains when comparing files for the original run and the restart run.

I'm comparing files at boundary times.

davegill:
@kkeene44 @weiwangncar 
I am manually going through the remainder of the large regression test on cheyenne, since we are forbidden to run in a queue.
Some of the restart issues may be fixed with an idea from Wei. I need to commit those changes, but I want to finish the regtest first. We initializing some fields to zero. Maybe they are somehow getting re-initialized to zero incorrectly in some nested restart runs.

weiwangncar:
Congratulations!

PR #700, opened 2018-11-16
Use PGI compiler (with option) instead of cpp, /lib/cpp is not universal

bnordgren:
TYPE: enhancement

KEYWORDS: PGI, OpenHPC, configure/build

SOURCE: Bryce Nordgren (USDA Forest Service)

DESCRIPTION OF CHANGES: 
1. The canned PGI build configurations assume the GCC c preprocessor is located at /lib/cpp. OpenHPC allows you to install many different compiler variants and select the one to use via the modules system, but none have executables in /lib. These changes to the PGI configurations specify the preprocessor as "$(SCC) -E". Since the PGI C preprocessor does not accept input from stdin, it was also necessary to make minor changes to the Makefiles to use scratch files which are later removed. 
2. Only the pure PGI toolchain architectures were modified. Allowing the GNU toolchain c preprocessor to exist in non-hardecoded locations is a task left to a separate pull request.
3. Several makefiles hardcoded a -P option for the C preprocessor. This must be removed because -P means different things to different toolchains. This command line option needs to be included in the definition of the $(CPP) variable (and already is).

LIST OF MODIFIED FILES: 
M       arch/noopt_exceptions
M       arch/noopt_exceptions_f
M       arch/postamble
M       external/io_netcdf/makefile
M       external/io_pio/Makefile
M       external/io_pnetcdf/Makefile

TESTS CONDUCTED: 
1. Compiles and executes the em_b_wave scenario.


davegill:
@bnordgren 
I have a couple of comments that I am going to make from the source code directly so that they are easier to follow.

.gitignore
```diff
@@ -16,3 +16,4 @@
 configure.wrf*
 *.backup
 *.f90
+build.log
```
davegill: Would you revert this file please. There are a large number of dummy file names that could be used for the build log.

external/io_netcdf/makefile
```diff
@@ -6,7 +6,7 @@ CODE    = ext_ncd_get_dom_ti.code ext_ncd_get_var_td.code ext_ncd_get_var_ti.cod
 FFLAGS  =  $(FCFLAGS) -I$(NETCDFPATH)/include -I../ioapi_share
 LIBS    = $(LIB_LOCAL) -L$(NETCDFPATH)/lib -lnetcdf
 LIBFFS  = $(LIB_LOCAL) -L$(NETCDFPATH)/lib -lnetcdff -lnetcdf $(NETCDF4_DEP_LIB)
-CPP1    = $(CPP) -P $(TRADFLAG)
+CPP1    = $(CPP) $(TRADFLAG)
```
davegill: Similarly, would you add a comment as to why we want to remove `-P` from the CPP command from all of the io makefiles.

davegill:
@bnordgren 
Bryce (finally found your name!),
Thanks for the PR and the work to put it together.  I'll run a test with your mods on our local machines and see if everything is OK.

arch/configure.defaults
```diff
@@ -1544,8 +1544,8 @@ BYTESWAPIO      =       -byteswapio
 FCBASEOPTS_NO_G =       -w $(FORMAT_FREE) $(BYTESWAPIO) $(OMP)
 FCBASEOPTS      =       $(FCBASEOPTS_NO_G) $(FCDEBUG)
 MODULE_SRCH_FLAG =     -module $(WRF_SRC_ROOT_DIR)/main
-TRADFLAG        =      CONFIGURE_TRADFLAG
-CPP             =      /lib/cpp CONFIGURE_CPPFLAGS
+TRADFLAG        =      
+CPP             =      $(SCC) -E
```
davegill: I do not understand why only three locations have been targeted for modification. Should every PGI build use the new option that is suggested? For example, I see ten locations where there are PGI stanzas in the configure.defaults file.
```
>grep -n PGI configure.defaults | grep ARCH
133:#ARCH    Linux x86_64 ppc64le i486 i586 i686, PGI compiler with gcc # serial smpar dmpar dm+sm
176:#ARCH    Linux x86_64 ppc64le, PGI compiler with pgcc, SGI MPT # serial smpar dmpar dm+sm
219:#ARCH    Linux x86_64 ppc64le, PGI accelerator compiler with gcc # serial smpar dmpar dm+sm
802:#ARCH    Darwin (MACOS) PGI compiler with pgcc #serial smpar dmpar dm+sm
1210:#ARCH    Cray XC CLE/Linux x86_64, PGI compiler with gcc # serial dmpar smpar dm+sm
1514:#ARCH    Linux x86_64 ppc64le i486 i586 i686, PGI compiler with pgcc # serial smpar dmpar dm+sm
1557:#ARCH    CYGWIN_NT i686, PGI compiler on Windows # serial smpar dmpar dm+sm
1614:#ARCH    Linux x86_64 ppc64le i486 i586 i686, PGI compiler with gcc -f90= # serial smpar dmpar dm+sm
1657:#ARCH    Darwin (MACOS) PGI compiler with pgcc -f90= #serial smpar dmpar dm+sm
1789:#ARCH    Linux x86_64 ppc64le i486 i586 i686, PGI compiler with pgcc -f90= # serial smpar dmpar dm+sm
```

external/io_int/makefile
```diff
@@ -30,7 +30,7 @@ io_int.f: io_int.F90 module_internal_header_util.o
 	fi
 
 io_int.o: io_int.f ../../inc/intio_tags.h
-	$(FC) $(FCFLAGS) -I../../inc -I../ioapi_share -o $@ -c $*.f
+	$(FC) $(FCFLAGS) -I../../inc -I../../frame -I../ioapi_share -o $@ -c $*.f
```
davegill: Add an explanation in the commit message as to why `-I../../frame` is required for the include files for CPP.

external/io_netcdf/makefile
bnordgren: The Makefiles should not hard code options for a specific compiler toolchain if you're going to allow for the possibility that another toolchain is to be used. For the gnu variants, '-P' is already included in the definition of $(CPP), so you're covered. Using the -P option with the PGI compilers means "capture the output in a file instead of sending to stdout", which is not what you want anyway. The options necessary for successful c preprocessor operation need to be included in the $(CPP) variable.

.gitignore
bnordgren: Will do.

arch/configure.defaults
bnordgren: I only changed the pure Linux x86_64 ones because I don't have access to those other platforms. I assume that similar changes should be made to those other cases, but cannot verify. Basically I'm just erring on the side of caution. 

Anything which is a hybrid PGI/GCC configuration I also left alone because it was outside the scope of what I was trying to get to work. For those hybrids, I don't see a similar option on the gcc compiler to stop at the preprocessing step, hence one needs to run the c preprocessor directly. It's probably worth adding another environment variable for the configure script to pay attention to, leaving the default (if it's not set) to be /lib/cpp.

I hadn't solved the gnu side of things, leaving that for some future pull request.

external/io_int/makefile
bnordgren: First, this isn't a c preprocessor line, it's a compilation step. :)

2nd, I added this in a few places to resolve compiler errors similar to:

    PGF90-F-0004-Unable to open MODULE file module_internal_header_util.mod (diffwrf.F90: 51)

I also had to make this change to get the gnu compiler toolchain to work. If there's only one instance left, I suspect some of them were eliminated when I rebased.  I can go back and check to make sure this one is needed. This was a showstopper for me until I fixed it.

bnordgren: Looks like this one wasn't necessary or isn't necessary anymore. Removing.

bnordgren: The other two are not necessary either. It looks like the module files are now copied to the directory containing the code where they're used.

kkeene44:
@davegill @bnordgren 
Where do we stand on this? Have all the concerns been addressed?

davegill:
@kkeene44 
Running a few build tests now with PGI on cheyenne.

davegill:
@bnordgren 
Bryce,
When I diff all of the built .f90 files in the WRF directory, I get unexpected diffs in these:

./phys/module_sf_clm.f90
./phys/module_cam_infnan.f90
./external/io_grib1/io_grib1.f90
./external/io_grib_share/io_grib_share.f90
./frame/module_internal_header_util.f90

Some are various ways to represent a nan or inf value, that is probably OK. 
```
./phys/module_sf_clm.f90
190,191c190,192
<   real(r8), parameter :: inf = 1.e19
<   real(r8), parameter :: nan = 1.e21
---
> 
>   real(r8), parameter :: inf = O'0777600000000000000000'
>   real(r8), parameter :: nan = O'0777700000000000000000'
=====================

./phys/module_cam_infnan.f90
20,21c22,24
<   real(r8), parameter :: inf = 1.e19
<   real(r8), parameter :: nan = 1.e21
---
> 
>   real(r8), parameter :: inf = O'0777600000000000000000'
>   real(r8), parameter :: nan = O'0777700000000000000000'
```

I am more interested in the diffs in the IO directories:
```
=====================

./external/io_grib1/io_grib1.f90
0a1
> # 1 "io_grib1.F"
10c11
< 
---
> # 11 "io_grib1.F"
15,16c16
< 
< 
---
> # 17 "io_grib1.F"
18c18
< 
---
> # 19 "io_grib1.F"
24c24
< 
---
> # 25 "io_grib1.F"
```
```
./external/io_grib_share/io_grib_share.f90
0a1
> # 1 "io_grib_share.F"
8c9
< 
---
> # 9 "io_grib_share.F"
10c11
< 
---
> # 11 "io_grib_share.F"
21c22
< 
---
> # 22 "io_grib_share.F"

```
```
./frame/module_internal_header_util.f90
0a1
> # 1 "module_internal_header_util.F"
2c3
<
---
> # 3 "module_internal_header_util.F"
33c34
<
---
> # 34 "module_internal_header_util.F"
46c47
<
---
> # 47 "module_internal_header_util.F"
```

bnordgren:
Hi Dave,

I don't know that I can explain those differences in the built files.  To
my knowledge I did not change anything about the build process except
pointing the makefile at the other c preprocessor, and modifying the
pipes/redirection appropriately.

I'm really not going to be much use trying to track it down either. The
computer I was doing this work on is in my office, powered down, and the
office is closed till the shutdown ends.

Bryce

arch/configure.defaults
DWesl: > I don't see a similar option on the gcc compiler to stop at the preprocessing step

From https://gcc.gnu.org/onlinedocs/gcc-8.2.0/gcc/Overall-Options.html#Overall-Options:
> gcc -E
>    Stop after the preprocessing stage; do not run the compiler proper. 
>    The output is in the form of preprocessed source code, which is sent to the standard output.
>
>    Input files that don’t require preprocessing are ignored. 
so using the same definition of CPP would also work there.  This would also include the compiler-specific include directories, but I don't think those are relevant for WRF.

It looks like Intel uses the same flag with the same meaning, as well as a -EP flag mirroring gcc -E -P:
ifort: https://software.intel.com/en-us/fortran-compiler-developer-guide-and-reference-ep
icc: https://software.intel.com/en-us/node/523064

It looks like you plan to leave other compilers for a later pull request, so this will probably serve mostly as a reference for that future pull request.

davegill:

PR #703, opened 2018-11-17
Simple way to see what fields are allocated in WRF system

davegill:
TYPE: new feature

KEYWORDS: allocation, field list

SOURCE: internal

DESCRIPTION OF CHANGES:
A cpp flag is available to produce a list of fields that are allocated by
the WRF system programs. The list appears in the standard out file.

After the ./configure step, the configure.wrf file includes the line
```
                      -DSHOW_ALL_VARS_USED=0 \
```
To activate the option, this is changed to
```
                      -DSHOW_ALL_VARS_USED=1 \
```

A small sample of the hundreds of lines of output would look like:
```
xlat: bytes =    23856
xlong: bytes =    23856
lu_index: bytes =    23856
lu_mask: bytes =    23856
znu: bytes =      132
znw: bytes =      132
zs: bytes =       16
dzs: bytes =       16
u_gc: bytes =   620256
v_gc: bytes =   620256
t_gc: bytes =   620256
rh_gc: bytes =   620256
ght_gc: bytes =   620256
p_gc: bytes =   620256
```

LIST OF MODIFIED FILES:
M   arch/postamble      - location of the default cpp flag
M   frame/module_alloc_space.h       - need two declarations for printout variables
M   tools/gen_allocs.c        - construct the auto Fortran code

TESTS CONDUCTED:
The programs generate a list of fields in the order that in which they appear
in the Registry file. Associated with the list of fields is the number of bytes
that are allocated for the field.

RELEASE NOTE: This is expected to be for internal use only, so this feature does not
require any public explanation.


tools/gen_allocs.c
```diff
@@ -270,6 +270,13 @@ if ( tag == 1 )
 	   fprintf(fp,"  num_bytes_allocated = num_bytes_allocated + &\n(%s) * %cWORDSIZE\n",
                    array_size_expression("", "(", -1, t2, p, post_for_count, "model_config_rec%"),
                    tchar) ;
+	   fprintf(fp,"  nba = &\n(%s) * %cWORDSIZE\n",
+                   array_size_expression("", "(", -1, t2, p, post_for_count, "model_config_rec%"),
+                   tchar) ;
+           fprintf(fp,"#if ( SHOW_ALL_VARS_USED == 1 )\n") ;
+           fprintf(fp,"  WRITE(message_string,fmt='(a,i8)') '%s: bytes = ',nba\n", fname) ;
```
mgduda: Wouldn't this format only allow us to see the size of variables that were <100 MB?

davegill:
@mgduda 
Michael,
Good catch, I did not consider individual fields >= 100 MB. This is approximately a 500x500 domain with 100 vertical levels. I have certainly done this size domain on a single processor when trying to track down some bugs.

Initially, I was going to change the code to output the field size in kB, but when I looked at the surrounding code, the existing total sum of the fields was using integer(kind=8). So that is what I chose to follow.

1. The value that is printed is an integer(kind=8). My original code had the size of the value as a default integer. That would only provide a single extra decimal digit.
2. The format of the value printed is i12 (bumped up from the original i8). 

This is suitable for variables within decomposed domains that are < 1 TB. Even for a large memory machine, this modification for the variable size seems sufficient for us to not ever worry about it again.

kkeene44:
@davegill @mgduda 
Since Dave has addressed the above concern, is there anything else holding back this from being approved?

mgduda:
The PR looks fine to me. Does someone from the Infrastructure group need to leave an approving review?

kkeene44:
@mgduda 
No, you can approve it (or I can).

PR #704, opened 2018-11-17
 More than 10000 MPI tasks with WRF

thomasedds:
TYPE: new feature

KEYWORDS: MPI, large domains, RSL

DESCRIPTION OF CHANGES:
Problem:
Currently, the WRF code only permits 10000 MPI tasks and will stop with an error message in case more tasks are used. As domain size increases, sometimes more than 10000 tasks can be used.

Solution:
An environment variable LARGE_NUM_CORES has been introduced which needs to be set equal 1 (before running the configure script) by the user if more than 10000 tasks are desired. In order to be able to test a very large domain (so that 10000 tasks could be used), I also changed the NetCDF format for PNetCDF output from CDF-2 to CDF-5. The PNetCDF modification does not affect NCO, cdo, ncview, or ncl tools as recent versions of these standard utilities all have CDF-5 support. In case the user does not want to use > 10000 tasks, the default behavior for rsl.out and rsl.error is maintained.  The new MPI task limit in the WRF code is set to 350000 tasks.

ISSUE: Partially fixes wrf-model#615

LIST OF MODIFIED FILES:
M arch/Config.pl
M arch/postamble
M external/RSL_LITE/c_code.c
M external/RSL_LITE/rsl_lite.h
M external/io_pnetcdf/wrf_io.F90
M frame/module_bdywrite.F

TESTS CONDUCTED:

1. I tested a very large domain on Cheyenne using 275 and 280 nodes. These give 9900 cores (99x100) and 10080 cores (96x105). Results between the two simulations reveal zero differences in the wrfout files after running the model for a few minutes. I used ncdiff to test this.

2. I also set the switches "stderr_logging", "io_servers_silent", and "compute_tasks_silent" to different values and see no impact.

3. In case > 10000 tasks are requested and the environment variable is not set, the model exits with an error message.

RELEASE NOTE: WRF can now use more than 10000 MPI tasks.  The user must set the environment variable LARGE_NUM_CORES = 1 before running the configure script.


external/RSL_LITE/c_code.c
```diff
@@ -44,7 +44,12 @@ void RSL_LITE_ERROR_DUP1 ( int *me )
 
 /* redirect standard out*/
 # ifndef RSL0_ONLY
-    sprintf(filename,"rsl.out.%04d",*me) ;
+#ifdef LARGE_NUM_CORES 
+    sprintf(filename,"rsl.out.%06d",*me) ;
```
mgduda: From where `rsl_error_dup1` is called in `module_dm.F`, we do have the total number of tasks. We could consider just passing `ntasks` to this routine, then setting the number of digits in the rsl filenames accordingly. We do something similar in MPAS: https://github.com/MPAS-Dev/MPAS-Model/blob/master/src/framework/mpas_log.F#L194-L209 .

share/module_check_a_mundo.F
```diff
@@ -1838,6 +1838,26 @@ SUBROUTINE check_nml_consistency
...
          count_fatal_error = count_fatal_error + 1
       END IF
+
+
+!-----------------------------------------------------------------------
+! Did the user ask for too many total MPI tasks.
+!-----------------------------------------------------------------------
+
+      oops = 0
+      IF ( model_config_rec % nproc_x * model_config_rec % nproc_y .GT. 10000 ) THEN
```
mgduda: If we modify the RSL code to dynamically size the names of RSL files, and with a few other modifications, I don't think this sort of check should be needed.

frame/module_bdywrite.F
```diff
@@ -143,7 +143,7 @@ subroutine bdywrite(grid,filename)
     !        ims,ime, jms,jme, kms,kme, &
     !        ips,ipe, jps,jpe, kps,kpe)
 
-    assert(nf_create(trim(filename),NF_64BIT_OFFSET,fid))
+    assert(nf_create(trim(filename),NF_64BIT_DATA,fid))
```
mgduda: Is it really the case that we're running into *boundary* variables that are larger than 4 GB each? I admit to being rather unfamiliar with what exactly is in boundary files.

external/io_pnetcdf/wrf_io.F90
```diff
@@ -1257,7 +1257,10 @@ SUBROUTINE ext_pnc_open_for_write_begin(FileName,Comm,IOComm,SysDepInfo,DataHand
 !     if(newFileName(i:i) == '-') newFileName(i:i) = '_'
      if(newFileName(i:i) == ':') newFileName(i:i) = '_'
   enddo
-  stat = NFMPI_CREATE(Comm, newFileName, IOR(NF_CLOBBER, NF_64BIT_OFFSET), info, DH%NCID)
+! Thomas changed to NF_64BIT_DATA for CDF5 standard!!!                                                                              
+  stat = NFMPI_CREATE(Comm, newFileName, IOR(NF_CLOBBER, NF_64BIT_DATA), info, DH%NCID)
```
mgduda: Rather than hard-wiring `NF_64BIT_DATA` here, we could choose how to call `NFMPI_CREATE` based on a flag that could be set in the `SysDepInfo` argument that is passed to the `ext_pnc_open_for_write_begin` routine. The `open_w_dataset` routine appears to have access to namelist records and the sysdepinfo string, which is then passed to `wrf_open_for_write_begin` and then on to `ext_pnc_open_for_write_begin`, so it may not be too much work to have a namelist option control whether we use `NF_64BIT_OFFET` or `NF_64BIT_DATA`.

domingom:
This is a very useful capability for the WRF community, since there are more and more users starting to perform very large WRF runs that can benefit or simply need > 10,000 cores. I ported the modifications into my current WRFv4.0.1 repository and tested it (without the netcdf changes). The part regarding the rsl.* files works as expected, consistently producing rsl.out/error.XXXXXX files. However, for these very large runs, it is beneficial (if not a requirement due to memory issues) to use io_form_history, io_form_restart and io_form_input = 102 options. I tested that, and could not get beyond real.exe, since the modifications to allow more than 10,000 rsl files do not seem to propagate to this option, which still uses wrfinput_d01_XXXX, and therefore WRF crashes since the number of cores and the wrfinput_d01_XXXX chunks do not match. I would be happy to help with some testing since I am in the midst of performing a few quite large runs, once a solution to this is issue provided.

davegill:
@thomasedds @domingom @mgduda 
Folks,
This should be two PRs:
1. Modify what the rsl format looks like rsl.out.0000 or rsl.out._more zeros_
2. How to support CDF-5: NF_64BIT_OFFSET -> NF_64BIT_DATA

davegill:
@thomasedds 
This is now implemented in #1055 "Allow > 9999 MPI processes"
Closing

Issue #629, opened 2018-09-21
Increase CO2 concentration in RRTMG Routines from 379ppm to 400ppm


thomasedds:
The CO2 concentration is currently set to 379ppm. Recent observations suggest that the current concentration is close to 400ppm so this needs to be adjusted. May have an effect on longer term simulations.

davegill:
@thomasedds @weiwangncar @dudhia  

Here is what I found with some grep'ing in the physics directory:
```
module_ra_clWRF_support.F:          out = 379.e-6
module_ra_rrtm.F:   co2vmr = 379.e-6
module_ra_rrtmg_lw.F:! carbon dioxide (379 ppmv)
module_ra_rrtmg_lw.F:    data co2 / 379.e-6 /
module_ra_rrtmg_lwf.F:! carbon dioxide (379 ppmv)
module_ra_rrtmg_lwf.F:    data co2 / 379.e-6 /
module_ra_rrtmg_sw.F:! carbon dioxide (379 ppmv)
module_ra_rrtmg_sw.F:    data co2 / 379.e-6 /
module_ra_rrtmg_swf.F:! carbon dioxide (379 ppmv)
module_ra_rrtmg_swf.F:    data co2 / 379.e-6 /
module_ra_rrtmg_swk.F:! carbon dioxide (379 ppmv)
module_ra_rrtmg_swk.F:   data co2 / 379.e-6 /
```

Is this all of them?

dudhia:
Yes, at least for those options. Goddard would also have one.
Ideally this would depend on the date you are simulating,
Jimy

dudhia:
That would be a value from about 1980

PR #705, opened 2018-11-18
User defined values for CO2 content in RRTMG and NOAH-MP

thomasedds:
TYPE:  new feature

KEYWORDS: CO2 content, RRTMG, NOAH-MP

DESCRIPTION OF CHANGES:
Problem: 
Currently, the CO2 content in RRTMG is set to a value of 379 ppm while the CO2 content in NOAH-MP is hard-coded to 395 ppm. Users may wish to change these values without applying CLWRFGHG.

Solution:
Two new namelist settings have been introduced:
- co2_content: sets the CO2 content in RRTMG ( namelist&physics)
- co2_content_noahmp: sets the CO2 content in NOAH-MP (namelist&noahmp)
- In the current version, a check is included whether the user defines different CO2 values in RRTMG and NOAH-MP. If this is the case, the user receives a message in rsl.out.0000 but the model will continue.

LIST OF MODIFIED FILES:
M       phys/module_ra_rrtmg_lw.F
M       phys/module_sf_noahmplsm.F

TESTS CONDUCTED: Short high-resolution simulation to check for erroneous values in radiation and temperature fields. With default values of CO2=379 ppm in RRTMG and 395 ppm in NOAH-MP the results are identical with the original code.

RELEASE NOTE: RRTMG and NOAH-MP can now work with user-defined CO2 values. New namelist switches "co2_content" and "co2_content_noahmp" have been introduced.

### DIFF
--- a/Registry/Registry.EM_COMMON
+++ b/Registry/Registry.EM_COMMON
@@ -2232,6 +2232,7 @@ rconfig   integer alevsiz_cu              namelist,physics      1             1 
 rconfig   integer aercu_opt               namelist,physics      1             0       -      "aercu_opt"    "aerosol input option for multiscale KF"      ""
 rconfig   real    aercu_fct               namelist,physics      1             1.0     -      "aercu_fct"    "aerosol multiplication factor"               ""
 rconfig   integer aercu_used              derived               1             0       -      "aercu_used"   "derived nml for packaging"      ""
+rconfig   real    co2_content             namelist,physics      1             395.E-06 irh   "co2_content"  "CO2 content in RRTMG"      "ppm"
 
 #BSINGH - added shallowcu_forced_ra, numBins, thBinSize, rBinSize, minDeepFreq, minShallowFreq, shcu_aerosols_opt for CuP scheme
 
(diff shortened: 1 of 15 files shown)


davegill:
@thomasedds 
Thomas,
Would you take a look at #629 "Increase CO2 concentration in RRTMG Routines from 379ppm to 400ppm"
These is a line for Goddard, also. 

If we are incorporating these changes, let's get them all.

thomasedds:
@davegill @dudhia 
Dave, 
It took a while and I hope I found all locations to replace them for the namelist defined CO2. module_ra_clWRF_support.F from #629 does not require any change as the scheme's default value is used as fallback in case CO2 < 0 (which should never happen).

dudhia:
Another idea is to make it a function of year in case someone is doing past simulations.I have a simple function that fits reasonably if interested.Jimy

    On Monday, November 19, 2018, 11:59:04 AM MST, thomasedds <notifications@github.com> wrote:  
 
 
@davegill @dudhia
Dave,
It took a while and I hope I found all locations to replace them for the namelist defined CO2. module_ra_clWRF_support.F from #629 does not require any change as the scheme's default value is used as fallback in case CO2 < 0 (which should never happen).

weiwangncar:
@thomasedds @dudhia If we want CO2 value used in radiation and LSM the same, do we need to declare two variables?

dudhia:
The function I have for year dependence is
co2 = 280. + 90.*exp(0.02*(year-2000))
2000: co2=370
2018: co2=409
1950: co2=313
1900: co2=292
all are reasonable.

thomasedds:
@weiwangncar @dudhia 

Having two variables gives the flexibility for potential impact studies. That's why I kept it separate.

dudhia:
I am thinking of doing this a different way.
I would use a CO2 function of year so that the user does not need to be changing this for each case. There is a good function that fits past CO2.
co2(ppm)=280+90*exp(0.02*(year-2000))
This is graphed against observed values below.
Screen Shot 2020-01-23 at 9 44 40 AM

I would plan to put this in only RRTMG maybe as the default for the next release. Feedback needed.

davegill:
@dudhia 
Jimy,
Given the large range on the ordinate axis, could you generate a similar plot where the y-axis is "Computed - Measured CO2".

dudhia:
Yes, I have that plot. max is about 4 ppm (~1%). Much smaller recently.
Screen Shot 2020-01-23 at 10 10 25 AM

dudhia:
I will leave this PR open as there are still other radiation options and NoahMP that have the old CO2.
The code now has a year-dependent CO2 for RRTMG which supersedes some of this issue. Ideally year-dependent values can be applied to all these other options too.

davegill:

PR #712, opened 2018-11-28
new tslist options for wider variety of usage

kkeene44:
TYPE: enhancement

KEYWORDS: tslist, ideal, i/j, W, cell centers, tslist_ij, tslist_interp, time-series

SOURCE: Pat Hawbecker (NREL)

DESCRIPTION OF CHANGES: 
1. This code modification provides a way to generate time-series output for 3d idealized cases, based on the i/j location. The wrf_tsin.F and wrf_timeseries.F files are modified so that if map_proj == 0, tslist will read in integers (as opposed to floats) and assign those to grid%itsloc and grid%jtsloc. Therefore if it's an idealized case, tslist will expect coordinates instead of lat/lon. 

2. This modification also allows a user to use either lat/lon or i/j coordinates for time-series output locations with real data cases. This is now based on the header of tslist. If a user wishes to output at i,j locations, they would replace the lat/lon column headers with i/j.

3. Modifications are made to now output time-series for W.

4. A namelist option tslist_unstagger_winds (&domains) allows users to output all variables at cell centers when set to true (default is false). This now includes UU and VV, which are modified in wrf_timeseries.F and Registry.EM_COMMON. The output uses earth-relative winds for each.

LIST OF MODIFIED FILES: 
M     Registry/Registry.EM_COMMON
M     main/depend.common
M     run/README.tslist
M     share/Makefile
A     share/module_string_tools.F
M     share/wrf_timeseries.F
M     share/wrf_tsin.F

TESTS CONDUCTED: 
1. em_quarter_ss with default namelist, including a tslist (with i/j coordinates listed) in the run directory produces expected output. Contributor performed additional testing to output the full 3d field at every time step, then compared time series output to the 3d solution at the same location with a time-height plot, and took the difference - found that diffs could be attributed to roundoff error.
2. same as above, but with lat/lon specified in tslist, confirmed no time series output.
3. repeated tests 1) and 2) with another ideal case (em_convrad).
4. small real case (default namelist options) using an i/j-specific tslist. The expected time-series output is produced. Contributor performed additional testing to output the full 3d field at every time step, then compared time series output to the 3d solution at the same location with a time-height plot, and took the difference - found that diffs could be attributed to roundoff error.
5. same as above, but with lat/lon specified in tslist, confirmed no time series output.
6. same as 4), but also set tslist_interp = .true. The expected time-series output is produced, and U/V are calculated at the cell centers.
7. same as 4) but had a lat/lon tslist in the run directory. As expected, no time-series files were produced.
8. real case, as in 4), but using a lat/lon specific tslist, setting tslist_interp = .true. As expected, U/V variables are calculated at the cell centers. 
9. time-series files for variable W were produced for all runs that produced time-series output.
10. wrf history files are identical for both real/ideal runs using code before changes, and after changes.
11. TEST1 - use (I,J)
```
#-----------------------------------------------#
# 24 characters for name | pfx |   i   |   j    |
#-----------------------------------------------#
tower0001                 t0001      10       20  
tower0001                 t0003      10       10  
tower0012                 t0012      30       30  
```
12. TEST2 - WRONG use (LAT,LON)
```
#-----------------------------------------------#
# 24 characters for name | pfx |   i   |   j    |
#-----------------------------------------------#
tower0001                 t0001  33.116  -90.380
tower0001                 t0003  30.410  -90.678
tower0012                 t0012  35.072  -83.364
```
In the WRF model output, this tslist file produces this error message:
```
Error in tslist, line    4   
FATAL CALLED FROM FILE:  <stdin>  LINE:     237 
Error --- Maybe check that the header (I,J) vs (LAT,LON) matches the provided information
```
13. TEST3 - use (LAT,LON)
```
#-----------------------------------------------#
# 24 characters for name | pfx |  LAT  |  LON   |
#-----------------------------------------------#
tower0001                 t0001  33.116  -90.380
tower0001                 t0003  30.410  -90.678
tower0012                 t0012  35.072  -83.364
```
These are the identical lat,lon coordinates computed from TEST1, so we should get identical output (and we do).
```
> diff TEST1/t0001.d01.TS TEST3/t0001.d01.TS
1c1
< tower0001                  1  1 t0001 (  0.000,   0.000) (  10,  20) ( 33.116, -90.380)   44.5 meters
---
> tower0001                  1  1 t0001 ( 33.116, -90.380) (  10,  20) ( 33.116, -90.380)   44.5 meters
```
14. TEST4 - WRONG use (LAT,LON)
```
#-----------------------------------------------#
# 24 characters for name | pfx |  LAT  |  LON   |
#-----------------------------------------------#
tower0001                 t0001      10       20  
tower0001                 t0003      10       10  
tower0012                 t0012      30       30  
```
These are hard to detect. We can't really do a hard failure. The WRF model prints out a warning message.
```
WARNING
WARNING --- Maybe you have (I,J) locations for the (LAT,LON) values in tslist?
WARNING --- Are you purposefully studying Null Island?
WARNING
```

Example i/j-specific tslist: 
```
#-----------------------------------------------#
# 24 characters for name | pfx |   i   |   j    |
#-----------------------------------------------#
tower0001                 t0001      10       20
tower0001                 t0003      10       10
tower0012                 t0012      30       30
```

 

RELEASE NOTE: The time-series output option has been modified to now allow output at i,j coordinate locations. One must specify i/j coordinates in the tslist and change the column headers to i/j (instead of lat/lon). Both real and ideal cases are capable of outputting all variables (including U/V) at the cell centers by setting the namelist option tslist_unstagger_winds = .true. Additionally, time series output now includes variable W.

### DIFF
--- a/Registry/Registry.EM_COMMON
+++ b/Registry/Registry.EM_COMMON
@@ -231,7 +231,7 @@ i1       real   ph_tendf       ikj     dyn_em      1         Z
 i1       real   ph_save        ikj     dyn_em      1         Z 
 
 # Potential Temperature
-state    real   th_phy_m_t0    ikj      dyn_em      1         -     ihd  "t"      "perturbation potential temperature theta-t0" "K"
+state    real   th_phy_m_t0    ikj      dyn_em      1         -     ihd "t"      "perturbation potential temperature theta-t0" "K"
 state    real   t              ikjb     dyn_em      2         -     \
        i0rhusdf=(bdy_interp:dt)   "thm"      "either 1) pert moist pot temp=(1+Rv/Rd Qv)*(theta)-T0, or 2) pert dry pot temp=theta-T0; based on use_theta_m setting" "K"
 
--- a/Registry/Registry.EM_COMMON
+++ b/Registry/Registry.EM_COMMON
@@ -231,7 +231,7 @@ i1       real   ph_tendf       ikj     dyn_em      1         Z
 i1       real   ph_save        ikj     dyn_em      1         Z 
 
 # Potential Temperature
-state    real   th_phy_m_t0    ikj      dyn_em      1         -     ihd "t"      "perturbation potential temperature theta-t0" "K"
+state    real   th_phy_m_t0    ikj      dyn_em      1         -     ihd  "t"      "perturbation potential temperature theta-t0" "K"
 state    real   t              ikjb     dyn_em      2         -     \
        i0rhusdf=(bdy_interp:dt)   "thm"      "either 1) pert moist pot temp=(1+Rv/Rd Qv)*(theta)-T0, or 2) pert dry pot temp=theta-T0; based on use_theta_m setting" "K"
 
@@ -693,6 +693,7 @@ state    real   ts_rainnc       ?!       misc      -         -      -        "TS
 # Time series of vertical profile of U, V, GHT, THETA and QVAPOR
 state    real   ts_u_profile   ?!k       misc      -         -      -        "TS_U_PROFILE"   "Wind u-Component, earth relative, vertical profile"
 state    real   ts_v_profile   ?!k       misc      -         -      -        "TS_V_PROFILE"   "Wind v-Component, earth relative, vertical profile"
+state    real   ts_w_profile   ?!k       misc      -         -      -        "TS_W_PROFILE"   "Wind w-Component, earth relative, vertical profile"
 state    real   ts_gph_profile ?!k       misc      -         -      -        "TS_GPH_PROFILE" "Total geopotential height, vertical profile"
 state    real   ts_th_profile  ?!k       misc      -         -      -        "TS_TH_PROFILE"  "Total potential temperature, vertical profile"
 state    real   ts_qv_profile  ?!k       misc      -         -      -        "TS_QV_PROFILE"  "Water vapor mixing ratio, vertical profile"
@@ -2046,6 +2047,8 @@ rconfig   integer alloc_qndropsource      derived               1             0 
 rconfig   integer   num_moves       namelist,domains    1                0
 rconfig   integer   ts_buf_size     namelist,domains    1                200          -       "ts_buf_size"   "Size of time series buffer"
 rconfig   integer   max_ts_locs     namelist,domains    1                5            -       "max_ts_locs"   "Maximum number of time series locations"
+rconfig   logical   tslist_ij             derived       1              .false. rh  "tslist_ij"    "Use i,j locations in tslist"   ""
+rconfig   logical   tslist_unstagger_winds   namelist,domains    1              .false. rh  "tslist_unstagger_winds"    "interpolate u and v to cell centers"   ""
 rconfig   integer   vortex_interval  namelist,domains   max_domains      15  -  "" "" "minutes"
 rconfig   integer   max_vortex_speed namelist,domains   max_domains      40  -  "" "" "meters per second"
 rconfig   integer   corral_dist     namelist,domains    max_domains      8
--- a/run/README.tslist
+++ b/run/README.tslist
@@ -27,7 +27,7 @@ tower0002                 t0002  20       20
 tower0003                 t0003  30       30
 
 Given a tslist file, for each location inside a model domain (either coarse 
-or nested), below files are created:
+or nested), the following files are created:
 
 1. pfx*.dNN.TS containing the regular time series output of surface variables.
 2. pfx.dNN.UU containing a vertical profile of u wind component for each time step
--- a/run/README.tslist
+++ b/run/README.tslist
@@ -1,9 +1,10 @@
 To activate time series output in WRF, a file named "tslist" must be present
 in the WRF run directory. The tslist file contains a list of locations, given
-by their latitude and longitude, along with a short description and an 
-abbreviation for each location. The first three lines in the file are 
-regarded as header information, and are ignored. The contents of an example 
-tslist file are shown below. 
+by their latitude and longitude, or i/j coordinates, along with a short 
+description and an abbreviation for each location. The first three lines in 
+the file are regarded as header information, and are used to determine if the
+given coordinates are lat/long or i/j. The contents of an example tslist file 
+for coordinate specification are shown below. 
 
 
 #-----------------------------------------------#
@@ -15,15 +16,26 @@ Bogus point A             pt_a   29.718  -75.772
 Bogus point B             pt_b   37.614  -74.650
 
 
+If cell locations are to be used (such as for idealized cases), the i/j 
+locations are to be specified as follows:
+
+#-----------------------------------------------#
+# 24 characters for name | pfx |   J   |    I   |
+#-----------------------------------------------#
+tower0001                 t0001  10       10
+tower0002                 t0002  20       20
+tower0003                 t0003  30       30
+
 Given a tslist file, for each location inside a model domain (either coarse 
 or nested), below files are created:
 
 1. pfx*.dNN.TS containing the regular time series output of surface variables.
 2. pfx.dNN.UU containing a vertical profile of u wind component for each time step
 3. pfx.dNN.VV containing a vertical profile of v wind component for each time step
-4. pfx.dNN.TH containing a vertical profile of potential temperature for time step
-5. pfx.dNN.PH containing a vertical profile of geopotential height in time step
-6. pfx.dNN.QV containing a vertical profile of water vapor mixing ratio time step
+4. pfx.dNN.WW containing a vertical profile of w wind component for each time step
+5. pfx.dNN.TH containing a vertical profile of potential temperature for time step
+6. pfx.dNN.PH containing a vertical profile of geopotential height in time step
+7. pfx.dNN.QV containing a vertical profile of water vapor mixing ratio time step
 
 
 Where pfx is the specified prefix for the dd location in the tslist file, and NN is the domain ID, as given in 
@@ -39,12 +51,17 @@ however, smaller buffers will need to be flushed to disk more often than
 larger buffers. Thus, it is recommended that the size of the buffer be set 
 to the maximum number of time steps for any domain in a model run. 
 
+By default, the u, v, and w component winds are output on the staggered grid. To unstagger them,
+the namelist variable tslist_unstagger_winds can be set to true.
+
 Namelist.input variables related to to this new capability:
 
  * max_ts_locs = maximum number of locations in 'tslist' ( default is 5)
  * ts_buf_size = buffer size for time series output (default is 200)
  * max_ts_level = number of model levels for time series vertical
    profiles (default is 15)
+ * tslist_unstagger_winds = output the unstaggered u, v, and w component
+   winds (default is false)
 
 The first line in a time-series output of surface variables (pfx*.dNN.TS) looks like this:
 
--- a/run/README.tslist
+++ b/run/README.tslist
@@ -20,7 +20,7 @@ If cell locations are to be used (such as for idealized cases), the i/j
 locations are to be specified as follows:
 
 #-----------------------------------------------#
-# 24 characters for name | pfx |   J   |    I   |
+# 24 characters for name | pfx |   I   |    J   |
 #-----------------------------------------------#
 tower0001                 t0001  10       10
 tower0002                 t0002  20       20
--- a/share/wrf_timeseries.F
+++ b/share/wrf_timeseries.F
@@ -4,6 +4,9 @@
 !
 ! Michael G. Duda -- 25 August 2005
 ! vertical profiles added by Torge Lorenz -- June 2012
+! ability to output at either i/j or lat/lon locations, and ability to 
+!     output W, added by Pat Hawbecker -- Jan 2019
+! 
 !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
 SUBROUTINE calc_ts_locations( grid )
 
@@ -29,7 +32,7 @@ SUBROUTINE calc_ts_locations( grid )
    REAL :: known_lat, known_lon
    CHARACTER (LEN=132) :: message
    CHARACTER (LEN=24) :: ts_profile_filename
-   CHARACTER (LEN=2), DIMENSION(5) :: ts_file_endings = (/ 'UU', 'VV', 'PH', 'TH', 'QV' /)
+   CHARACTER (LEN=2), DIMENSION(6) :: ts_file_endings = (/ 'UU', 'VV', 'PH', 'TH', 'QV' ,'WW'/) 
    TYPE (PROJ_INFO) :: ts_proj
    TYPE (grid_config_rec_type) :: config_flags
 
@@ -148,12 +151,16 @@ SUBROUTINE calc_ts_locations( grid )
          CALL wrf_message(message)
    
          ntsloc_temp = 0
+
+         ! Boolean if/else for ij and lat/lon that sets ts_rx(y) to either ij or lat/lon value from tslist
          DO k=1,grid%ntsloc
-         
-            IF (config_flags%map_proj == 0) THEN  ! For idealized cases, no map transformation needed
+            ! Can be ideal case or specified i,j in tslist
+            IF (config_flags%map_proj == 0 .OR. grid%tslist_ij) THEN 
+                ts_rx = grid%itsloc(k)  
+                ts_ry = grid%jtsloc(k)
+            ELSE
                ts_rx = grid%lattsloc(k)           ! NB: (x,y) = (lat,lon) rather than (x,y) = (lon,lat)
                ts_ry = grid%lontsloc(k)
-            ELSE
                CALL latlon_to_ij(ts_proj, grid%lattsloc(k), grid%lontsloc(k), ts_rx, ts_ry)
             END IF
             
@@ -177,7 +184,7 @@ SUBROUTINE calc_ts_locations( grid )
          grid%ntsloc_domain = ntsloc_temp
    
          DO k=1,grid%ntsloc_domain
-   
+
             ! If location is outside of patch, we need to get lat/lon of TS grid cell from another patch
             IF (grid%itsloc(k) < ips .OR. grid%itsloc(k) > ipe .OR. &
                 grid%jtsloc(k) < jps .OR. grid%jtsloc(k) > jpe) THEN
@@ -232,7 +239,7 @@ SUBROUTINE calc_ts_locations( grid )
                CLOSE(UNIT=iunit)
 
                ts_profile_filename = grid%ts_filename(k)
-               DO j=1,5
+               DO j=1,SIZE(ts_file_endings)
                   ! Create the output files for the vertical profiles, one file for each variable
                   iunit = get_unused_unit()
                   IF ( iunit <= 0 ) THEN
@@ -275,6 +282,7 @@ END SUBROUTINE calc_ts_locations
 SUBROUTINE calc_ts( grid )
 
    USE module_domain
+   USE module_configure, ONLY : model_config_rec, grid_config_rec_type, model_to_grid_config_rec
    USE module_model_constants
 
    IMPLICIT NONE
@@ -290,6 +298,7 @@ SUBROUTINE calc_ts( grid )
            output_t, output_q, clw, xtime_minutes
    REAL, ALLOCATABLE, DIMENSION(:) :: p8w
    REAL, ALLOCATABLE, DIMENSION(:) :: earth_u_profile, earth_v_profile
+   TYPE (grid_config_rec_type) :: config_flags
 
    ! Parameter ts_model_level:  
        ! TRUE to output T, Q, and wind at lowest model level
@@ -310,6 +319,7 @@ SUBROUTINE calc_ts( grid )
    n = grid%next_ts_time
 
    ALLOCATE(p8w(grid%sm32:grid%em32))
+   CALL model_to_grid_config_rec ( grid%id , model_config_rec , config_flags )
 
    DO i=1,grid%ntsloc_domain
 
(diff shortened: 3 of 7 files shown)


share/wrf_tsin.F
```diff
@@ -24,7 +26,7 @@ SUBROUTINE wrf_tsin ( grid , ierr )
 #if ((EM_CORE == 1) && (DA_CORE != 1))
     IF ( grid%dfi_opt == DFI_NODFI .OR. (grid%dfi_opt /= DFI_NODFI .AND. grid%dfi_stage == DFI_SETUP) ) THEN
 #endif
-
+CALL model_to_grid_config_rec ( grid%id , model_config_rec , config_flags )
```
mgduda: I think this line should be indented.

mgduda:
I'm not completely convinced of the benefits of the (i,j) versus (lat,lon) changes that are part of this PR. It's been a while since I last looked at the timeseries code, but if the map projection is 0, we assume an idealized case, in which case the latitude column specified in the `tslist` file is used as the real-valued i-coordinate and the longitude column specified in the file is used as the real-valued j-coordinate. Part of what this PR does -- if I'm reading the code correctly -- is to add more complicated logic for reading the columns from the `tslist` file into `itsloc` and `jtsloc` before assigning to `ts_rx` and `ts_ry`, rather than reading columns into `lattsloc` and `lontsloc` before `ts_rx` and `ts_ry`. Of course, I'm probably missing something here. But, it does seem like all of the extra logic that provides essentially no change to the interface seen by the user is not worth the complication it introduces.

share/wrf_timeseries.F
```diff
@@ -342,8 +355,14 @@ SUBROUTINE calc_ts( grid )
             !
 #if (EM_CORE == 1)
             DO k=1,grid%max_ts_level
-            earth_u_profile(k) = grid%u_2(ix,k,iy)*grid%cosa(ix,iy)-grid%v_2(ix,k,iy)*grid%sina(ix,iy)
-            earth_v_profile(k) = grid%v_2(ix,k,iy)*grid%cosa(ix,iy)+grid%u_2(ix,k,iy)*grid%sina(ix,iy)
+                ! interpolation loop: if you want u,v on cell centers, tslist_interp = .true.
+                IF (config_flags%tslist_interp) THEN
+                    earth_u_profile(k) = (grid%u_2(ix,k,iy)+grid%u_2(ix+1,k,iy))/2.0 
+                    earth_v_profile(k) = (grid%v_2(ix,k,iy)+grid%v_2(ix,k,iy+1))/2.0
```
mgduda: One potential issue here is that, if `config_flags%tslist_interp` is `.true.`, we get *grid-relative* winds rather than *earth-relative* winds. A user may think they are just getting cell-centered winds that are still earth-relative, but what they're actually getting are cell-centered winds that are grid-relative. I would suggest this need to be reworked to provide a consistent basis for the wind fields, whether they are cell-centered or at their native locations.

Registry/Registry.EM_COMMON
```diff
@@ -2046,6 +2047,8 @@ rconfig   integer alloc_qndropsource      derived               1             0
 rconfig   integer   num_moves       namelist,domains    1                0
 rconfig   integer   ts_buf_size     namelist,domains    1                200          -       "ts_buf_size"   "Size of time series buffer"
 rconfig   integer   max_ts_locs     namelist,domains    1                5            -       "max_ts_locs"   "Maximum number of time series locations"
+rconfig   logical   tslist_ij       namelist,domains    1              .false.        rh      "tslist_ij"    "Use i,j locations in tslist"   ""
+rconfig   logical   tslist_interp   namelist,domains    1              .false.        rh      "tslist_interp"    "interpolate u and v to cell centers"   ""
```
mgduda: `tslist_interp` seems like an overly general name for an option that specifically controls whether the horizontal winds are interpolated from their C-grid staggering to an A-grid. Perhaps this option could be called something like `tslist_destagger_winds` or something similar?

share/wrf_tsin.F
```diff
@@ -24,7 +26,7 @@ SUBROUTINE wrf_tsin ( grid , ierr )
 #if ((EM_CORE == 1) && (DA_CORE != 1))
     IF ( grid%dfi_opt == DFI_NODFI .OR. (grid%dfi_opt /= DFI_NODFI .AND. grid%dfi_stage == DFI_SETUP) ) THEN
 #endif
-
+    CALL model_to_grid_config_rec ( grid%id , model_config_rec , config_flags )
```
mgduda: A few more columns of indentation would be nice, since we're inside of an if-test.

share/wrf_tsin.F
```diff
@@ -54,9 +56,22 @@ SUBROUTINE wrf_tsin ( grid , ierr )
...
-                   READ(UNIT=iunit, FMT='(A25,1X,A5,1X,F7.3,1X,F8.3)', IOSTAT=istatus)            &
-                        grid%desctsloc(grid%ntsloc+1), grid%nametsloc(grid%ntsloc+1), &
-                        grid%lattsloc(grid%ntsloc+1), grid%lontsloc(grid%ntsloc+1)
+                   ! if tslist is ij, read in integers and not floats
+                   IF (config_flags%map_proj == 0) THEN !ideal run will only use i,j
+                       READ(UNIT=iunit, FMT='(A25,1X,A5,1X,I9,1X,I10)', IOSTAT=istatus)            &
+                       grid%desctsloc(grid%ntsloc+1), grid%nametsloc(grid%ntsloc+1), &
+                       grid%itsloc(grid%ntsloc+1), grid%jtsloc(grid%ntsloc+1)
+                   ELSE
+                       IF (config_flags%tslist_ij) THEN
```
mgduda: To avoid repeating three dense lines of reading code, we could combine this if-test with the one on `map_proj` so that the code block looks like this:
```fortran
                   ! If this is an idealized case (with no map projection) or if coordinates are
                   ! specified as (i,j) read into itsloc and jtsloc; otherwise, read into lattsloc
                   ! and lontsloc.
                   IF ( (config_flags%map_proj == 0) .OR. config_flags%tslist_ij ) THEN 
                       READ(UNIT=iunit, FMT='(A25,1X,A5,1X,I9,1X,I10)', IOSTAT=istatus) &
                       grid%desctsloc(grid%ntsloc+1), grid%nametsloc(grid%ntsloc+1),    &
                       grid%itsloc(grid%ntsloc+1), grid%jtsloc(grid%ntsloc+1)
                   ELSE
                       READ(UNIT=iunit, FMT='(A25,1X,A5,1X,F7.3,1X,F8.3)', IOSTAT=istatus) &  
                       grid%desctsloc(grid%ntsloc+1), grid%nametsloc(grid%ntsloc+1),       &
                       grid%lattsloc(grid%ntsloc+1), grid%lontsloc(grid%ntsloc+1)
                   END IF
```

davegill:
@kkeene44 @mgduda 
Kelly,

Has the developer responded to Michael's suggested modifications? The original tslist option was largely put together without using the namelist. The existence of a file was sufficient to enable the option. Continuing on with this notion, where now the form of the headers in this input file dictates the interpretation of the data, seems a reasonable idea. It follows with the original intent.

 - [ ] The run/README.tslist file should be updated to reflect whatever changes are eventually made.

kkeene44:
@davegill @mgduda 
Yes, he has responded in the forum. It's on my "to-do" list to take a look at the changes he made. I'll do this soon.

kkeene44:
@mgduda 
I have updated the PR with the user's modified files. He has incorporated the following changes:
1. The code now determines if lat/lon or i,j should be used based on the header of tslist, and it is no longer a namelist option. Ideal cases will only use i,j.
2. The previous 'tslist_interp' is now changed to 'tslist_unstagger_winds' and the output is now earth-relative for each.
3. The recommended coding logic modifications have been included. 
4. The w-component wind speed is still output by default with these changes.

Take a look and see if you have any additional comments/suggestions/concerns.

Registry/Registry.EM_COMMON
```diff
@@ -231,7 +231,7 @@ i1       real   ph_tendf       ikj     dyn_em      1         Z
 i1       real   ph_save        ikj     dyn_em      1         Z 
 
 # Potential Temperature
-state    real   th_phy_m_t0    ikj      dyn_em      1         -     ihd "t"      "perturbation potential temperature theta-t0" "K"
+state    real   th_phy_m_t0    ikj      dyn_em      1         -     ih  "t"      "perturbation potential temperature theta-t0" "K"
```
mgduda: Why has this line been changed?

Registry/Registry.EM_COMMON
```diff
@@ -693,6 +693,8 @@ state    real   ts_rainnc       ?!       misc      -         -      -        "TS
 # Time series of vertical profile of U, V, GHT, THETA and QVAPOR
 state    real   ts_u_profile   ?!k       misc      -         -      -        "TS_U_PROFILE"   "Wind u-Component, earth relative, vertical profile"
 state    real   ts_v_profile   ?!k       misc      -         -      -        "TS_V_PROFILE"   "Wind v-Component, earth relative, vertical profile"
+#PSH
```
mgduda: @davegill can comment authoritatively, but I think we're generally recommending that contributors not leave their initials around the code.

Registry/Registry.EM_COMMON
```diff
@@ -2424,7 +2429,7 @@ rconfig   real    scm_force_dx            namelist,scm  1       4000.       rh
 rconfig   integer num_force_layers        namelist,scm  1       8           rh   "num_force_layers"     "Number of SCM forcing layers" ""
 rconfig   integer scm_lu_index            namelist,scm  1       2           rh   "scm_lu_index"         "SCM landuse index" ""
 rconfig   integer scm_isltyp              namelist,scm  1       4           rh   "scm_isltyp"           "SCM soil category" ""
-rconfig   real    scm_vegfra              namelist,scm  1       50.         rh   "scm_vegfra"           "SCM vegetation fraction" ""
+rconfig   real    scm_vegfra              namelist,scm  1       0.5         rh   "scm_vegfra"           "SCM vegetation fraction" ""
```
mgduda: Why is this change part of the PR?

share/wrf_timeseries.F
```diff
@@ -177,7 +181,6 @@ SUBROUTINE calc_ts_locations( grid )
          grid%ntsloc_domain = ntsloc_temp
    
          DO k=1,grid%ntsloc_domain
-
```
mgduda: I prefer having whitespace between an executable line of code and a comment line, but that's just my opinion.

share/wrf_timeseries.F
```diff
@@ -232,7 +235,8 @@ SUBROUTINE calc_ts_locations( grid )
                CLOSE(UNIT=iunit)
 
                ts_profile_filename = grid%ts_filename(k)
-               DO j=1,5
+               !PSH - had to change from 5 to 6 to include header in 'WW'
```
mgduda: As written, this comment isn't quite as useful as it could be (and it has initials). I haven't actually tried it, but would simply setting the loop to
```fortran
               DO j=1,SIZE(ts_file_endings)
```
and omitting the comment altogether work?

share/wrf_timeseries.F
```diff
@@ -645,6 +671,23 @@ SUBROUTINE write_ts( grid )
          END DO
          CLOSE(UNIT=iunit)
 
+         iunit = get_unused_unit()
+            IF ( iunit <= 0 ) THEN
+            CALL wrf_error_fatal('Error in write_ts: could not find a free Fortran unit.')
+            END IF
```
mgduda: The indentation here looks off.

mgduda: I just noticed that we're following what was done for other profiles. I suppose we can keep this as is, or fix the indentation in the corresponding code for other profiles.

share/wrf_tsin.F
```diff
@@ -46,17 +49,35 @@ SUBROUTINE wrf_tsin ( grid , ierr )
    
              IF (istatus == 0) THEN
    
-                ! Ignore first three lines, which constitute a header
-                READ(UNIT=iunit, FMT='(1X)')
+                ! The header will be parsed to see if lat/lon or i/j is given...
                 READ(UNIT=iunit, FMT='(1X)')
+                READ(UNIT=iunit, FMT='(A32,A7,A2,A7,A1)') head1, JorLat, head2, IorLon, head3
```
mgduda: Since `head1`, `head2`, and `head3` don't seem to be used, I think we can discard these variables and change the `READ` line to look something like:
```fortran
                READ(UNIT=iunit, FMT='(32X,A7,2X,A7)') JorLat, IorLon
```

share/wrf_tsin.F
```diff
@@ -46,17 +49,35 @@ SUBROUTINE wrf_tsin ( grid , ierr )
...
                 READ(UNIT=iunit, FMT='(1X)')
+                grid%tslist_ij = .false.
+                ! If the header contains "LAT" and "LON" then it will expect
+                !... lat/lon coords in the file
+                ! If the header has "J" and "I", then i,j will be used.
+                if(index(JorLat,'LAT').ne.0 .and. index(IorLon,'LON').ne.0) then
+                    grid%tslist_ij = .false.
+                elseif(index(JorLat,'J').ne.0 .and. index(IorLon,'I').ne.0) then
+                    grid%tslist_ij = .true.
+                endif
```
mgduda: It may be nice to handle the case where creative users write these in lower-case letters, or in mixed-case.

@davegill Does WRF already provide a standard facility for capitalizing a string? If not, we could add something like the following to an appropriate place in WRF:
```fortran
    function capitalize(str) result(capStr)
!<DESCRIPTION>
!
! Returns a copy of 'str' in which all lower-case letters have been converted
! to upper-case letters.
!
!</DESCRIPTION>

        character(len=*), intent(in) :: str
        character(len=len(str)) :: capStr

        integer :: i
        integer, parameter :: offset = (iachar('a') - iachar('A'))


        do i=1,len(str)
            if (iachar(str(i:i)) >= iachar('a') .and. iachar(str(i:i)) <= iachar('z')) then
               capStr(i:i) = achar(iachar(str(i:i)) - offset)
            else
               capStr(i:i) = str(i:i)
            end if
        end do

    end function capitalize
```
If we go this route, I can provide a version that has capitalized Fortran keywords to match the rest of the WRF infrastructure.

mgduda:
As a general comment, we should probably also update the `README.tslist` file to reflect the changes in this PR, specifically, the option to specify coordinates in (lat,lon) or (i,j), and the addition of a w profile.

mgduda:
Also, I had mentioned removing the contributor's initials from the code changes, but, I think it would be completely appropriate to give credit to the contributor by adding their name, e.g., to the top of `wrf_timeseries.F` file.

cmattocks:
iunit = get_unused_unit()

Isn’t this what the Fortran 2008 newUnit function was designed for?

Craig

Registry/Registry.EM_COMMON
kkeene44: @mgduda 
The addition of the 'd' is recent in the code (sometime after V4.0) and so was on the develop branch, from where I created this branch. This is an oversight by me. I'll add it back.

Registry/Registry.EM_COMMON
kkeene44: @mgduda @davegill 
I'll remove them.

Registry/Registry.EM_COMMON
kkeene44: @mgduda 
Same as the oversight above. I'll correct it.

share/wrf_timeseries.F
kkeene44: @mgduda 
I don't see harm in having the space. I'll add it back in.

share/wrf_tsin.F
kkeene44: @mgduda 
Agreed. I'll change it.

kkeene44:
@mgduda 
I added a credit to the developer at the top of the wrf_timeseries.F file.

davegill:
@kkeene44 @mgduda 
Folks,
As far as a standard "turn this to uppercase" function in WRF, I can't find anything that would be available in the executables. There are functions in the tools directory, but those are unavailable during a real/wrf run.

This pair is used frequently to go the other way, upper case to lower case.
First the offset between upper and lower case:
```
     INTEGER       ,PARAMETER   :: upper_to_lower =IACHAR('a')-IACHAR('A')
```
Then each element of the character string is tested if it is uppercase. If so, then the string element is modified.
```
       if('A'<=c .and. c <='Z') outstr(i:i)=achar(iachar(c)+upper_to_lower)
```

share/wrf_timeseries.F
phawbeck: I have incorporated this change and removed the comment.

kkeene44:
@mgduda @davegill 
All concerns have been addressed except the question regarding upper/lower case above (which I think the two of you are trying to hash out).

Registry/Registry.EM_COMMON
```diff
@@ -231,7 +231,7 @@ i1       real   ph_tendf       ikj     dyn_em      1         Z
 i1       real   ph_save        ikj     dyn_em      1         Z 
 
 # Potential Temperature
-state    real   th_phy_m_t0    ikj      dyn_em      1         -     ihd "t"      "perturbation potential temperature theta-t0" "K"
+state    real   th_phy_m_t0    ikj      dyn_em      1         -     ihd  "t"      "perturbation potential temperature theta-t0" "K"
```
mgduda: Let's avoid this whitespace change.

share/wrf_tsin.F
```diff
@@ -46,17 +49,35 @@ SUBROUTINE wrf_tsin ( grid , ierr )
...
                 ! Read in time series locations
                 istatus = 0
                 DO WHILE (istatus == 0)
-                   READ(UNIT=iunit, FMT='(A25,1X,A5,1X,F7.3,1X,F8.3)', IOSTAT=istatus)            &
-                        grid%desctsloc(grid%ntsloc+1), grid%nametsloc(grid%ntsloc+1), &
-                        grid%lattsloc(grid%ntsloc+1), grid%lontsloc(grid%ntsloc+1)
+                   IF (config_flags%map_proj == 0 .OR. grid%tslist_ij) THEN !ideal run will only use i,j
+                       READ(UNIT=iunit, FMT='(A25,1X,A5,1X,I9,1X,I10)', IOSTAT=istatus)            &
+                       grid%desctsloc(grid%ntsloc+1), grid%nametsloc(grid%ntsloc+1), &
+                       grid%itsloc(grid%ntsloc+1), grid%jtsloc(grid%ntsloc+1)
```
mgduda: Is this correct? The header suggests that the coordinates are given as (J,I), but here we're reading them as (I,J). I'd actually prefer to give the coordinates as I first, then J, as that seems most natural; but that's only my opinion.

share/wrf_timeseries.F
```diff
@@ -148,12 +151,16 @@ SUBROUTINE calc_ts_locations( grid )
...
-         
-            IF (config_flags%map_proj == 0) THEN  ! For idealized cases, no map transformation needed
+            ! Can be ideal case or specified i,j in tslist
+            IF (config_flags%map_proj == 0 .OR. grid%tslist_ij) THEN 
+                ts_rx = grid%itsloc(k)  
+                ts_ry = grid%jtsloc(k)
+            ELSE
                ts_rx = grid%lattsloc(k)           ! NB: (x,y) = (lat,lon) rather than (x,y) = (lon,lat)
                ts_ry = grid%lontsloc(k)
-            ELSE
```
mgduda: The code for setting `ts_rx` and `ts_ry` now looks like:
```fortran
            IF (config_flags%map_proj == 0 .OR. grid%tslist_ij) THEN 
                ts_rx = grid%itsloc(k)  
                ts_ry = grid%jtsloc(k)
            ELSE
               ts_rx = grid%lattsloc(k)           ! NB: (x,y) = (lat,lon) rather than (x,y) = (lon,lat)
               ts_ry = grid%lontsloc(k)
               CALL latlon_to_ij(ts_proj, grid%lattsloc(k), grid%lontsloc(k), ts_rx, ts_ry)
            END IF
```
The `NB` comment no longer makes sense in its new position, and it seems that there's no point to assigning `ts_rx` and `ts_ry` the values from `lattsloc` and `lontsloc`, only to overwrite them with the call to `latlon_to_ij`.

mgduda:
If there are no objections, I can push a commit to handle the upper-case/lower-case issue.

share/wrf_tsin.F
phawbeck: I will run some tests to verify this - I kept it as (J,I) so it would be equivalent to lat/lon, but I agree that (I,J) is more natural. I will change it to (I,J) if this is not the case.

phawbeck: I have verified that it is, indeed, reading in (I,J) but the header says (J,I). I have modified this so that the header must say (I,J); if it says (J,I) then no towers are generated. 

Also, I think it may be redundant for the 'IF' statement to include the `config_flags%map_proj == 0` statement now. Any idealized case will use (I,J) in the header, as well as, any real case that specifies (I,J). If everybody agrees, the new 'IF' statement will read as: `IF ( grid%tslist_ij) THEN`.

kkeene44:
@mgduda 
Yes, go ahead. Thanks!

run/README.tslist
```diff
@@ -15,15 +16,26 @@ Bogus point A             pt_a   29.718  -75.772
 Bogus point B             pt_b   37.614  -74.650
 
 
+If cell locations are to be used (such as for idealized cases), the i/j 
+locations are to be specified as follows:
+
+#-----------------------------------------------#
+# 24 characters for name | pfx |   J   |    I   |
```
davegill: Did we agree that this was going to be lat lon, but also i j (not J I)?

phawbeck: The header should be I, J (not J, I). Should I update the README.tslist file?

davegill: Yes, please update this README.

phawbeck: See attached file.
[README.tslist.txt](https://github.com/wrf-model/WRF/files/2815237/README.tslist.txt)

share/wrf_timeseries.F
```diff
@@ -379,6 +399,7 @@ SUBROUTINE calc_ts( grid )
             DO k=1,grid%max_ts_level
             grid%ts_u_profile(n,i,k)   = earth_u_profile(k)
             grid%ts_v_profile(n,i,k)   = earth_v_profile(k)
+            grid%ts_w_profile(n,i,k)   = (grid%w_2(ix,k,iy)+grid%w_2(ix,k+1,iy))/2.0 ! w on cell center
             grid%ts_gph_profile(n,i,k) = (grid%phb(ix,k,iy)+grid%ph_2(ix,k,iy))/9.81 
             grid%ts_th_profile(n,i,k)  = grid%t_2(ix,k,iy) + 300
```
davegill: @phawbeck @mgduda @kkeene44 
The grid%t_2 variable is likely to be a moist potential temperature perturbation (based on a nml setting).
t_2 = ( full dry theta ) * ( 1 + Rv/Rd Qv ) - 300
Should I fix this in a separate PR, or address this theta inconsistency in this PR?

davegill: It is possible that the sensible temperature (if T is based on an assumed dry potential temperature) is incorrect also, but I have not looked at this code thoroughly.

davegill: @phawbeck @kkeene44 @mgduda 
Folks,
There is likely a bug in the treatment of potential temperature for the tslist option, but there is likely this exact same bug in several other diagnostic routines (with the introduction of moist theta in v4.0). We will not deal with the dry potential temperature issue in this PR.

run/README.tslist
davegill: I was confused with the attachment of a text file. I just modified the file README.tslist.

davegill:
@phawbeck @kkeene44 @mgduda 
Folks,
Trying to figure out the status of this PR. Can I get a quick summary of what is outstanding?

run/README.tslist
phawbeck: I've never used github - modifying and attaching a file seemed easiest.

davegill: @phawbeck 
OK, gotcha.

kkeene44:
@davegill 
It looks like there are a couple of issues that @mgduda had above. @phawbeck made an adjustment to one, and it looks like we just need approval that what he's done is how we want it done. The other has not been addressed.

phawbeck:
Which issue hasn't been addressed?

share/wrf_timeseries.F
kkeene44: @phawbeck 
This is the comment I was referring to that had not been resolved (I don't think - I could be wrong).

phawbeck: Ok, thanks - I wasn't sure what the comment was proposing. Remove the `!NB` comment and remove the `ts_rx = ...` and `ts_ry = ...` lines after the `ELSE` statement? Will ts_rx and ts_ry what will be set within the `latlon_to_ij`command?

kkeene44:
@davegill 
I made the changes we discussed (regarding ts_rx/ts_ry), so now the only thing outstanding is the note to you (from Michael) above on the wrf_tsin file from Jan 3rd.

share/wrf_tsin.F
davegill: @mgduda @kkeene44 @phawbeck 
Folks,
I have implemented Michael's suggestion (and if you look at the code, you'll see how closely I took his suggestion). In the share directory is a module_string_tools.F This has a single function (right now) that turns an input string into an upper case string.

cmattocks:
Excellent, Dave!

Having a separate string utility ‘class' (module_string_tools.F) is an advance in the WRF software architecture.

Craig

gsever:
While this pull request is still in-progress, would you be able to add logic to output additional variables (either profiles of pressure and/or temperature)? Thanks.

PR #717, opened 2018-11-30
BF: Fix registry for input geogrid fields that are missing either "i2" or "d"

smileMchen:
TYPE: bug fix

KEYWORDS: ndown, nesting run

SOURCE: internal, problem discovered by researchers at Shanghai Typhoon Institute

DESCRIPTION OF CHANGES: 
Problem:
A few variables are identically zero in the fine-grid ndown run or nesting run with input_from_file = .false..
 
Solution:
The particular variables are those static fields that go from geogrid -> metgrid -> real -> WRF. The fields must all be "i2" variables (static fields from geogrid that should be in the wrfinput file during ndown). These are also fields that need to be horizontally interpolated in case there is no fine grid input file (requiring a "d"). 

The Registry.EM_COMMON file is modified to ensure these variables are available for fine-grid.
1. VAR_SSO and FRC_URB2D have "i2" added
2. EROD and SNOALB have "d" added

LIST OF MODIFIED FILES:

To make sure that we have not overlooked any other similar variables, we follow the steps below to conduct a complete check: 
1. Run geogrid, and get a list of all of the fields in the geo_em* file.
2. Identify all fields in the geo_em file (they are all passed out of metgrid unchanged) that ALSO have an "i01" setting in the WRF Registry (basically identifying fields that meet the criteria that the field came in from metgrid and that the field that MUST go into the WRF model).
3.  Make sure that those identified fields have two settings: "i2" (so that they are identified as static fields for the fine grid for ndown, for example) and the nested option "d" (to interpolate down to the fine grid from the coarse grid).

TESTS CONDUCTED:  
Tests are conducted with and without the fix. Results indicate that these previously missing variables are available for fine-grid. Figures below show the differences in VAR_SSO between wrfinput_d02 with and without the bug fix. Similar differences can be found for other variables (i.e., EROD, SNOALB, FRC_URB2D)

screen shot 2018-11-30 at 2 51 08 pm

RELEASE NOTE: 
When running the ndown program or nesting without an input file on a particular domain, the following variables incorrectly remained zero on the child domain: VAR_SSO, EROD, SNOALB, FRC_URB2D.


weiwangncar:
It is probably from Shanghai Typhoon Institute. Jimy can confirm.

davegill:
@dudhia 
Jimy,
Do you know the individual that found this problem? We'd like to give them credit.

dudhia:
Yes, it was at Shanghai.
Actually I found the problem after telling them to test topo_wind in their
1-way nested run.
Their results showed the problem,
Jimy

davegill:
@smileMchen 
Ming,
I think this one is almost ready. The only piece I want is for you add a sentence about how you located the fields that needed modification in the Registry. That way people know that you found all of the problem fields.

smileMchen:
@davegill 
Dave, 
I have added a few lines in the commit message to explain what we did. Please let me know you need more information.

davegill:
@smileMchen 
Ming,
Looks good
Dave

PR #720, opened 2018-12-05
Adding Apache 2.0 License

kkeene44:
TYPE: text only

KEYWORDS: license, apache-2.0,

SOURCE: internal

DESCRIPTION OF CHANGES: This license is for open source code. It is suitable for research, academia, and commercial use of WRF. It allows the source code to be openly available, but does not endorse its usage. All contributors are also bound by this license.

LIST OF MODIFIED FILES: 
A    LICENSE

TESTS CONDUCTED: none

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
+      outstanding shares, or (iii) beneficial ownership of such entity.
+
+      "You" (or "Your") shall mean an individual or Legal Entity
+      exercising permissions granted by this License.
+
+      "Source" form shall mean the preferred form for making modifications,
+      including but not limited to software source code, documentation
+      source, and configuration files.
+
+      "Object" form shall mean any form resulting from mechanical
+      transformation or translation of a Source form, including but
+      not limited to compiled object code, generated documentation,
+      and conversions to other media types.
+
+      "Work" shall mean the work of authorship, whether in Source or
+      Object form, made available under the License, as indicated by a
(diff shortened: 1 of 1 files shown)


davegill:
@llpcarson @duda @kkeene44 
Folks,
Are you OK with this recognizable license for WRF and WPS?

LICENSE
```diff
@@ -0,0 +1,201 @@
...
+      License. However, in accepting such obligations, You may act only
+      on Your own behalf and on Your sole responsibility, not on behalf
+      of any other Contributor, and only if You agree to indemnify,
+      defend, and hold each Contributor harmless for any liability
+      incurred by, or claims asserted against, such Contributor by reason
+      of your accepting any such warranty or additional liability.
+
+   END OF TERMS AND CONDITIONS
+
+   APPENDIX: How to apply the Apache License to your work.
```
davegill: @kkeene44 @duda @llpcarson 
Do we need to do something here?

llpcarson:
Kelly - We use a much simpler reference to the Apache 2.0 license for some of our other work - attached here.  Might be more straightforward?

Also, generally, with the Apache license, one would also include a Copyright clause (which is often included in every file).  With wide-ranging contributors, I'm not sure if you would use a simple UCAR copyright?
[license.txt](https://github.com/wrf-model/WRF/files/2650030/license.txt)

I don't have any objection to the Apache license in general, but copyright issues might be tricky going from public domain to licensed open source!

LICENSE
```diff
@@ -0,0 +1,201 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright [yyyy] [name of copyright owner]
```
dmey: This should be filled in

llpcarson: Agree completely with the copyright info - my example was from a different project (not-WRF), so had the NOAA/UCAR/CIRES/etc....  whatever is appropriate for WRF should be included!

davegill:
@dmey 
Thanks for the correction on the intended person for the comment.

davegill:
@dmey 
Thanks for the correction on the intended person for the comment.

powersjg:
Kelly:
Just a request for an edit to the description of changes.
Delete: "It maintains our copyright."  The reason is that we don't have a copyright on WRF;
we declined to pursue copyrighting early on. As open source code, WRF is free for all to copy and redistribute.

Thanks,
Jordan

kkeene44:
@powersjg Thanks for pointing that out. I made the modification.

@davegill @weiwangncar  @llpcarson 
I don't think we want something that requires including a copyright clause in every file - especially since, as was pointed out above, we don't have a copyright on WRF. Perhaps something more like the MIT license would have more relevancy to our code. 

@mgduda 
Based on what you guys do for MPAS, do you have any suggestions?

davegill:
@kkeene44 
Several people have mentioned a simpler license, which I am inclined to lean towards.

PR #724, opened 2018-12-07
Additional WRF/MPAS unification to reduce if-defs for MPAS-specific n…

kkeene44:
TYPE: no impact

KEYWORDS: pbl, ysu, wrf, mpas, unify

SOURCE: internal

DESCRIPTION OF CHANGES: 
These modifications update the previously merged mods to unify the YSU PBL scheme with MPAS code. In an effort to reduce the differences between the necessities of the 2 models, MPAS now computes dry hydrostatic pressure prior to the call to the ysu scheme, and therefore it's no longer done inside this routine. Additionally, the exchange coefficients for MPAS have been updated to adopt the names used in the WRF model, so the if-defs/optional code for MPAS-specific names have been removed.

LIST OF MODIFIED FILES: 
M     phys/module_bl_ysu.F
M     phys/module_pbl_driver.F

TESTS CONDUCTED: Verified that both WRF/MPAS compile and run okay. WRF output is bit-for-bit identical before/after the mods.

### DIFF
--- a/phys/module_bl_ysu.F
+++ b/phys/module_bl_ysu.F
@@ -1,7 +1,12 @@
 !=================================================================================================================
-!module_bl_ysu.F was modified to integrate MPAS-specific code 
-!integration took place between WRF Master Repository (from 28Feb2018) and MPAS HWT Repository (from 27Feb2018)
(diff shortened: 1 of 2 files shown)


weiwangncar:
I'm ok with the changes.

kkeene44:
@weiwangncar 
Would you like to approve this commit, or do you still want to hear from others?

PR #732, opened 2018-12-14
LES fix for use_theta_m=1

dudhia:
TYPE: bug-fix

KEYWORDS: LES option, use_theta_m=1, V4.0 bug in surface flux fixed, calculate_N2 minor change

SOURCE: internal plus Xu Zhang (Shanghai Typhoon Institute)

DESCRIPTION OF CHANGES: 
When use_theta_m was updated to V4.0, diffusion was also applied to theta_m. However the surface flux was not modified to include a QFX term. This leads to a cooler PBL than in V3.9 with use_theta_m=1 and is incorrect. The fix is to add a QFX term that restores temperatures to correct values.

ISSUE: none posted

LIST OF MODIFIED FILES:
M       dyn_em/module_diffusion_em.F
M       dyn_em/module_first_rk_step_part2.F

TESTS CONDUCTED: 
No WTF yet. 
LES tests with fix have corrected the problem and give use_theta_m=1 results more like V3.9.1.
Theta profile plots show the cool anomaly with the old use_theta_m=1 (green) code compared to both the fixed version (red) and and use_theta_m=0 (blue). The fixed code shows more similarity with V3.9 too (not shown). The U profile shows more difference in the fixed code from the use_theta_m=0 case.
profile

RELEASE NOTE: 
LES options in V4.0 with use_theta_m=1 (default) contained an error with surface fluxes leading to a cooler PBL. QFX term has been added to surface theta_m flux to correct this problem.

### DIFF
--- a/dyn_em/module_diffusion_em.F
+++ b/dyn_em/module_diffusion_em.F
@@ -1637,7 +1637,7 @@ END SUBROUTINE cal_dampkm
       ELSE
         BN2(i,k,j) = g * ( (theta(i,k+1,j) - theta(i,k-1,j) ) /  &
                      theta(i,k,j) / tmpdz +  &
-                     0.61 * ( moist(i,k+1,j,P_QV) - moist(i,k-1,j,P_QV) ) / &
+                     1.61 * ( moist(i,k+1,j,P_QV) - moist(i,k-1,j,P_QV) ) / &
                      tmpdz -   &
                      ( tmp1(i,k+1,j) - tmp1(i,k-1,j) ) / tmpdz )
       ENDIF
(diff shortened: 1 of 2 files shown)


weiwangncar:
Should this be two commits?

dyn_em/module_diffusion_em.F
```diff
@@ -1637,7 +1637,7 @@ SUBROUTINE calculate_N2( config_flags, BN2, moist,           &
       ELSE
         BN2(i,k,j) = g * ( (theta(i,k+1,j) - theta(i,k-1,j) ) /  &
                      theta(i,k,j) / tmpdz +  &
-                     1.61 * ( moist(i,k+1,j,P_QV) - moist(i,k-1,j,P_QV) ) / &
+                     0.61 * ( moist(i,k+1,j,P_QV) - moist(i,k-1,j,P_QV) ) / &
```
davegill: @dudhia @weiwangncar 
Jimy,
I agree with Wei that these three N^2 corrections are separate from the vertical diffusion fix.

wgustafson:
Hi @dudhia and @davegill, Quick question for clarification. Does this patch apply to pre-V4 when one has theta_m turned on in the namelist? Or, does this only apply for v4+ with theta_m turned on? I am double checking to see if I need to backport this patch if I am using theta_m with v3.9.

Thanks, Bill

dudhia:
Pre-V4 was OK because the dynamics including diffusion and LES was using
regular theta even with theta_m turned on.
The problem was spotted in a comparison of V3.9 with V4.0. The new code now
diffuses theta_m so it needs the moisture flux
as part of its surface flux.

dudhia:
...to clarify, the dynamics switched to theta_m after the diffusion and
used it for the advection and small steps.

wgustafson:
Thanks!

PR #736, opened 2018-12-14
fix restart problem

weiwangncar:
YPE: bug fix
  
KEYWORDS: diff_6th_opt, restart

SOURCE: internal, reported by J. Fidel González Rouco Instituto de Geociencias (UCM-CSIC) of Spain

DESCRIPTION OF CHANGES:
When diff_6th_opt is turned on, restart fails in producing bit-for-bit results in real-data (reported) as well as idealized cases. This is due to the start and end indices too close to the domain boundaries when doing 6th order horizontal diffusions. This is a bug since this option was put in in V3.1.

LIST OF MODIFIED FILES: list of changed files:
M dyn_em/module_big_step_utilities_em.F

TESTS CONDUCTED:
No regression test yet. But tested with real-data case, quarter_ss case with periodic and open boundary conditions, and b_wave case with symmetric boundary condition in y.

RELEASE NOTE:
Fixed a restart problem with diff_6th_opt option. This was a bug since this option was added to the model in V3.1. Thanks for J. Fidel González Rouco of Spain for reporting the problem.

### DIFF
--- a/dyn_em/module_big_step_utilities_em.F
+++ b/dyn_em/module_big_step_utilities_em.F
@@ -6301,8 +6301,9 @@ END SUBROUTINE set_tend
 
 !------------------------------------------------------------------------------
 ! Begin: Assign limits of spatial loops depending on variable to be diffused.
-! The halo regions are already filled with values by the time this subroutine
-! is called, which allows the stencil to extend beyond the domains' edges.
+! The halo regions outside the physical domain are not defined, hence the loop
+! indices should start 3 rows and columns in for specified boundary conditions.
+! Also modified to work with idealized boundary conditions.
 
     ktf = MIN( kte, kde-1 )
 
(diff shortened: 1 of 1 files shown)


dyn_em/module_big_step_utilities_em.F
```diff
@@ -6312,6 +6313,24 @@ SUBROUTINE sixth_order_diffusion( name, field, tendency, MUT, dt,  &
       i_end   = ite
       j_start = jts
       j_end   = MIN(jde-1,jte)
+      IF ( config_flags%open_xs ) THEN
+      i_start = MAX(its,ids+3)
```
davegill: @weiwangncar 
Wei,
Looking at the rest of the routine, there is indenting of either 2 or 3 spaces. Go ahead and indent the assignments inside of your IF tests.

davegill:
@weiwangncar 
Wei,
For the testing, would you explicitly say that after this mod we now do get bit-for-bit with restarts with the diff6th option. 

For the release notes, I think it is important to explain what a restart bug actually is. Before this mod and when using diff6th_opt, a restarted simulation did not give bit-for-bit identical answers for cases with _xxxxx_ lateral boundary conditions.

dudhia:
I think this looks OK. The source was probably Jorge Navarro,
Jimy

dudhia:
release note needs to say Jorge Navarro too.

PR #737, opened 2018-12-17
Fix N^2 calculation (1.61->0.61) for LES

dudhia:
TYPE: bug fix

KEYWORDS: Stability N2 in LES

SOURCE: internal

DESCRIPTION OF CHANGES: 
calculate_N2 routine contains 1.61 instead of 0.61 when calculating moist version of N2. This has been since the beginning. Extremely small effect on results. Affects both use_theta_m=0 and 1.

ISSUE: no issue reported

LIST OF MODIFIED FILES: 
M       dyn_em/module_diffusion_em.F

TESTS CONDUCTED: 
No WTF. Compare use_theta_m=0 before and after (see fig of theta difference in LES case after 4 hours).

screen shot 2018-12-17 at 9 32 13 am
RELEASE NOTE: 
LES calculation of N2 slightly corrected. Very small effect.

### DIFF
--- a/dyn_em/module_diffusion_em.F
+++ b/dyn_em/module_diffusion_em.F
@@ -1637,7 +1637,7 @@ END SUBROUTINE cal_dampkm
       ELSE
         BN2(i,k,j) = g * ( (theta(i,k+1,j) - theta(i,k-1,j) ) /  &
                      theta(i,k,j) / tmpdz +  &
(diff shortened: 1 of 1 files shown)


dudhia:
I am reverting this N^2 fix based on new information.
I will propose a revert PR.

smileMchen:
@dudhia 
Jimy,
I just finish the regression test with your previous PR. Please let me know if I need to rerun the regression test with your new PR.

PR #738, opened 2018-12-17
Adding an official LICENSE file to the top level directory

kkeene44:
TYPE: text only

KEYWORDS: license

SOURCE: internal

DESCRIPTION OF CHANGES: This license is simply our public domain notice on our web pages. We do not have a copyright for WRF, so that is not necessary to include. This will serve as our license unless/until the time comes that we must update to a specific type required by NSF.

LIST OF MODIFIED FILES:
A LICENSE

TESTS CONDUCTED: none

RELEASE NOTE: No note necessary.


davegill:
@kkeene44 
Kelly,
I am good with this. Let's allow others to comment for a day or two. Basically, it is what we have had public for since forever, so I do not see what could be an issue.

dmey:
@kkeene44 this looks good but is there any reason for not using a more informative file extension -- e.g. `LICENSE.txt`?

kkeene44:
@dmey 
No, there wasn't really a good reason. We were just following what another modeling group in our department did, but it doesn't really matter to me. I updated the file name to include the .txt extension.

PR #746, opened 2019-01-02
Fix mismatch isltyp and landmask 

weiwangncar:
TYPE: bug fix

KEYWORDS: surface_input_source, isltyp, landmask

SOURCE: internal (reported by many users, including Priscilla Mooney from Norway)

DESCRIPTION OF CHANGES: 
To accommodate users who would like to change _LU_INDEX_ after running geogrid program, we changed namelist surface_input_source from default value of 1 (recomputing dominant categories in real) to 3 (using dominant categories from geogrid) in v3.8. 

Unfortunately, we neglected to consider in the v3.8 fix that when dominant categories for land and soil are recomputed in real, the real program also does mismatch checks: ensuring that the lower resolution _ISLTYP_ matches with _LANDMASK_ data. This check in real has two two effects:
1. the actual _ISLTYP_, _IVGTYP_, and _XLAND_ can be different from that from surface_input_source = 1 (a check later in real program effectively matched _IVGTYP_ based on _ISLTYP_); 
2. for certain cases, the real program would stop due to un-matched _ISLTYP_ and P_. 

This PR fixes the problem of this missing consistency check in the real program when surface_input_source = 3. The _ISLTYP_, _IVGTYP_, and _XLAND_ are now identical to those coming from real after setting surface_input_source = 1. 

Below, the right image shows the wrong _XLAND_ field from the wrong code, and left image shows the correct _XLAND_ field after the fix. These images are created on a 1 km grid.

before

LIST OF MODIFIED FILES: 

TESTS CONDUCTED: 
1. Ran real before and after the change and results are correct. 
2. Ran real for a case that failed before and now it succeeds (thanks to Kelly).

RELEASE NOTE: 
In V3.8, we changed default surface_input_source option value from 1 to 3, which uses the dominant categories computed in geogrid. But we neglected that when dominant categories for land and soil are recomputed in real, it also does mismatch checks, making sure the lower resolution isltyp matches with landmask data. This has two two effects: one is the actual isltyp, ivgtyp and xland can be different from that from surface_input_source = 1 (a check later in real program effectively matched ivgtyp based on isltyp); and the other effect is that for certain cases, the real program would stop due to un-matched isltyp and ivgtyp. This change fixes this problem, and isltyp, ivgtyp, and xland are now identical to those coming from surface_input_source = 1.

### DIFF
--- a/dyn_em/module_initialize_real.F
+++ b/dyn_em/module_initialize_real.F
@@ -3002,7 +3002,7 @@ integer::oops1,oops2
          change_soilw = 0
          DO j = jts, MIN(jde-1,jte)
             DO i = its, MIN(ide-1,ite)
-!              IF ( skip_middle_points_t ( ids , ide , jds , jde , i , j , em_width , hold_ups ) ) CYCLE
+               IF ( skip_middle_points_t ( ids , ide , jds , jde , i , j , em_width , hold_ups ) ) CYCLE
                   IF ( grid%landmask(i,j) .GT. 0.5 .AND. grid%isltyp(i,j) .EQ. grid%isoilwater ) THEN
                      grid%isltyp(i,j) = 8
                      change_soilw =  change_soilw + 1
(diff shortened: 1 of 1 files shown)


davegill:
@weiwangncar 
Wei,
This looks good to me.

PR #750, opened 2019-01-03
Make 3DVAR and hybrid-EnVar work with moist potential T

liujake:
TYPE: enhancement

KEYWORDS: Moist Theta, 3DVAR, hybrid-3D/4DEnVar

SOURCE: internal

DESCRIPTION OF CHANGES:
1. Moist theta is the default option (use_theta_m=1) since WRF V4.0 (PR #443, #524, #621), but did NOT work for WRFDA. This change is to make WRFDA's 3DVAR and hybrid-3D/4DEnVar work with moist theta. WRFPlus code needs to be modified to make 4DVAR work with moist theta in the future.

2. WRFDA's analysis variable is temperature. Dry theta perturbation (i.e., T in WRF NC file) is read in by WRFDA and converted to temperature (in da_transfer_wrftoxb step). Now WRFDA updates also moist theta perturbation (i.e., THM in WRF NC file) in addition to update dry theta perturbation in da_update_firstguess step after 3DVAR or hybrid-EnVar analysis. The formulation of converting T to THM is: THM = ( T + T0 ) * [ 1. + (R_v/R_d) * Qv ] - T0, where T0=300.

3. Additional check of global attribute "USE_THETA_M" in WRF NC file is added when reading the background file.

4. Note that wrfbdy file includes only the variable T, which represents either dry or moist theta perturbation according to USE_THETA_M=0 or 1. 

LIST OF MODIFIED FILES:
M    var/da/da_main/da_med_initialdata_input.inc
M    var/da/da_main/da_update_firstguess.inc

TESTS CONDUCTED:
1. WRFDA regression tests passed.
2. A cycling case (cycle_sens_hires) of WRFDA regtests with use_theta_m=1, hybrid_opt=1, etac=0.2 worked well. Similar pattern of T and THM analysis increment is obtained. Cycling result of new setting is reasonably close to the old result with use_theta_m=0, hybrid_opt=0. Difference is expected. Benchmark output is updated for this cycling case for future regtests.

RELEASE NOTE: 3DVAR and hybrid-3D/4DEnVar now work with moist potential temperature.


jjguerrette:
This looks good to me as soon as that 3D-Var test is completed.

liujake:
@jjguerrette @davegill Dave and JJ, I made tests for this PR. Can one of you approve it so that I can put it in. Jake

var/da/da_main/da_update_firstguess.inc
```diff
@@ -16,7 +16,8 @@ subroutine da_update_firstguess(grid, out_filename)
 !    grid%w_2
 !    grid%mu_2
 !    grid%ph_2
-!    grid%t_2
+!    grid%t_2 (dry potential temperature perturbation)
```
davegill: @liujake 
Jake,
In the WRF model, grid%t_2 is either moist or dry theta perturbation, depending on the use_theta_m switch. The grid%th_phy_m_t0 is always the dry theta perturbation. The THM name is a character string of the input / output field for the grid%t_2 variable. 

Inside of the DA Registry, grid%t_2 is always the dry theta perturbation.

For this comment, we should be careful about referring to the internal field (such as grid%t_2) and the string that is used to name the field in the input / output file.

var/da/da_main/da_update_firstguess.inc
```diff
@@ -585,17 +588,37 @@ subroutine da_update_firstguess(grid, out_filename)
...
+      call ext_ncd_get_var_info (dh1,trim(rmse_var),ndim1,ordering,staggering, &
+           start_index,end_index1, WrfType, ierr    )
+
+      if (grid%use_theta_m == 1) then ! convert dry theta perturbation to moist one
+         write(unit=message(1),fmt='(A, I2)') 'convert T to THM when use_theta_m = ', grid%use_theta_m
+         call da_message(message(1:1))
+         do k= 1,nsig_regional
+            do j= 1,nlat_regional
+               do i= 1,nlon_regional
+                 qvf = 1.+(R_v/R_d)*grid%moist(i,j,k,p_qv)
```
davegill: @liujake 
Jake,
Is grid%moist the size of the computational region that is decomposed on a particular MPI rank?  Are the n{lat,lon}_regional dimensions just the computational size without halo zones?

davegill: @liujake 
Jake,
The write seems to be using "if(rootproc)" as if the whole domain is output, so I am confused about the moist dimensions.

Also, I looked in this routine, and you don't do computations anywhere else. Perhaps the moist theta computation should not be done in this output routine? I am not a DA person, so I am not familiar with the conventions.

var/da/da_main/da_update_firstguess.inc
liujake: @davegill I added your text in the subroutine as a note to clarify this.

var/da/da_main/da_update_firstguess.inc
liujake: @davegill Thanks for catching this! You are right. n{lat,lon}_regional dimensions are domain size. I adjusted qvf calculation right after patch_to_global for QVAPOR and update THM before updating QVAPOR. This way, I do not have to define another 3D field array or collect QVAPOR twice. Test passed.

liujake:
@davegill Let me know whether these changes addressed your concerns and Ok to push them in.

davegill:
Jake
Thanks for addressing this memory overwrite issue. I'm good with this PR

PR #760, opened 2019-01-11
Fix broken NMM builds, introduced by PR #749

weiwangncar:
TYPE: bug fix

KEYWORDS: NMM compile, #749

SOURCE: internal

DESCRIPTION OF CHANGES: 
Merge #749 (86f7675) broke NMM compile. This PR fixes NMM builds by moving code that references the ARW-only variable `model_config_rec % nwp_diagnostics` into a section that is only active if `(EM_CORE == 1)`. This change also moves the code from subroutine set_physics_rconfigs to check_nml_consistency. It seems appropriate to place the check there.

LIST OF MODIFIED FILES: list of changed files:
M  share/module_check_a_mundo.F

TESTS CONDUCTED: 
NMM compiles now! Tested with ARW too with expected outcome.


mgduda:
The important part of the fix seems to be the fact that the setting of `model_config_rec % do_radar_ref = 1` happens inside of a `#if (EM_CORE == 1)` section. Is there a reason why it's better to move the setting of this `model_config_rec` option from the `set_physics_rconfigs` subroutine to the `check_nml_consistency` subroutine?

weiwangncar:
@mgduda Yes.

mgduda:
The history of this branch is a bit odd: it contains four commits, two of which appear to have previously been merged to `develop`. Since the WRF practice is to "squash and merge", this probably doesn't make any difference, though.

mgduda:
@weiwangncar Can you provide a more descriptive title for this PR? Could you also add the following details to the description section: This PR fixes NMM builds by moving code that references the ARW-only variable `model_config_rec % nwp_diagnostics` into a section that is only active if `(EM_CORE == 1)`.

mgduda:
I opened an issue (#761) for the build errors with `module_cu_mskf.F`.

PR #764, opened 2019-01-25
MYNN PBL (only) update

joeolson42:
TYPE: Bug fix, code trim, new feature (dissipative heating - small impact), code sync with FV3

KEYWORDS: MYNN, PBL, mass-flux, subgrid clouds, HRRR, FV3

SOURCE: Joe Olson (NOAA-GSD) and Jaymes Kenyon (NOAA-GSD)

DESCRIPTION OF CHANGES: 
            Big improvements in downward SW radiation due to revision of subgrid-scale (SGS) clouds              
                - better cloud fraction and subgrid scale mixing ratios.                             
                - may experience a small cool bias during the daytime now that high                  
                  SW-down bias is greatly reduced...                                                 
            Some tweaks to increase the turbulent mixing during the daytime for                      
                bl_mynn_mixlength option 2 to alleviate cool bias (very small impact).               
            Improved ensemble spread from changes to SPP in MYNN                                     
                - now perturbing eddy diffusivity and eddy viscosity directly                        
                - now perturbing background rh (in SGS cloud calc only)                              
                - now perturbing entrainment rates in mass-flux scheme                               
            Added IF checks (within IFDEFS) to protect "mixchem" code from being used                  
                when HRRR smoke is used (no impact on regular non-wrf chem use)                      
            Important bug fix for wrf chem when transporting chemical species in MF scheme           
            Removed 2nd mass-flux scheme (now only bl_mynn_edmf = 1, no option 2)                     
            Removed unused stochastic code for mass-flux scheme                                      
            Changed mass-flux scheme to be integrated on interface levels instead of                 
                mass levels - impact is small                                                        
            Added option to mix 2nd moments in MYNN as opposed to the scalar_pblmix option.          
                - activated with bl_mynn_mixscalars = 1; this sets scalar_pblmix = 0
                - this mixes the second moments and aerosols within the MYNN, making the MYNN portable
                    to other models (like FV3) because it doesn't have to rely upon external WRF code.            
                - added tridiagonal solver used in scalar_pblmix option to duplicate tendencies       
                - this alone changes the interface call considerably from v4.0.                      
            Slight revision to TKE production due to radiation cooling at top of clouds              
            Added the non-Guassian buoyancy flux function of Bechtold and Siebesma (1998, JAS).      
                - improves TKE in SGS clouds                                                         
            Added heating due to dissipation of TKE (small impact, maybe + 0.1 C daytime PBL temp)   
            Added QNWFA and QNIFA into the MYNN for mixing when run in tandem with the Thompson 
                 Aerosol-aware microphysics scheme. Only mixed when bl_mynn_mixscalars = 1.
            Misc changes made for FV3/MPAS compatibility

LIST OF MODIFIED FILES: 
M       phys/module_bl_mynn.F
M       phys/module_physics_addtendc.F

TESTS CONDUCTED: 
 - [x] Small regression test OK on cheyenne (GNU only, all builds, single case per build, serial / openmp / mpi constructed and b4b).
 - [x] This code has been exercised in the experimental RAP and HRRR (is a leading candidate for RAPv5, HRRRv4) and FV3.

RELEASE NOTE: Although the set of changes is large, for most variables/metrics, the improvement is small. The biggest improvement is the reduction in the downward shortwave radiation bias:
screen shot 2019-01-25 at 12 59 11 pm


dyn_em/module_first_rk_step_part1.F
```diff
@@ -952,6 +952,8 @@ SUBROUTINE first_rk_step_part1 (   grid , config_flags              &
      &        ,QI_CURR=moist(ims,kms,jms,P_QI), F_QI=F_QI                 &
      &        ,QS_CURR=moist(ims,kms,jms,P_QS), F_QS=F_QS                 &
      &        ,QG_CURR=moist(ims,kms,jms,P_QG), F_QG=F_QG                 &
+     &        ,QNIFA_CURR=scalar(ims,kms,jms,P_QNIFA),F_QNIFA=F_QNIFA          &
+     &        ,QNWFA_CURR=scalar(ims,kms,jms,P_QNWFA),F_QNWFA=F_QNWFA          &
```
weiwangncar: @joeolson42 Can you line up &?

phys/module_physics_addtendc.F
```diff
@@ -714,12 +714,13 @@ SUBROUTINE phy_bl_ten(config_flags,rk_step,n_moist,n_scalar,     &
...
-               ids,ide, jds, jde, kds, kde,                     &
-               ims, ime, jms, jme, kms, kme,                    &
-               its, ite, jts, jte, kts, kte                     )
+!Note: QNC and QNI are updated in the pbl driver
+!        if (P_QT .ge. PARAM_FIRST_SCALAR)                                         &
+!           CALL add_a2a(scalar_tendf(ims,kms,jms,P_QT),RQNIBLTEN,&
+!               config_flags,                                    &
+!               ids,ide, jds, jde, kds, kde,                     &
+!               ims, ime, jms, jme, kms, kme,                    &
+!               its, ite, jts, jte, kts, kte                     )
```
weiwangncar: @joeolson42 If this is commented out, should it be just removed? The same for lines 748 - 754.

phys/module_pbl_driver.F
```diff
@@ -823,6 +835,8 @@ SUBROUTINE pbl_driver(                                          &
   flag_qg = .FALSE. ; IF ( PRESENT( F_QG ) ) flag_qg = F_QG
   flag_qnc = .FALSE. ; IF ( PRESENT( F_QNC ) ) flag_qnc = F_QNC !Used in CAMUWPBL
   flag_qni = .FALSE. ; IF ( PRESENT( F_QNI ) ) flag_qni = F_QNI !Used in CAMUWPBL
+  flag_qnwfa = .FALSE. ; IF ( PRESENT( F_QNWFA ) ) flag_qnwfa = F_QNWFA
+  flag_qnifa = .FALSE. ; IF ( PRESENT( F_QNIFA ) ) flag_qnifa = F_QNIFA
```
davegill: This seems to be a good way to grab the value from an optional variable, and then this flag can be safely used in the mynn_bl_driver call.

Registry/Registry.EM_COMMON
```diff
@@ -2140,12 +2140,12 @@ rconfig   integer     bl_mynn_tkebudget   namelist,physics      max_domains    0
...
 rconfig   integer     bl_mynn_edmf        namelist,physics      max_domains    1       irh      "bl_mynn_edmf"                  "0:off,1:activate mass-flux in mynn"                    ""
-rconfig   integer     bl_mynn_edmf_mom    namelist,physics      max_domains    0       irh      "bl_mynn_edmf_mom"              "0:off,1:activate mass-flux transport of momentum"      ""
+rconfig   integer     bl_mynn_edmf_mom    namelist,physics      max_domains    1       irh      "bl_mynn_edmf_mom"              "0:off,1:activate mass-flux transport of momentum"      ""
 rconfig   integer     bl_mynn_edmf_tke    namelist,physics      max_domains    0       irh      "bl_mynn_edmf_tke"              "0:off,1:activate mass-flux transport of tke"           ""
-rconfig   integer     bl_mynn_edmf_part   namelist,physics      max_domains    0       irh      "bl_mynn_edmf_part"             "0:off,1:activate areal partitioning of ed and mf"      ""
+rconfig   integer     bl_mynn_mixscalars  namelist,physics      max_domains    0       irh      "bl_mynn_mixscalars"            "0:off,1:activate mixing of scalars (qnx, qnxfa) in MYNN"      ""
```
davegill: Can we assume that no one else would have set bl_mynn_edmf_part in a namelist? If this is largely an "internal developer" switch that nobody else would know about or use, then I am OK with the re-name (without some version overlap and helpful comments in the check_a_mundo).

phys/module_physics_addtendc.F
```diff
@@ -714,20 +714,6 @@ SUBROUTINE phy_bl_ten(config_flags,rk_step,n_moist,n_scalar,     &
                 ims, ime, jms, jme, kms, kme,                    &
                 its, ite, jts, jte, kts, kte                     )
 
-        if (P_QT .ge. PARAM_FIRST_SCALAR)                                         &
```
davegill: For both P_QT and below P_QNI, since these calls are deleted, are these only ever used by MYNN

davegill:
@joeolson42 
Joe,
Change the location where you are proposing to merge this code change from "master" to "develop". Just click the "Edit" button at the top of this PR's pull page, and then use the pull downs to change from master to develop.

davegill:
@joeolson42 
Joe,
I do not explicitly see any mention of the QNIFA QNWFA fields that are now passed into your scheme. Would you add something into your commit message.

dyn_em/module_first_rk_step_part1.F
joeolson42: now aligned and code committed.

Registry/Registry.EM_COMMON
joeolson42: Yes, or if they did, the users were wasting their time. The flag bl_mynn_edmf_part partitioned the eddy diffusivity (ED) and mass-flux (MF) by reducing the ED when the MF scheme was active. This was done by assuming there is no ED in grid cell where the MF is active. This is a poor assumption, but it had very little or no impact, so as I saw it, it was a waste of electricity and needed to be removed.

Thus, bl_mynn_edmf_part was replaced with bl_mynn_mixscalars, which serves a completely different function: to mix the second moments and aerosols within the MYNN. This makes the MYNN portable to other models (like FV3) because it doesn't have to rely upon the code associated with the scalar_pblmix option in WRF.

I will add more detail in the commit description to clarify this.

phys/module_pbl_driver.F
joeolson42: Pardon my confusion Dave, but is this just a comment? If so, thank you. If not, thanks anyway.

phys/module_physics_addtendc.F
joeolson42: I see for CAMUWPBLSCHEME and CAMZMSCHEME, there is also code for mixing QNI, but I don't think there needs to be if scalar_pblmix = 1.  I wonder if there is a double-counting if scalar_pblmix set to 1 or if the scalar_tendf array is just written over (I think the latter). If QNI is mixed non-locally in CAMUWPBLSCHEME and CAMZMSCHEME, perhaps the way the developers prefer it. Its probably a small impact either way.

I did a quick check for "QNI" in check_a_mundo and didn't see a check to make sure scalar_pblmix = 0 when running either of these two schemes. So, I think there is a potential to mix QNI twice, but it probably just overwrites the tendencies from the scalar_pblmix code, not resulting in a double-counting.

phys/module_pbl_driver.F
davegill: @joeolson42 
Joe,
Yes, just a comment on recognizing what you have to do to use these fields that are both optional and my not be set even if they are present.
You're welcome

davegill:
@joeolson42 
Joe,
Yes, I am OK with the structural aspects of this PR.

@dudhia @weiwangncar 
Wei and Jimy,
Would you please review this PR.

davegill:
@joeolson42 
Joe,
The small regtest results look good for ARW. I am re-running to check out NMM and HWRF.

weiwangncar:
I'm ok with the change.

PR #766, opened 2019-01-29
BF full landread.c file should not be compiled by default

davegill:
TYPE: bug fix

KEYWORDS: landread.c, LANDREAD_STUB, TERRAIN_AND_LANDUSE, moving nest

SOURCE: Found by Kevin Manning

DESCRIPTION OF CHANGES: 
Problems:
1.  This code used to not be compiled unless a vortex following build with additional env vars were set.
2. A compilation problem in landread.c is present with the latest Fedora/gfortran combo

Solution:
The share/landread.c file has a large surrounding ifdef that is supposed to allow only the stub portion of the code to compile by default. This correct, conditional compilation relied on two items, neither of which were working:
1. The `postamble` file needed a line, `CONFIGURE_TERRAIN_AND_LANDUSE`, for substitution by a perl script into either of two expected values. This line was not present.
2. The `Config.pl` needed to replace the line, `CONFIGURE_TERRAIN_AND_LANDUSE`, with either `-DTERRAIN_AND_LANDUSE` (for a moving nest with high-res topo and LU for the moving domains), or `-DLANDREAD_STUB`. The stub option was missing.
3. The modification effectively bypasses the troublesome code, parts of landread.c that reference include files from rpc (which have apparently been moved/reorganized in recent glibc updates).

LIST OF MODIFIED FILES: 
arch/postamble
arch/Config.pl

TESTS CONDUCTED: 
 - [x] Passed small regtest GNU, Intel, PGI on cheyenne
Legend:
 * GNU 15, 16, 17 refer to using the GNU compiler (build option, not version!) on a Linux machine. 15, 16, 17 refer to the first argument used for configure, and are the serial, OpenMP, and MPI builds, respectively.
 * Nest Opt 0, 1, 2, 3 refer to the second setting that a user supplies for configure. 0, 1, 2, 3 refer to no nest, standard nest, specified moves, TC tracking.
 * The -DTERRAIN_AND_LANDUSE column refers to the env variable set when doing a moving nest to bring in optional high-res topo and LU
 * All expected builds are successfully completed with an OK.
 * The configurations that are not permitted are denoted by XXXXX.

|     Nest Opt   | Serial (15)      | OpenMP (16) | MPI (17)    | -DTERRAIN_AND_LANDUSE |
|  ------------ | ------------- | ------------- | ---------- | --------------------------- |
|   0                  |    OK               |       OK            | XXXXX      |   XXXXX                                |
|   1                  |    OK                |       OK            |  OK            |   XXXXX                                |
|   2                  |  XXXXX           |     XXXXX      |  OK            |   XXXXX                                |
|   3                  |   XXXXX          |      XXXXX     |  OK            |   OK                                       |


weiwangncar:
In the past, we provided a special file landread.c.dist every time we release the code. Then this becomes not necessary. I believe the code can always be compiled, but the default option is not to read from the high-resolution dataset.

davegill:
@weiwangncar 
Wei,
Correct. We used to provide the landread.c.dist so that the capabilities of the high-res topo and LU were kept for publication purposes. That "private vs public" file trick is no longer necessary. Since we have a single file, share/landread.c, we needed to clean up the build process a bit.

davegill:
@kwman44 
Kevin,
Would you please take a look at this PR and see if this fixes the rpc trouble that you noticed on your Red Hat machine.

kwman44:
This PR solves the compilation problem on my system with the latest Fedora/gfortran combo.
It effectively bypasses the troublesome code, parts of landread.c that reference include files from rpc (which have apparently been moved/reorganized in recent glibc updates).

PR #786, opened 2019-02-13
Remove Goddard radiation nested OpenMP regions

davegill:
TYPE: bug fix

KEYWORDS: radiation, Goddard, OpenMP

SOURCE: internal

DESCRIPTION OF CHANGES: 
Problem:
An OpenMP failure (invalid access) as soon as the model enters the the Goddard radiation code 
(the same top-level routine handles both the LW and SW schemes).

Solution:
The radiation driver has a large OpenMP loop. The Goddard radiation scheme is among those 
schemes that is inside of this larger OpenMP loop. Traditionally, the WRF model does not support 
nested threaded regions. The OpenMP directives (only a single block at the top of the Goddard 
radiation scheme) are now "ifdef'ed" out. They are not removed as this radiation code is used by 
other modeling systems, and they apparently have a different threading paradigm compared to WRF 
(inside the physics scheme as opposed to outside of the physics scheme).

LIST OF MODIFIED FILES: 
M   module_ra_goddard.F

TESTS CONDUCTED: 
 - [x] Without mod, WRF cannot run to completion with the Goddard radiation scheme, using OpenMP.
 - [x] With mod, WRF is able to run simulation to completion, using OpenMP.

### DIFF
--- a/phys/module_ra_goddard.F
+++ b/phys/module_ra_goddard.F
@@ -1503,26 +1503,28 @@ data ((mcdat(ilev,10,ifld),ifld=1,6),ilev=12,22)/  &
  endif
 
 ! Removing these OpenMP directives since this is already in a 
-! threaded region from the radiation driver. 
+! threaded region from the radiation driver for the WRF model.
 ! Perhaps other models use these, so we will keep them around.
-!!$OMP PARALLEL DO &
-!!$OMP PRIVATE ( ic, ii) &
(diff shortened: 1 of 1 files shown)


dudhia:
could also consider ifdef or #if 0 for easier reactivation.

davegill:
@dudhia 
Jimy,
Hmmm, I think I do prefer your idea of using an `#if 0` or similar cpp directive-based modification instead of commenting out all of those lines.

PR #788, opened 2019-02-14
Remove extraneous HVC comments in module_check_a_mundo

davegill:
TYPE: no impact

KEYWORDS: HVC, check_a_mundo, comments

SOURCE: internal

DESCRIPTION OF CHANGES: 
An original IF block in check_a_mundo that guarded against an inconsistent build was removed
in v4.0. However, the comments ahead of that IF block remained. That block of comments, which
now describe nothing but empty lines, is removed.

LIST OF MODIFIED FILES: 
M  share/module_check_a_mundo.F

TESTS CONDUCTED: 
No tests required as these changes are only removing incorrect comments.

### DIFF
--- a/share/module_check_a_mundo.F
+++ b/share/module_check_a_mundo.F
@@ -1593,12 +1593,6 @@
      END DO
 
 !-----------------------------------------------------------------------
-!  The hybrid vertical coordinate (HVC) namelist.input option (hybrid_opt=2)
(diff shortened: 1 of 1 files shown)


kkeene44:
@davegill
I approved this, but first, do we want this to go on the release-v4.1 branch, instead of develop?

PR #789, opened 2019-02-15
Finish the quieting of "./clean -a"

davegill:
TYPE: enhancement

KEYWORDS: clean script

SOURCE: internal

DESCRIPTION OF CHANGES: 

The purpose of the `clean` script is to run through multiple directories and either remove 
files directly or to issue some `make clean` or `make superclean` command. 

In the `clean` script itself, remove `set echo`, so the script does not report each command
contained within the script: cd here, cd there, this if test, that else branch, etc.

To stop the "Entering such and such dir" and "Leaving such and such dir"
paired messages as the Makefiles traverse the source code, add the `-s` option to all 
`make clean` and `make superclean` commands. This stops make from printing a rule
before executing it. We still get error messages if they should occur. With a much smaller
output from `clean -a`, we can actually notice the error messages. We have not added the
`-s` option to any other portion of the makefile system than the `make clean` or `make superclean` options.

Remove an outright echo statement from grib2 IO.

Add a small, new makefile for CRTM, expressly for the `clean` option. This is because the full 
Makefile assumes that the configure.wrf file exists, the Makefile cannot include it (if it does not exist), 
and immediately fails the clean step in that directory.

LIST OF MODIFIED FILES:
M   clean
M   external/Makefile
M   external/io_grib_share/build/library_rules.mk
M   var/Makefile
A   var/external/crtm_2.3.0/libsrc/makefile_clean
M   var/external/crtm_2.3.0/makefile

TESTS CONDUCTED: 
 - [x] Before mods, output from `clean -a` was 60 lines; now it is quiet.
 - [x] The small amount of output pointed out an obvious error that needed to be fixed.
 - [x] Before mods, CRTM directory was not actually cleaned
 - [x] `clean -a` still reports errors 
```
/bin/rm: cannot remove ‘foo_foo_fee_doo’: No such file or directory
```
 - [x] Build still works


clean
```diff
@@ -20,7 +20,7 @@ endif
 ( cd inc ; /bin/rm -f *.inc namelist.default )
 
 
-set echo
+#set echo
```
kkeene44: @davegill 
Can we just remove this line altogether, instead of commenting it out?

external/io_grib_share/build/library_rules.mk
```diff
@@ -148,7 +148,7 @@ thisdir_clean: thisdir_clean_lib
 	@/bin/rm -f *.o *.obj *.mod *.f90 core so_locations Makefile.bak *~ #*#
 	@/bin/rm -fr ii_files
 	@if [ -d utest ] ; then \
-		echo "        Doing make clean on utest subdirectory"; \
+#		echo "        Doing make clean on utest subdirectory"; \
```
kkeene44: @davegill 
Same with this one - should we just remove the line altogether?

kkeene44:
@davegill 
Do we want this to go on the release-v4.1 branch, instead of develop?

clean
davegill: Done

external/io_grib_share/build/library_rules.mk
davegill: Done

davegill:
@kkeene44 
Kelly,
Yes, the branch should now be release-v4.1. I have changed the base branch.
I have addressed the other two concerns with the commented out lines.

PR #793, opened 2019-02-26
Use linear vertical interpolation for pos def fields

davegill:
TYPE: bug fix

KEYWORDS: positive definite, vertical interpolation, real

SOURCE: Found by Pedro Jimenez (NCAR), fixed internally

DESCRIPTION OF CHANGES:
Problem:
Users noticed that the constituent for P_QI was negative after running real.

Solution:
The order of the vertical interpolation was originally handled by the
`lagrange_order` namelist variable. By default, this value was set to
2, meaning a quadratic was used for vertical iterpolation. For positive
definite fields, we should use only linear vertical interpolation. The
results of a linear interpolation will maintain a positive definite
field. The new namelist variable `linear_interp` is always set
to 1, and is used in the calling function for all of the known positive
definite fields (for example, constituents of moist).

LIST OF MODIFIED FILES:
M Registry/Registry.EM_COMMON

TESTS CONDUCTED:
 - [x] Original moisture field, RH, stays the same.

RELEASE NOTE: For the real program, the order of the vertical interpolation option is now always set to a linear interpolation for meteorological fields that are supposed to be positive definite (for example, moist process constituents and number concentration). However, the RH field in the real program continues to use the original vertical interpolation option.

### DIFF
--- a/Registry/Registry.EM_COMMON
+++ b/Registry/Registry.EM_COMMON
@@ -2089,7 +2089,7 @@ rconfig   logical use_levels_below_ground namelist,domains	1             .true. 
 rconfig   logical use_tavg_for_tsk        namelist,domains	1             .false.  irh    "use_tavg_for_tsk"   "T/F: use diurnal avg sfc temp for tsk" ""
 rconfig   logical use_surface             namelist,domains	1             .true.   irh    "use_surface"   "T/F: use input surface level in interpolation" ""
 rconfig   integer lagrange_order          namelist,domains	1              2       irh    "lagrange_order"   "1=linear, 2=quadratic vertical interpolation, 9=cubic spline"      ""
-rconfig   integer pos_def_vert_interp     namelist,domains	1              1       irh    "pos_def_vert_interp"   "1=linear"      ""
+rconfig   integer linear_interp           namelist,domains	1              1       irh    "linear_interp"   "1=linear"      ""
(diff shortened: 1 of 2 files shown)


davegill:
@pedro-jm @dudhia @gthompsnWRF 
Folks,
Here is what I am thinking to solve this issue of negative values for positive definite fields. This is ONLY dealing with ensuring that the vertical interpolation is linear. There may be other steps required, but let's focus on this aspect first.

davegill:
@dudhia 
Jimy,
I don't like adding hard-coded mods. It would be easy to have someone introduce a positive definite interpolator, and we would want to be able to select that at run-time.

dudhia:
lagrange_order = 1 seems to be what we want, so we already have that
switch, and the
problem is related to only having one namelist option for all fields.
Jimy

davegill:
@dudhia 
Jimy,
Absolutely correct. Instead of a single option, now we have two. We have the existing option for the order of the vertical interpolation, which we will now use for all variables except for the positive definite fields. We have a new, separate option for the order of the vertical interpolation for the positive definite fields.

dudhia:
It seems this could have been achieved with changes to the meaning of the Q
var_typ.

davegill:
@weiwangncar @dudhia 
Wei and Jimy,
I have modified the PR so now the RH uses the original vertical interpolation method. There is no vertical interpolation of Qv. RH is vertically interpolated, then the Qv is diagnosed. Would you provide a review of this PR.

PR #801, opened 2019-03-07
update module_cu_ntiedtke file to identically match MPAS version

kkeene44:
TYPE: no impact

KEYWORDS: wrf, mpas, unification, ntiedtke, cumulus

SOURCE: internal

DESCRIPTION OF CHANGES: For the purpose of WRF/MPAS physics unification, we want the 2 models to have identical versions of each physics routine. After some mods were made to the MPAS version, this version then was behind. This particular PR defines t13 in the code rather using the Prandtl number since the Prandtl number is different in WRF and MPAS.

LIST OF MODIFIED FILES: 
M   phys/module_cu_ntiedtke.F

TESTS CONDUCTED: Verified that it compiles and runs fine. There were no diffs in output before/after modifications.

### DIFF
--- a/phys/module_cu_ntiedtke.F
+++ b/phys/module_cu_ntiedtke.F
@@ -49,26 +49,21 @@ module module_cu_ntiedtke
 #if defined(mpas)
      use mpas_atmphys_constants, only: rd=>R_d, rv=>R_v, &
    &       cpd=>cp, alv=>xlv, als=>xls, alf=>xlf, g=>gravity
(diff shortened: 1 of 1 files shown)


weiwangncar:
@kkeene44 Perhaps you can add in the commit message that this particular PR defines t13 in the code rather using the Prandtl number since the Prandtl number is different in WRF and MPAS.

kkeene44:
@weiwangncar 
Okay, I added that.

PR #803, opened 2019-03-09
Replace module interfaces for netcdf and mpi 

davegill:
TYPE: bug fix

KEYWORDS: netcdf, mpi, module

SOURCE: internal

DESCRIPTION OF CHANGES: 
There are two similar types of changes. Most of the WRF modeling system uses the `netcdf.inc`
file instead of `use netcdf`, and most of the WRF modeling system uses `mpif.h` instead of
`use mpi`. The Fortran module constructs are removed for consistency with the rest of the WRF
system and also to lower the bar for porting the WRF code to small Linux boxes. 

With this PR, the WRF code now builds and runs with newer GNU compilers (starting to become 
important with Fortran 2003 and Fortran 2008 constructs) but still using the default installed libraries 
on the small desktop Linux machines (where the libs are built with GNU 4.8-ish). In addition to 
assisting the user base that has access to user-managed Linux systems, these mods simplify building
containers for WRF which can be used for automated testing.

MPI Differences
When the `use mpi` construct is removed, it is replaced by a Fortran include statement 
`include mpif.h`.

NETCDF Differences
When the `use netcdf` construct is removed, it is typically replaced by a Fortran include statement 
`include netcdf.inc`. In both instances in the PR, no netcdf code was used within the scoping unit 
covered by the Fortran USE statement, so the `use netcdf` statements were removed without
any source code replacement.

LIST OF MODIFIED FILES: 
MPI
M dyn_nmm/module_HIFREQ.F
M dyn_nmm/module_initialize_real.F
M frame/module_bdywrite.F
M phys/module_ra_cam_support.F

NETCDF
M phys/module_gocart_coupling.F
M phys/module_physics_init.F

TESTS CONDUCTED: 
There were two types of modifications. 

The MPI changes required replacing the USE module construct with a Fortran include. For the MPI
changes, it is necessary to make sure that the parallel vs serial, and the before vs after results are 
the same. Those tests were conducted.
1. For ARW, a 12-h Jan 2000, single domain: ARW12
2. For ARW, a 1-h Jan 2000, nested domain: ARW01
3. For NMM, a 3-h, 3-domain run: NMM03
 - [x] ARW12 bit-for-bit: serial, MPI 1pe, MPI 3pe
 - [x] ARW01 bit-for-bit for both domains: serial, MPI 1pe, MPI 3 pe
 - [x] NMM03 bit-for-bit for three domains: MPI 1pe, MPI 3pe
 - [x] ARW12 bit for bit: original vs new
 - [x] NMM03 bit for bit: original vs new

The NETCDF modifications were simpler. This entailed removing unnecessary code. In this case,
removing required code would have broken the compilation process (since both of the impacted
files have `implicit none`). In this case, the model "running" is sufficient indication that the code
modifications were OK. Since one of the locations where the NETCDF changes were introduced 
was inside a WRF_CHEM ifdef, a WRF Chem build and test was conducted.
1. For Chem, a 10-timestep, single domain: CHEM
 - [x] CHEM completes simulation


weiwangncar:
@davegill How general is this? Does it require certain versions of netCDF?

davegill:
@weiwangncar 
Wei,
This PR does not introduce more or less generality regarding versions of netcdf. This PR allows a user to have access to a newer compiler, but neither the netcdf nor the mpi libs needed to be built with that particular version of the compiler. Without too much detail, I installed gnu-7 and netcdf4, both already built. Importantly, the netcdf4 libraries were not built by the gnu-7 compiler.

kkeene44:
@weiwangncar @llpcarson 
Are you guys okay with this? If so, we can go ahead and approve/merge it.

weiwangncar:
@davegill Does your last sentence imply that the netCDF library doesn't have be built with the same Fortran compiler that used by the model?

davegill:
@weiwangncar 
Wei,
The libraries will need to be built with the same compiler that was used to build the *.o files for the model. With this PR the libraries may not always need to be built with the same _version_ of the compiler. This is handy when Linux utilities (yum, get-apt, etc) have canned versions of libraries that they install, and it is likely that the compiler versions used to build those libraries is not the same as the  compiler version required to build WRF.

mgduda:
Does this mean that we will now be supporting the use of libraries installed via a package manager? I.e., do we generally support users who:
1. yum install gcc
2. yum install mpi
3. yum install hdf5
4. yum install netcdf4
5. ./configure and ./compile WRF

davegill:
> Does this mean that we will now be supporting the use of libraries installed via a package manager? I.e., do we generally support users who:
> 
> 1. yum install gcc
> 2. yum install mpi
> 3. yum install hdf5
> 4. yum install netcdf4
> 5. ./configure and ./compile WRF

@mgduda 
Michael
The primary purpose for this commit is to make automated testing (via travis on github) of the WRF model easier. The WRF model requirements for newer portions of the Fortran standard (the IEEE component is an example) make the standard Linux distribution of GCC not a viable alternative. With the Linux  Red Hat `yum` commands, we are able to install a new version of GCC, and also install libraries such as netcdf and openmpi, and use those "installed by a standard utility" codes to run the WRF model. I have not tried any other installation tool.

We generally do not support the installation of external libraries for users. However, on our FAQ page, we have sample source code that can be downloaded, and fairly detailed instructions as to how to use the WPS compression libraries, the netcdf library, and the MPICH variant of the MPI standard. Typically, this information is only useful for users that do not really have any admin support. [For example on a maintained system, it is unlikely that the version of MPICH would be useful.] It is a secondary and nice impact that the user community benefits from a direction that benefits us.

The docker containers can be modified to install the libraries from source.

cmattocks:
To install an up-to-date developer toolchain on RHEL7:

Red Hat Developer Toolset
-------------------------
https://developers.redhat.com/products/developertoolset/hello-world/

sudo yum install centos-release-scl                         # For CentOS 7
sudo yum-config-manager --enable rhel-server-rhscl-7-rpms   # For RHEL 7
sudo yum install devtoolset-8

Modify ~./bashrc:
#-------------------------------
#... Red Hat Developer Tools ...
#-------------------------------
source scl_source enable devtoolset-8

gfortran --version
GNU Fortran (GCC) 8.2.1 20180905 (Red Hat 8.2.1-3)

Note: NOAA/NWS/NCO/WCOSS use RHEL7 for development and production.

mgduda:
@cmattocks I think the issue that @davegill was running into was that the netCDF library installed by yum was built with an older GNU toolchain, and so the `netcdf.mod` file that is part of the Fortran90 interface was incompatible with the newer GNU compilers.

cmattocks:
Okay, got it, thanks for the explanation. Yes, default RHEL7 toolchain is quite old.

davegill:
Going a whole other way with this

BijayGuha:
@davegill how can I get the modified codes you have mentioned??

davegill:
@BijayGuha 
> how can I get the modified codes you have mentioned??

Short answer:
clone my fork (davegill), checkout this branch: zap_f90_netcdf_and_mpi_use
```
git clone https://github.com/davegill/WRF
cd WRF
git checkout zap_f90_netcdf_and_mpi_use
```

Also, take a look at the WRF Forum for a question related to this:
http://forum.mmm.ucar.edu/phpBB3/viewtopic.php?f=37&t=5317

BijayGuha:
Thanks @davegill..I already seen those topics in the forum..

letmaik:
@davegill What does "Going a whole other way with this" mean and what happened to this PR?

PR #804, opened 2019-03-09
Make random seeds for stochastic schemes able to be in restart file and be allocatable

### LINKED ISSUE #799: random seed length for stochastic option
Email from Judith
```
Depending on architecture the length of the seed array
used in the fortran intrinsic subroutines
        call random_seed(get=iseed(1:how_many))
        call random_seed(put=iseed(1:how_many))
can now be longer than "it used to be". The stochastic suite saves this
array as part of the restart files using an array dimensioned with the
number of vertical levels. This results in non-bit reproducible
results for simulations started from restart files, if the stochastic
suite is used. In addition, there is a possibility that the index
refers to an element which is outside the declared array dimensions.
Please let me know when you get a bug-fix version out, so I can merge
the changes on top of the latest version.
```

davegill:
TYPE: bug fix

KEYWORDS: stochastic, restart, b4b, hrrr_cycling

SOURCE: internal

DESCRIPTION OF CHANGES: 
1. Made the random number seed to have a dimension that is intrinsic to the compiler.
2. Force hrrr_cycling = .true. to avoid resetting seeds on restart (the logic was broken).
3. The seeds now have a "Z" stagger (no empty or zero value in last entry).
4. Modify code to accommodate the swap from vertical dimension of seed to the new `seed_dim`.

ISSUE: 
Fixes #799 random seed length for stochastic option

LIST OF MODIFIED FILES:
M Registry/registry.dimspec
M Registry/registry.stoch
M dyn_em/module_first_rk_step_part2.F
M dyn_em/module_stoch.F

TESTS CONDUCTED: 
1. The patterns are identical after a restart.
👍 

RELEASE NOTE: After hrrr_cycling was introduced to WRF, the stochastic solutions were not bit reproducible when compared to the solutions that used a restart. Also, with the newer GNU compilers, the size of the seed array exceeded the vertical dimensions in WRF (the vertical dimension was originally used as an easy and safe value). Both problems have been addressed.

### DIFF
--- a/Registry/registry.dimspec
+++ b/Registry/registry.dimspec
@@ -30,6 +30,7 @@ dimspec    nndgvar -     namelist=nobs_ndg_vars            c     nobs_ndg_vars
 dimspec    [       -     namelist=obs_prt_max              c     obs_prt_max
 dimspec    obsstid -     constant=40                       c     obs station id names
 dimspec    &       2     namelist=lagday                   z     lagday
+dimspec    seed    1     namelist=seed_dim                 z     seed_dim
 endif
 
 
--- a/Registry/registry.stoch
+++ b/Registry/registry.stoch
@@ -55,14 +55,14 @@ state    real  SPPT_AMP       ij   misc   1 -   r          "SPPT_AMP"       "amp
 
 
 # 1d arrays
-state    real  VERTAMPT        k   misc   1 -   r          "VERTAMPT"       "vert. amplitude of stoch. temperature perturb." "" ""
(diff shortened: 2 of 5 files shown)


kkeene44:
@dudhia 
Can you take a look at this one? If you're okay with it, we can approve and merge it.

davegill:
@weiwangncar @dudhia 
Wei and Jimy,
Judith is OK with the status of the restart with the stochastic option activated. Would you please provide a review.

PR #808, opened 2019-03-12
Jensen look-up table files: Makefile (add links), clean (add removes)

kkeene44:
TYPE: bug fix

KEYWORDS: jensen, bin, ishmael, mp_physics

SOURCE: internal

DESCRIPTION OF CHANGES: 
1. The top-level Makefile did not have the ishmael\*bin files linking from run/ to test/em_real/. That was added so that Jensen mp will run correctly without having to manually copy those files over. 
2. The ishmael\*bin files were also added to the clean script so that they will be removed from test/\*/ upon a clean.

LIST OF MODIFIED FILES: 
M    Makefile
M    clean

TESTS CONDUCTED: Verified that it configures/builds okay and properly links the files to the test/em_real directory during the build.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -627,6 +627,9 @@ em_real : wrf
              ln -sf ../../run/CCN_ACTIVATE.BIN . ;                   \
              ln -sf ../../run/p3_lookup_table_1.dat-v2.8.2 . ;                   \
              ln -sf ../../run/p3_lookup_table_2.dat-v2.8.2 . ;                   \
+             ln -sf ../../run/ishmael-gamma-tab.bin . ;             \
+             ln -sf ../../run/ishmael-qi-qc.bin . ;             \
+             ln -sf ../../run/ishmael-qi-qr.bin . ;             \
              ln -sf ../../run/BROADBAND_CLOUD_GODDARD.bin . ;        \
              if [ $(RWORDSIZE) -eq 8 ] ; then                       \
                 ln -sf ../../run/ETAMPNEW_DATA_DBL ETAMPNEW_DATA ;  \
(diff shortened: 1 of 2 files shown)


davegill:
@kkeene44 
Kelly,
Would you please also update the clean script with these files? If the Jensen files are linked into various directories, we need to be able to get rid of them.

kkeene44:
@davegill 
Good point. I included the modified clean script in the commit. 
As for the Makefile, the ishmael* files were included in the section for linking files to em_esmf_exp, which I don't think is necessary? I don't think this hurts anything, but if it's not needed there, for cleanliness, should those lines be removed from that part of the Makefile?

davegill:
@kkeene44 
Kelly,
Even though we don't use that esmf dir, there is no reason to break with tradition RIGHT NOW and remove those. Good catch to put the files actually where we need them.

davegill:
@kkeene44 
Kelly,
Modify the title, commit message, and files changed to reflect the inclusion of the clean script in this PR.

kkeene44:
@davegill 
Oh yeah, thanks. I updated it.

davegill:
Let's get Wei or Jimy to give a thumbs up. I am approving it, but wait to merge until they leave an OK comment.

kkeene44:
@weiwangncar @dudhia 
Dave approved this, but wanted one of you to look at it before I merge it. Any objections?

weiwangncar:
I don't have problem with it, and it's up to Anders. @graupely If you can think of any other places that this kind of things can happen in the code (such as preventing negative cloud variables from happening), it would be good to add some protection.

weiwangncar:
I guess I put the comment for the wrong PR. I'm ok with this change.

PR #809, opened 2019-03-13
WSM6: remove unnecessary ifdef for purpose of WRF/MPAS unification

kkeene44:
TYPE: no impact

KEYWORDS: wsm6, wrf, mpas, unify, if loop

SOURCE: internal

DESCRIPTION OF CHANGES: 
At the top of the wsm6 module was an if defined (wrfmodel) loop that included a few 'use' 
statements for models/routines that are currently not used in this MP scheme. Removed that 
loop to unify this routine with MPAS and to clean-up unnecessary dependencies.

LIST OF MODIFIED FILES: 
M   phys/module_mp_wsm6.F

TESTS CONDUCTED: 
 - [x] verified that this builds without problems and produces bit-for-bit results before and after mods.

### DIFF
--- a/phys/module_mp_wsm6.F
+++ b/phys/module_mp_wsm6.F
@@ -8,10 +8,6 @@
 
 MODULE module_mp_wsm6
 !
-#if ( defined(wrfmodel) )
-   USE module_utility, ONLY: WRFU_Clock, WRFU_Alarm
-   USE module_domain, ONLY : HISTORY_ALARM, Is_alarm_tstep
-#endif
    USE module_mp_radar
 !
    REAL, PARAMETER, PRIVATE :: dtcldcr     = 120. ! maximum time step for minor loops


davegill:
@kkeene44 
Kelly,
Weird.
These USE associations are used to determine if a particular alarm is ringing, in this case the history alarm (is it time to write to wrfout_d0x). This could have been a "one-off" a long time ago, or perhaps it was used to determine when to compute the radar computations before those were centralized. That computation was expensive and only done when it was a known history output time.

davegill:
@dudhia @weiwangncar 
Jimy and Wei,
This is in the physics dir, so you need to take a look. 
This looks just fine to me. I approve.

weiwangncar:
@kkeene44 @davegill Good point. @kkeene44 Can you do a test with 'do_radar_ref=1 on?  It seems that it will work. But it doesn't hurt to check. Once the model is run, take a look at the output to look for refl_10cm field in 3D (or 4D) list.

kkeene44:
@weiwangncar @davegill 
I verified that refl_10cm looks okay when do_radar_ref is turned on.

PR #812, opened 2019-03-14
Jensen Ishmael microphysics: updates for initial release

graupely:
TYPE: enhancement

KEYWORDS: code changes/enhancements 

DESCRIPTION OF CHANGES: 
1. Added double precision-ness for lookup tables
2. Added ice fall speed re-calculation before sedimentation but after tendency updates
3. Aggregation tendencies (mass and number) no longer occur in the aggregation subroutine and now occur along with the other microphysical tendencies
4. Ice particle shape prediction limited to threshold ice particle size and number concentration (helps limit the occurrence of extreme ice particle shapes from advection/diffusion)
5. Limited riming to threshold QCLOUD amount consistent with lookup table
6. Updated formulation for ice number mixing ratio tendency during melting
7. Consistently set the small ice size limit to 2 microns
8. Moved aggregation call to be in the same location as other process rate calculations
9. Combined all microphysical tendencies for calculation to ensure over-depletion of any hydrometeor does not occur
10. Removed shedding to rain during riming when temperature is above melting
11. Combined all microphysical tendency terms
12. Check added to ice distribution shape change during nucleation
13. Allow aggregates to densify/become spherical during melting
14. Added number and size threshold to output diagnostic calculations
15. Removed diagnostic to determine ice types, e.g. cloud ice, snow, graupel (to be added later when I find some more time)
16. Added limit to precipitation based on mass mixing ratio threshold
17. Added variable check after sedimentation (to keep prognostics in bounds)
18. Changed large ice size threshold to a mass-weight maximum diameter of 14 mm
19. Set a maximum vapor-growth density threshold of 700 kg m^-3
20. Changed the minimum aggregation efficiency allowed
21. Set a low threshold for the number mixing ratio of aggregates used to calculate size

LIST OF MODIFIED FILES: 

TESTS CONDUCTED: 
Builds and runs correctly including restarts

### DIFF
--- a/phys/module_mp_jensen_ishmael.F
+++ b/phys/module_mp_jensen_ishmael.F
@@ -2831,7 +2831,7 @@ contains
   !.. Set incoming limit after sedimentation
   !.. smallest values for which we care about shape              
              if(ai(cc,k).lt.1.e-12.or.ci(cc,k).lt.1.e-12) then
-                ai(cc,k)=max(ai(cc,k),ci(cc,k))
+                ai(cc,k)=min(ai(cc,k),ci(cc,k))
                 ci(cc,k)=ai(cc,k)
                 ani(cc)=(ai(cc,k)/ni(cc,k))**0.3333333333
                 cni(cc)=(ci(cc,k)/ni(cc,k))**0.3333333333
@@ -3828,7 +3828,7 @@ contains
 
     implicit none
     
-    real, parameter :: QASMALL= 1.e-19   ! Smallest ice size   (m^-3 kg^-1)
+    real, parameter :: QNSMALL= 1.25e-7   ! Smallest ice number
     integer icat,k,kp1,kp2,isnow,ipris,it,ihab,jhab
     integer, parameter :: ncat=7,npair=35,ndn=60,nftab=1000,ngam=5000,nz=1
     real, parameter :: rsmall=1.e-12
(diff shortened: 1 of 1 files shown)


weiwangncar:
@graupely Can you provide some guidance on what the recommended resolutions are to use this scheme? If it is general enough to use for any resolution, that would be good to know too.

graupely:
@weiwangncar: For a short answer, I have run the scheme at resolutions spanning 148-m (1-second timestep) to 12-km (54 second timestep). 

Now for the long answer: 
The parameterization predicts ice particle shape evolution and ice shape can change rapidly when ice grows by vapor deposition or riming, and thus, short microphysical timesteps are preferred. Originally, I build the microphysics parameterization as a research model, and thus, my WRF test cases that have been published were run using 1-3 km grid spacing and 2-6 second timesteps. I personally prefer to run the microphysics with dx <= 3 km and dt <= 6 s (this is the preferred setup), but I have tried to designed the parameterization to be flexible. I've had success running a wintertime WRF test cases using the HRRR CONUS domain with 9-km grid spacing and an 18-s timestep. I have also successfully run nested domains without issue. One thing that I have noticed when using longer timesteps/larger grid spacing is that for small ice mass mixing ratios near cloud boundaries, extreme (very non spherical) ice particle shapes can occur as grid-point values. Using diff_6th_opt = 2 seems to help with this issue. I also prefer to use moist_adv_opt=2 and scalar_adv_opt = 2 with the scheme. The scheme is not officially coupled to cumulus parameterizations, though I do use the official Qice registry variable for planar ice. The other two ice species (qice2 or columnar and qice3 or aggregates) won't interact with cumulus parameterizations.

weiwangncar:
@graupely Thanks for the answers. Regarding the long time step issue, have you tried the WENO scheme? I know it can help with other microphysical variables at the cloud edges.

graupely:
@weiwangncar, I haven't tested WENO recently, but I will give it a try. Thanks!

PR #815, opened 2019-03-14
Update to MYNN-EDMF

joeolson42:
Update MYNN-EDMF

TYPE: bug fix, small enhancement

KEYWORDS: subgrid cloud fractions

SOURCE: Joseph Olson/Jaymes Kenyon (NOAA/CIRES)

DESCRIPTION OF CHANGES: Bug fix for mass-flux cloud fractions. This bug has little/no impact on SW-down since LWP was not impacted (mixing ratio is assumed to be diluted when cloud fraction exceed the area of the updrafts). This bug was inserted in the previous MYNN commit, when the plume properties (qv & qv_sat) were accidentally used to compute the subgrid cloud fraction instead of the environmental properties, (2) small adjustment to non-convective subgrid clouds to improve ceiling statistics. This update will keep the MYNN-EDMF very close to the version used for the next operational upgrade (frozen code date only 2-3 weeks away).

LIST OF MODIFIED FILES: 
phys/module_bl_mynn.F

TESTS CONDUCTED: Retrospective tests performed in RAP show no change in skill for downward SW radiation, but the subgrid cloud fractions are more reasonable.

### DIFF
--- a/inc/version_decl
+++ b/inc/version_decl
@@ -1,1 +1,1 @@
-   CHARACTER (LEN=10) :: release_version = 'V4.1-alpha'
+   CHARACTER (LEN=10) :: release_version = 'V4.0.3    '
--- a/inc/version_decl
+++ b/inc/version_decl
@@ -1,1 +1,1 @@
-   CHARACTER (LEN=10) :: release_version = 'V4.0.3    '
+   CHARACTER (LEN=10) :: release_version = 'V4.1-alpha'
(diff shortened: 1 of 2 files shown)


davegill:
@joeolson42 
Joe,
1. The "develop" branch is OK. However, are you trying to get these mods into the release branch for v4.1? If so, the the base branch should be "release-v4.1".
2. Update the PR title. Right now all we have is "Develop".
3. Update the PR commit message. Currently, it is just the boilerplate template we provide.
4. Would you provide some information about testing.

PR #817, opened 2019-03-14
Update parameters in icloud3, changes to AOD for Thompson-Eidhammer aerosol-aware scheme

gthompsnWRF:
TYPE: enhancement

KEYWORDS: cloud fraction scheme (icloud3)

SOURCE:  Greg Thompson (NCAR)

DESCRIPTION OF CHANGES: 
1. icloud=3 
   a. Updates to internal parameters
   b. Changing from 3D structure to 1D column only
2. Minor RH increase in the AOD calculation.

LIST OF MODIFIED FILES:
phys/module_radiation_driver.F

TESTS CONDUCTED: Ran in real-time forecast mode for a month. Compared to older version, expect far lower cloud fraction amount at high altitude (cold tropopause level temperatures for example). Full year (every 5th day) test by Mei Xu gives results with better shortwave radiation at ground comparison to observations.

RELEASE NOTE:
Cloud fraction scheme (icloud=3 namelist option) compared to prior version generally produces lower cloud fraction amount at high altitude (cold tropopause level temperatures for example). Full year (every 5th day) test by Mei Xu gives results with better shortwave radiation at ground comparison to observations.

### DIFF
--- a/phys/module_radiation_driver.F
+++ b/phys/module_radiation_driver.F
@@ -826,6 +826,7 @@ CONTAINS
    REAL, DIMENSION( ims:ime, kms:kme, jms:jme ) :: qs_cu_weight
 
    REAL    ::    gridkm, Wice,Wh2o
+   REAL, DIMENSION(kms:kme):: t_1d, p_1d, Dz_1d, qv_1d, qc_1d, qi_1d, qs_1d, cf_1d
 
    REAL    ::    next_rad_time, DTaccum
    LOGICAL ::    run_param , doing_adapt_dt , decided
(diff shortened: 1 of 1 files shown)


davegill:
@gthompsnWRF 
Greg,
Would you add just another sentence or two to the commit message: what parameters (for a particular term, process, to address something in particular, etc). Also the RH increase, was this direct or a cause of a change.

gthompsnWRF:
(1) The prior release of this option had too low RH threshold for cold temperature ice clouds, causing cloud fraction to be often very near 100%; although the assumed ice water contents were very low keeping those high cloud fractions from being too opaque to radiation.  This newly changed code greatly reduces cloud fraction amount at upper troposphere temperatures. Another change is making the critical RH value include the vertical grid spacing in addition to the horizontal spacing.

(2) The AOD calculations for swelling the size of aerosols due to high RH was cut off at 98%, but this is now permitted to come up to 99% as the upper limit.

PR #820, opened 2019-03-15
Jensen microphysics: negative qi in vapor growth

kkeene44:
TYPE: bug fix

KEYWORDS: jensen, mp_physics, qi, floating_point

SOURCE: internal

DESCRIPTION OF CHANGES: In the vapor growth check in Jensen microphysics, qi had the ability to be negative, which means that in the equation for ani(cc), there was potential to raise a negative number to an exponent of a non-logical fractional value, which caused a floating point error. This only stopped the model from running when compiled with a check for floating point traps. Inserting an if statement to prevent this from happening corrects the problem.

LIST OF MODIFIED FILES: 
M    phys/module_mp_jensen_ishmael.F

TESTS CONDUCTED: Verified that code builds and runs to completion, even when compiled with floating point trap checks.

### DIFF
--- a/phys/module_mp_jensen_ishmael.F
+++ b/phys/module_mp_jensen_ishmael.F
@@ -1647,7 +1647,7 @@ contains
   !.. If vapor growth of ice is going to cause
   !.. subsaturated condtions, reduce vapor growth
   !.. rates (done above) and rediagnose the axes
-             if(vgflag) then
+             if(vgflag.and.qi(cc,k).gt.QSMALL) then
                 alphstr=ao**(1.-deltastr(cc))
                 alphv=fourthirdspi*alphstr
                 betam=2.+deltastr(cc)


kkeene44:
@graupely
Will you take a look at the modified code? I used a low value of 1.e-20 for qi, but I'd like your input on the best value to use, and if you'd rather me declare a variable for this, similar to the 'QSMALL' variable you created.

graupely:
> @graupely
> Will you take a look at the modified code? I used a low value of 1.e-20 for qi, but I'd like your input on the best value to use. If you'd rather me declare a variable for this, similar to the 'QSMALL' variable you created, or do this any differently, just let me know.

@kkeene44, can you move the check to line 1650, and yes you can just use QSMALL:
 if(vgflag.and.qi(cc,k).gt.QSMALL) then

Thanks!

kkeene44:
@graupely 
Thanks for getting back to me. By line 1650, do you prefer this before or after the 'if (vgflag)' test?

graupely:
@kkeene44 
Can you do it on the same line as the vgflag test, i.e. if(vgflag.and.qi(cc,k).gt.QSMALL)?

kkeene44:
@graupely 
I apologize. You definitely did say that and I overlooked that. If you refresh the 'Files changed' page, you should see the update. Thanks!

weiwangncar:
I don't have problem with it, and it's up to Anders. @graupely If you can think of any other places that this kind of things can happen in the code (such as preventing negative cloud variables from happening), it would be good to add some protection. (moved from PR#808)

graupely:
@kkeene44, thanks for the fix. 
@weiwangncar, I have included small ice mass checks throughout the microphysics, but I'll look over the code again to double check that I didn't miss any places where negative ice mass could occur.

PR #822, opened 2019-03-16
KSAS: scale dependent trigger, conversion parameter, xmb trigger condition

weiwangncar:
TYPE: bug fix

KEYWORDS: KSAS

SOURCE: Ji-Young Han and Songyou Hong (KIAPS)

DESCRIPTION OF CHANGES: 
1. Revised scale-dependent triggering function (cinpcr). This makes suppression of the convection triggering at resolutions less than 10 km more significant as compared to the original KSAS.

2. Fix the conversion parameter that determines the fraction of condensate converted to precipitation (c0) for convective overshooting layer. This leads to an increase in detrained moisture near the cloud top, resulting a decreased LW cooling effect in the tropical upper troposphere.

3. Fix cloud base mass flux xmb trigger condition, and this has a minor effect.

LIST OF MODIFIED FILES: 
M  phys/module_cu_ksas.F

TESTS CONDUCTED: 
It compiles. More later.

RELEASE NOTE:
For KSAS, the scale-dependent trigger function is revised to suppress more convection at model resolution less than 10 km. Fixed the conversion parameter that determines the fraction of condensate converted to precipitation (c0) for convective overshooting layer. This leads to an increase in detrained moisture near the cloud top, resulting a decreased LW cooling effect in the tropical upper troposphere.

### DIFF
--- a/phys/module_cu_ksas.F
+++ b/phys/module_cu_ksas.F
@@ -353,6 +353,11 @@ CONTAINS
 !   16-11-01  ji-young han     revised pgcon & bug fix in vshear
 !   17-02-23  ji-young han     revised xlamb
 !   18-03-01  ji-young han     kim sas
+!   19-03-01  ji-yeon jang     revised triggering ftn for higher resolution<10km
+!             yong-hee lee
+!             young cheol kwon
+!             eun-jeong lee    bug fix in c0 for overshooting layer
+!             seung-bu park    bug fix in xmb trigger
 !
 ! usage:    call phys_cps_sas(delt,delx,del,prsl,prsi,prslk,prsik,zl,          &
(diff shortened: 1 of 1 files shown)


davegill:
@weiwangncar 
Wei,
Would you also add in that the references were updated

weiwangncar:
@davegill This PR doesn't have any references.

davegill:
@weiwangncar 
Wei,
I saw this
```
!   19-03-01  ji-yeon jang     revised triggering ftn for higher resolution<10km
!             yong-hee lee
!             young cheol kwon
!             eun-jeong lee    bug fix in c0 for overshooting layer
!             seung-bu park    bug fix in xmb trigger
```
... and just assumed references. I withdraw my request.

davegill:
@weiwangncar 
Wei,
Would you define `xmb` in the PR commit message.

PR #826, opened 2019-03-18
Cray: gcc -> cc

davegill:
TYPE: enhancement

KEYWORDS: Cray, gcc, cc

SOURCE: Peter Johnsen and Patricia Balle (Cray, Inc.)

DESCRIPTION OF CHANGES: 
Change the default C compiler from gcc to cc, for the stanza
```
Cray XE and XC CLE/Linux x86_64, Cray CCE compiler
```

LIST OF MODIFIED FILES: 
M arch/configure.defaults

TESTS CONDUCTED: 
None required.


davegill:
@kkeene44 
Kelly,
This is one of those 1 character changes. The Cray support engineers would like to replace `gcc` with `cc` in the arch/configure.defaults file. Would you do a careful review.

PR #829, opened 2019-03-19
Syntax errors: (a ** -b) becomes (a ** (-b)); in new Jensen and Goddard MP schemes

smileMchen:
Type: bug fix

KEYWORDS: cray, exponent, syntax

SOURCE: Found by Peter Johnsen and Patricia Balle (Cray, Inc.), fixed internally

DESCRIPTION OF CHANGES: 
A few  illegal syntax examples of the form below were found by the Cray compiler in release-v4.1. 
These are the only instances in the phys directory (excluding comments).
```
a ** -b
```
The correct form is
```
a ** (-b)
```
None of the NCAR cheyenne compilers (Intel, PGI, GNU) complain about this particular syntax, 
since it is treated as an extension of the standard.
However, the code cannot compile with the CRAY compiler. 

If the Fortran 2003 standard is enforced with Intel, this construct flags a fatal error. 
```
> cat foo.f90
program foo1
print *, 'Expecting 5/9 = ',5./9., 3. ** -2 * 5.
end program foo1
```
With Intel as the compiler:
```
> ml
Currently Loaded Modules:
  1) ncarenv/1.2   2) intel/17.0.1   3) ncarcompilers/0.4.1   4) mpt/2.19   5) netcdf/4.6.1

> ifort -e03 foo.f90
foo.f90(2): error #8810: Consecutive operators are an extension to the Fortran 2003 standard.
print *, 'Expecting 5/9 = ',5./9., 3. ** -2 * 5.
-----------------------------------------^
compilation aborted for foo.f90 (code 1)
```
Even with newer GNU compilers, we can get a warning by default:
```
> ml gnu/8.1.0
Lmod is automatically replacing "intel/17.0.1" with "gnu/8.1.0".

Due to MODULEPATH changes, the following have been reloaded:
  1) mpt/2.19     2) ncarcompilers/0.4.1     3) netcdf/4.6.1

> gfortran foo.f90
foo.f90:2:42:

 print *, 'Expecting 5/9 = ',5./9., 3. ** -2 * 5.
                                          1
Warning: Extension: Unary operator following arithmetic operator (use parentheses) at (1)
```

LIST OF MODIFIED FILES:
M  phys/module_mp_gsfcgce_4ice_nuwrf.F
M  phys/module_mp_jensen_ishmael.F

TESTS CONDUCTED: 
 - [x] Model results are bit-for-bit (before vs after)

### DIFF
--- a/phys/module_mp_gsfcgce_4ice_nuwrf.F
+++ b/phys/module_mp_gsfcgce_4ice_nuwrf.F
@@ -2044,7 +2044,7 @@ CONTAINS
 
        ftnw=1.                                                       
            if(ql.lt.cmin .and. tair .gt. t0)then                      
-             bin_factor=0.11*(1000.*qri)**-1.27 + 0.98      
+             bin_factor=0.11*(1000.*qri)**(-1.27) + 0.98      
              bin_factor=min(bin_factor, 1.30)       
              ftnw=1./bin_factor**3.35                             
              ftnwmin=r00*qri/draimax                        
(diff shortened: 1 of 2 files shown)


davegill:
@smileMchen 
Ming,
The only test that would be required would be to make sure that the code compiles and the results are bit for bit after a couple of time steps for these two schemes.

davegill:
@graupely
Anders,
Would you test this PR to you code? We want, for a few time steps, to demonstrate bit-for-bit identical results between the original code and the code with this mod. If it is OK, edit the commit message, and send a note out to Kelly, Wei, Jimy, and me.

graupely:
@davegill @kkeene44 @smileMchen @dudhia, the change is bit-for-bit in Jensen micro.

PR #832, opened 2019-03-22
Jensen Ishmael microphysics: additional update for initial release

graupely:
TYPE: enhancement

KEYWORDS: code changes/enhancements

DESCRIPTION OF CHANGES:
In the check to ensure that microphysical processes do not over-deleted ice (lead to negative ice mass mixing ratios), reduce the ice mass tendency during melting (qmlt) if necessary, but do not reduce the ice number and volume mixing ratio tendencies (nmlt, amlt, cmlt), consistent with sublimation (prd). Checks to ensure that mass number and volumes stay positive after melting exist just below. 

Changes added 27 March 2019:
1) Initialize value of dsold=1 (tracks the shape) before microphysics occurs. 
2) Lower the minimum vapor growth limit and add a timestep dependence (line 1142).
3) Check on melting ice to ensure that it becomes spherical (phi=1) but that shape does not go past spherical, i.e. (phi<1 before melting and phi>1 after melting).
4) Limit aggregate shape to oblate and spherical (never prolate), lines 2439 and 2875.
Bug fixes:
5) Set a low limit on the number of aggregates, fx(idx), in aggregation lookup table
6) Set a low limit on the fallspeed of the collector, vx, in the aggregation lookup table
7) Set a low/high limit on dxy in the aggregation lookup table for gfortran happiness. 

LIST OF MODIFIED FILES:

TESTS CONDUCTED:
Builds and runs correctly

### DIFF
--- a/phys/module_mp_jensen_ishmael.F
+++ b/phys/module_mp_jensen_ishmael.F
@@ -1697,9 +1697,6 @@ contains
              if(sink.gt.source.and.qi(ICE1,k).gt.QSMALL) then
                 ratio= source/sink
                 qmlt(ICE1) = qmlt(ICE1)*ratio
-                nmlt(ICE1) = nmlt(ICE1)*ratio
-                amlt(ICE1) = amlt(ICE1)*ratio
-                cmlt(ICE1) = cmlt(ICE1)*ratio
                 qagg(ICE1) = qagg(ICE1)*ratio
                 nagg(ICE1) = nagg(ICE1)*ratio
                 if(prd(ICE1).lt.0.) then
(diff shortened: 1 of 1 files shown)


graupely:
@davegill, This will (likely) be my final code change before release as long as I don't find any bugs. I'll have one more "comments only" pull request to improve the documentation. When will you need that by?

davegill:
@weiwangncar @dudhia 
Folks,
Would you please review these suggested mods by the developer?

davegill:
@graupely 
Anders,
Get your changes in this week, and there are no troubles.

weiwangncar:
I'm ok with the change.

PR #839, opened 2019-03-28
Require monotonic vertical pressure decrease: test after hybrid vert coordinate calculation in real 

davegill:
TYPE: enhancement

KEYWORDS: etac, monotonic pressure, hybrid coordinate, hvc

SOURCE: Found by Michael Toy (NOAA), fix internal

DESCRIPTION OF CHANGES: 
Problem:
Over very high topography, the flattened eta surfaces may intersect the eta surfaces below. The 
WRF model dies quickly, and it is very difficult to figure out what happened. In simple terms, the
reference pressure must decrease as you go up the k-index, for each (i,j). If not, then that is a 
fatal error.

Solution:
Trap this event via looking at the columns of reference pressure. Any column that is NOT monotonic 
flags a fatal error. The user is given the location, the pressure values, and a suggestion to reduce the 
etac (the critical eta surface where the levels becomes isobaric).

A similar warning was issued in 1984, though it dealt with crossing _streams_ and _total protonic 
reversal_ instead of crossing _eta surfaces_.
https://www.youtube.com/watch?v=jyaLZHiJJnE

LIST OF MODIFIED FILES: 

TESTS CONDUCTED: 
1. For the user's case that originally brought up this problem, the real program correctly stopped.
```
----- ERROR: The reference pressure is not monotonically decreasing
             This tends to be caused by very high topography
 (i,j) =          979         414, topography =    7346.500     m
 k =           13, reference pressure =    35442.12     Pa
 k =           14, reference pressure =    35500.98     Pa
FATAL CALLED FROM FILE:  <stdin>  LINE:    1376
 In the dynamics namelist record, reduce etac from   0.2000000
```
2. For Jan 2000 case, artificially introduced a point with 9000 m elevation: caught by real.
```
----- ERROR: The reference pressure is not monotonically decreasing
             This tends to be caused by very high topography
 (i,j) =           10          10 , topography =    9000.00000      m
 k =            8 , reference pressure =    29485.4062      Pa
 k =            9 , reference pressure =    29518.7754      Pa
FATAL CALLED FROM FILE:  <stdin>  LINE:    1690
 In the dynamics namelist record, reduce etac from   0.200000003
```

### DIFF
--- a/dyn_em/module_initialize_real.F
+++ b/dyn_em/module_initialize_real.F
@@ -174,6 +174,10 @@ integer::oops1,oops2
 
       INTEGER :: geogrid_flag_error
 
+      !  Vertical pressure checks
+
+      REAL :: press_above, press_below
+
       !  Dimension information stored in grid data structure.
 
       CALL cpu_time(t_start)
@@ -627,6 +631,8 @@ integer::oops1,oops2
          END IF
 
          !  Variables that are named differently between SI and WPS.
+
+grid%ht_gc(10,10)=9000
 
          DO j = jts, MIN(jte,jde-1)
             DO i = its, MIN(ite,ide-1)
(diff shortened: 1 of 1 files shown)


davegill:
@dudhia @weiwangncar 
Jimy and Wei,

This fixes the concern that Joe had a while back. He was concerned that if we had a high-enough topography, then we could end up having the flatter eta surface at index k+1 intersect the eta surface at index k. The only solution looks to be to set etac lower (such as from 0.2 to 0.02) or to set ptop much higher. 

However, the important thing is to identify this problem in the real program before WRF dies.

The coefficients computed by Joe's polynomial are used to compute a reference pressure. Each column is checked for a strictly decreasing reference pressure profile.

davegill:
@dudhia @weiwangncar 
Jimy and Wei,
We can augment this later. For example:
1. We can check each column, looking at dz. If that value is < the first dz in the column, FAIL.
2. Look at the dp_terrain in the column compared to the associated dp_nominal at the level over a nominal ocean point. Any dp_terrain/dp_nominal < the column_depth_terrain / column_depth_nominal would be flagged to fail.
We would need reasonable default values, and we would need a way to over rule the default behavior. Let's do this second part, which deals with a more subtle problem of thin layers, afterwards.

weiwangncar:
I'm ok with the change.

dudhia:
OK, remember we also want to check for very thin layers at some point

PR #844, opened 2019-03-29
Allow Deng shcu to work with deep cu, remove prints, edit README

weiwangncar:
TYPE: no impact, and text only

KEYWORDS: Deng shallow cu, README

SOURCE: Internal

DESCRIPTION OF CHANGES: 
After some tests, we decided that we will remove the hard constraint against allowing the Deng 
scheme to run with another deep scheme (i.e., we WILL allow the Deng scheme to run with another 
deep convection scheme). This is in part because, even though there is a component in the scheme 
for deep convection, it isn't very active. Without another deep scheme, the Deng scheme could 
potentially generate grid-point storms by not removing enough instability. 

Other changes include removing prints from Deng's module, and adding some text for the 
README.namelist file.

LIST OF MODIFIED FILES: 
M       phys/module_shcu_deng.F - Only removing prints and debugging comments
M       run/README.namelist - Better description of Deng as shallow + deep
M       share/module_check_a_mundo.F - Remove requirement of cu_physics == 0

TESTS CONDUCTED: 
The compiles!

### DIFF
--- a/run/README.namelist
+++ b/run/README.namelist
@@ -797,7 +797,9 @@ Namelist variables for controlling the adaptive time step option:
                                      = 1, Grell 3D ensemble scheme (use with cu_physics=93 or 5) (PLACEHOLDER: SWITCH NOT YET IMPLEMENTED--use ishallow)
                                      = 2, Park and Bretherton shallow cumulus from CAM5 (CESM 1_0_1)
                                      = 3, GRIMS shallow cumulus from YSU group
-                                     = 5, Deng cumulus scheme (including both shallow and deep convections) from PSU and WRF-Solar
+                                     = 5, Deng cumulus scheme (including both shallow and deep convections) from PSU and WRF-Solar.
+                                          However the deep part isn't very active. Not recommended to use along for deep convection 
+                                          case. Could work well for grid sizes 3-9 km.
                                           This scheme only works with MYJ or MYNN PBL schemes 
 
  ishallow                            = 0, !  = 1 turns on shallow convection, used with Grell 3D ensemble schemes (cu_physics = 3 or 5)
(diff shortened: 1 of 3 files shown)


run/README.namelist
```diff
@@ -797,7 +797,9 @@ Namelist variables for controlling the adaptive time step option:
                                      = 1, Grell 3D ensemble scheme (use with cu_physics=93 or 5) (PLACEHOLDER: SWITCH NOT YET IMPLEMENTED--use ishallow)
                                      = 2, Park and Bretherton shallow cumulus from CAM5 (CESM 1_0_1)
                                      = 3, GRIMS shallow cumulus from YSU group
-                                     = 5, Deng cumulus scheme (including both shallow and deep convections) from PSU and WRF-Solar
+                                     = 5, Deng cumulus scheme (including both shallow and deep convections) from PSU and WRF-Solar.
+                                          However the deep part isn't very active. Not recommended to use along for deep convection
```
davegill: "to use alone for deep convection"
along -> alone ??

davegill:
@weiwangncar 
Wei,
To be clear, now with Deng scheme a user can:
1. Run with Deng scheme without any other cumulus scheme
2. Run Deng scheme with any cumulus scheme

Can a user run Deng with another shallow scheme? If not, then maybe change the description in the README.namelist, changing it from a _cumulus_ scheme to a _shallow cumulus_ scheme.

run/README.namelist
davegill: maybe "Deng cumulus scheme" to "Deng shallow cumulus scheme"??

weiwangncar:
@davegill Now we allow a user to use Deng scheme alone, or Deng + another deep scheme. No a user definitely should not use Deng scheme with another shallow scheme.

weiwangncar:
@dudhia Are you ok with this change?

dudhia:
Yes, I support this change,
Jimy

PR #850, opened 2019-04-01
Fix DA default gen_be_vertloc function

jamiebresch:
TYPE: enhancement

KEYWORDS: WRFDA_EnVar, gen_be_vertloc, alpha_vertloc

SOURCE: Jamie Bresch (NCAR)

DESCRIPTION OF CHANGES:
To change the default function in a stand-alone utility for generating
be.vertloc.dat, an input file required when alpha_vertloc=.true. and ensdim_alpha > 0.
Old default: kscale = 10.0 * real(k) * nk_inv
New default: kscale = 5.0 + real(nk)/real(k+3)
An example of log-P method (dependent on vertical level configuration) is also provided in the code.

The old default suppresses increments at low levels above surface where vertical correlation
exists in ensemble perturbations.

LIST OF MODIFIED FILES:
M       var/gen_be/gen_be_vertloc.f90

TESTS CONDUCTED:
Month-long hourly cycling over CONUS domain to discover the problem in the original default gen_be_vertloc
and to test the impact of various alpha vertical localization settings.

RELEASE NOTE: For WRFDA EnVar applications, gen_be_vertloc is fixed to allow proper increments from ensemble contributions at low levels above surface. Users are encouraged to modify the empirical 5.0 + real(nk)/real(k+3) function in the code for their own model levels and ensemble configurations.

### DIFF
--- a/var/gen_be/gen_be_vertloc.f90
+++ b/var/gen_be/gen_be_vertloc.f90
@@ -26,12 +26,65 @@ program gen_be_vertloc
    real*8,allocatable      :: evec(:,:)
    real*8,allocatable      :: v(:)
    real*8,allocatable      :: x(:)
-   
+   ! variables for log-P method
+   real*8,allocatable      :: znw(:)
+   real*8,allocatable      :: p_w(:) ! pressure at full levels
+   real*8,allocatable      :: ln_p_w(:)
+   real*8,allocatable      :: dlnp(:)
+   real*8                  :: p00, ptop, s_ens_v, grd
+   ! if set use_logp =.true., need to edit ,s_ens_v, ptop and znw below in the code.
+   logical                 :: use_logp = .false.
+
   cnk=""
   call getarg( 1, cnk )
   read(cnk,'(i3)')nk2
   nk=nk2-1
   write(stdout,'(a,i6)') ' vertical level = ', nk2
+
+  if ( use_logp ) then
+     allocate(znw(1:nk+1))
+     allocate(p_w(1:nk+1))
+     allocate(ln_p_w(1:nk+1))
+     allocate(dlnp(1:nk))
+     znw(:) = -99.0
+     p00 = 100000.0
+
+     ! empirical settings
+     s_ens_v = 0.388
+
+     ! HRRR settings
+     !ptop = 1500.0
+     !znw(1:nk+1) = (/ 1.0, 0.998, 0.994, 0.987, 0.975, 0.959, 0.939, 0.916, 0.892, 0.865, 0.835, &
+     !   0.802, 0.766, 0.727, 0.685, 0.64, 0.592, 0.542, 0.497, 0.4565, 0.4205,  &
+     !   0.3877, 0.3582, 0.3317, 0.3078, 0.2863, 0.267, 0.2496, 0.2329, 0.2188,  &
+     !   0.2047, 0.1906, 0.1765, 0.1624, 0.1483, 0.1342, 0.1201, 0.106, 0.0919,  &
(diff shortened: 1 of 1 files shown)


jamiebresch:
One example of u/v/t/rh vertical correlation of an 80-member ensemble perturbations.
ep_vert_corr_d01

jamiebresch:
Correlation of several vertical localization as specified in gen_be_vertloc.f90.

gen_be_vertloc

jjguerrette:
@jamiebresch this definitely looks like an improvement, but it seems like the vertical correlation length scale is too large relative to the 80-member ensemble.  Are the choices of 5 and 3 in the parameterization taken from GSI? Or from literature?  If not, then it seems to me there needs to be an adjustment still.

jamiebresch:
@jjguerrette 
The vertical localization does not increase the correlation that does not
exist in the ensemble perturbations.

I don't know how Dale Barker (the original vertloc developer) decided the
model-level based (a), (b), (c) function and kept (a) as the default.
I made up the (d) relation and it is totally empirical.
The logP method is taken from GSI. The 0.388 shown is the plot is also
a user-defined value. It can also be level-dependent rather than one
value for all levels.

I did not intend to implement a comprehensive vertical localization for WRFDA.
The main purpose is to replace the very bad default and raise the awareness
that users should look at their ensembles and decide their localization scales.

liujake:
@jamiebresch @davegill I am Ok to put this into V4.1 release repo for users' convenience, but think it is more appropriate to categorize this as 'enhancement' instead of 'bug-fix'.

liujake:
@jamiebresch Can you also plot default localization function at each level before/after your change?

jjguerrette:
While the correlations will not be increased, making the localization length scale longer than the correlation length scale defeats the purpose of doing localization.  The number of ensemble members required for convergence could actually increase.

jamiebresch:
@jjguerrette Yes, I agree with your comment.
When the ensemble size is large enough, the localization does not matter much.
I don't insist on the 5 and 3 values. The kscale relation should actually be user input values.

jamiebresch:
Red: kscale = 10.0 * real(k) * nk_inv
Blue: kscale = 5.0 + real(nk)/real(k+3)
vert_corr_levels

jjguerrette:
Sorry, I'm not completely right on this as the new localization clearly will remove some spurious correlations from a small ensemble.  However, I'm wondering whether the localization is filtering out enough of those noisy correlations?  For example, at a fixed number of ensembles (40? 80?), is there a degradation to the analysis when used in a forecast?

I'm fine with this change to go in, but it would be nice to have a more rigorous analysis.

liujake:
From the localization functions before/after change, this is a quite significant change for levels below 30. This will likely substantially affect the results. I would suggest this not going into V4.1 or develop branch before we have more evidence it is really an improvement to the forecast results.

jjguerrette:
@jamiebresch, have you had more experience with these vertical localization lengthscales since closing the PR?  I wonder if the lengthscales need to be increased for all-sky radiance experiments.  It is possible the vertical correlation of hydrometeors is getting truncated.

jjguerrette:
Please see attached my own vertical correlations analyzed from an 80 member ensemble over Eastern 2/3 of CONUS:
be_vert_corr

jamiebresch:
@jjguerrette I have finalized the proper formulation for a physically-based log-P method for vertical localization.  Dale also agrees that the new plot looks more reasonable.
vert_corr_20190416

jjguerrette:
That does look very nice!

liujake:
@jamiebresch You may reopen this PR to be merged in for this logP method if you like.

jamiebresch:
Will consider open a new PR in the future for a finalized better log-P implementation.

PR #851, opened 2019-04-01
Handle lowest-level microphysics species in real.exe

### LINKED ISSUE #837: initialize_real of microphysics variables; need to fix model lowest level not from surface
Within dyn_em/module_initialize_real.F, there is a block similar to one I pasted below to perform the vertical interpolation of moist physics variables (cloud water, rain, snow, etc.).  Each one looks nearly the same for mixing ratio and number concentration variables.  They all have a fault of using the "lowest_lev_from_sfc" variable similar to temperature or winds or humidity.  In fact, this should always be false as there are no 'surface' variables of microphysics so they result in becoming zero at model bottom K-level.  Simple fix is to enter a hardwired false or zero for lowest_lev_from_sfc.

CODE SNIPPET - there are a large number of these for the different species, one for mixing ratio, another for number concentration variables...
         IF ( flag_qr .EQ. 1 ) THEN
            DO im = PARAM_FIRST_SCALAR, num_3d_m
               IF ( im .EQ. P_QR ) THEN
                  CALL vert_interp ( grid%qr_gc , grid%pd_gc , moist(:,:,:,P_QR) , grid%pb , &
                                     grid%hgtmaxw , grid%hgttrop , grid%pmaxw , grid%ptrop , &
                                     grid%pmaxwnn , grid%ptropnn , &
                                     0 , 0 , &
                                     config_flags%maxw_horiz_pres_diff , config_flags%trop_horiz_pres_diff , &
                                     config_flags%maxw_above_this_level , &
                                     num_metgrid_levels , 'Q' , &
                                     interp_type , linear_interp , extrap_type , &
                                     lowest_lev_from_sfc , use_levels_below_ground , use_surface , &
                                     zap_close_levels , force_sfc_in_vinterp , grid%id , &
                                     ids , ide , jds , jde , kds , kde , &
                                     ims , ime , jms , jme , kms , kme , &
                                     its , ite , jts , jte , kts , kte ) 
               END IF
            END DO
         END IF

gthompsnWRF:
TYPE:  bug fix

KEYWORDS: real, vertical interpolation, microphysics, lowest level

SOURCE:  Greg Thompson (NCAR/RAL)

DESCRIPTION OF CHANGES:  For microphysical mixing ratios and number concentrations (and
chemical species) we do not actually have values at the so-called surface level, so do not try using 
surface level for vertical interpolation.  RH is *NOT* affected as we do wish the water vapor to 
continue using standard method, but we shall never have mixing ratios or number concentrations of 
microphysical species at the "surface" level per se.

ISSUE: Fixes #837 

LIST OF MODIFIED FILES:
 dyn_em/module_initialize_real.F

TESTS CONDUCTED:  
1. From met_em.d0X files to wrfinput_d0X, I confirmed that microphysics mass mixing ratios and number concentrations on WRF first model level (K=1) were previously zero and now get a proper value from interpolation procedure.  
2. New results with rain number concentrations, similar to the input data.
Untitled

### DIFF
--- a/dyn_em/module_initialize_real.F
+++ b/dyn_em/module_initialize_real.F
@@ -1898,7 +1898,7 @@ integer::oops1,oops2
                                      config_flags%maxw_above_this_level , &
                                      num_metgrid_levels , 'Q' , &
                                      interp_type , linear_interp , extrap_type , &
-                                     lowest_lev_from_sfc , use_levels_below_ground , use_surface , &
+                                     .false. , use_levels_below_ground , use_surface , &
                                      zap_close_levels , force_sfc_in_vinterp , grid%id , &
                                      ids , ide , jds , jde , kds , kde , &
                                      ims , ime , jms , jme , kms , kme , &
@@ -1928,7 +1928,7 @@ integer::oops1,oops2
                                      config_flags%maxw_above_this_level , &
                                      num_metgrid_levels , 'Q' , &
                                      interp_type , linear_interp , extrap_type , &
-                                     lowest_lev_from_sfc , use_levels_below_ground , use_surface , &
+                                     .false. , use_levels_below_ground , use_surface , &
                                      zap_close_levels , force_sfc_in_vinterp , grid%id , &
                                      ids , ide , jds , jde , kds , kde , &
                                      ims , ime , jms , jme , kms , kme , &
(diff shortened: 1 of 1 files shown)


davegill:
@gthompsnWRF 
Greg,
gonna need a better title than "..." 
😄

gthompsnWRF:
@davegill  Guess I goofed in fixing the pull request title. Now that is fixed, thanks.

davegill:
These changes are only to the real program, where the vertical interpolation takes place. For the moist and scalar (in the WRF sense of the word scalar) arrays, the option is set to false that defines the lowest level in WRF from the surface input value from metgrid. Included are a few initializations for WRF Chem. Each of the modified lines was verified to ONLY modify a moist or scalar arrays.

I am good with this PR.

PR #869, opened 2019-04-04
Several "AFWA diagnostic" memory fixes

davegill:
TYPE: bug fix

KEYWORDS: AFWA, moist, scalar, package, memory

SOURCE: internal

DESCRIPTION OF CHANGES: 
1. For each routine that uses a particular moist of scalar 3d component, verify that component has 
been allocated. For example:
Original code:
```
      DO i=ims, ime
        DO k=kms, kme
          DO j=jms, jme
            qvapr(i,k,j) = moist(i,k,j,P_QV)
          ENDDO
        ENDDO
      ENDDO
```
New code:
```
    IF ( F_QV ) THEN
      DO i=ims, ime
        DO k=kms, kme
          DO j=jms, jme
            qvapr(i,k,j) = moist(i,k,j,P_QV)
          ENDDO
        ENDDO
      ENDDO
    END IF
```
2. When using dz on half levels, do not use z(k)-z(k-1), since k-1 gives you a 0 index. Use dz8w.
Original code:
```
                melt(i,j)=melt(i,j)+9.8*(((t_phy(i,k,j)-273.15)/273.15)* &
                          (z(i,k,j)-z(i,k-1,j)))
```
New code:
```
                melt(i,j)=melt(i,j)+9.8*(((t_phy(i,k,j)-273.15)/273.15)* &
                          (dz8w(i,k,j)))
```
3. Verify for each AFWA diagnostic option that if any domain is activated, all domains are activated.
4. Verify that if any AFWA diagnostic option is activated, the main afwa_diag_opt option is activated.
5. Turns out that the field number concentration graupel has never been filled, and was always 0.
```
    IF ( F_QNG ) THEN
      DO i=ims, ime
        DO k=kms, kme
          DO j=jms, jme
            ngraup(i,k,j) = scalar(i,k,j,P_QNG)
          ENDDO
        ENDDO
      ENDDO
    END IF
```
6. Move the max wind speed field to the severe option in the Registry packaging.
7. The icing option is only used for certain known MP schemes. Added positive IF test with ELSE:
```
          ELSE
            CALL wrf_debug ( DEBUG_LEVEL, 'Icing diagnostics not processed due to unknown MP scheme' )
```

LIST OF MODIFIED FILES:
M phys/module_diag_afwa.F
M Registry/registry.afwa

TESTS CONDUCTED: 
I could get the dz option to be trapped (and now work successively). 
Verification also that the memory troubles which caused code to hang are STILL fixed in this PR.
The warnings inside of module_diag_afwa.F are wrf_debug (level=1), so no one will actually see 
the warnings. Users will not know why they are getting zero-valued diagnostics, oy vey.

### DIFF
--- a/phys/module_diag_afwa.F
+++ b/phys/module_diag_afwa.F
@@ -112,6 +112,7 @@ CONTAINS
     TYPE(WRFU_Time) :: hist_time, aux2_time, CurrTime, StartTime
     TYPE(WRFU_TimeInterval) :: dtint, histint, aux2int
     LOGICAL :: is_after_history_dump, is_output_timestep, is_first_timestep
+    INTEGER , PARAMETER :: DEBUG_LEVEL = 1
 
     ! Chirp the routine name for debugging purposes
     ! ---------------------------------------------
@@ -284,6 +285,8 @@ CONTAINS
           ENDDO
         ENDDO
       ENDDO
+    ELSE
+      CALL wrf_debug ( DEBUG_LEVEL, 'Option rh requires: QV' )
     END IF
     IF ( F_QV .AND. F_QC .AND. F_QI ) THEN
       DO i=ims,ime
@@ -293,6 +296,8 @@ CONTAINS
           ENDDO
         ENDDO
       ENDDO
+    ELSE
+      CALL wrf_debug ( DEBUG_LEVEL, 'Option rh_cld requires: QV, QC, QI' )
     END IF
 
     ! Time-step precipitation (convective + nonconvective)
@@ -317,6 +322,8 @@ CONTAINS
                                             rho(i,kms:kme,j) )
         ENDDO
       ENDDO
+    ELSE
+      CALL wrf_debug ( DEBUG_LEVEL, 'Option pwat requires: QV, QC' )
     END IF
 
     ! After each history dump, reset max/min value arrays
@@ -529,6 +536,8 @@ CONTAINS
                                , k_start, k_end                   &
                                , j_start, j_end, i_start, i_end   )
 
+      ELSE
+        CALL wrf_debug ( DEBUG_LEVEL, 'Option severe_wx_diagnostics requires: QV, QR, QC, QI, QS, QG, QNG' )
       END IF
     ENDIF   ! afwa_severe_opt == 1
 
@@ -556,6 +565,7 @@ CONTAINS
                              , grid % t2                        &
                              , rh                               &
                              , grid % z                         &
+                             , dz8w                             &
                              , grid % ht                        &
                              , grid % afwa_precip               &
                              , grid % swdown                    &
@@ -820,6 +830,8 @@ CONTAINS
                                , ims, ime, jms, jme, kms, kme     &
                                , ips, ipe, jps, jpe, kps, kpe )
         ENDIF  ! afwa_icing_opt
+      ELSE
+        CALL wrf_debug ( DEBUG_LEVEL, 'Option icing_diagnostics requires: QC, QR, QNC' )
       END IF
 
       ! Calculate visiblility diagnostics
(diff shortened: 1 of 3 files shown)


weiwangncar:
@davegill Thanks for taking this a step further. If we won't allow user to go into those routines because of lack of variables, then perhaps we could add some thing like 
ELSE
     write ( message, * ) 'Lack arguments to call severe_wx_diagnostics'
     CALL wrf_debug( 0 , message )

so that the prints tell the user they are not going to get these diagnostics.

davegill:
@weiwangncar 
>Thanks for taking this a step further. If we won't allow user to go into those routines because of lack of variables, then perhaps we could add some thing like
ELSE
write ( message, * ) 'Lack arguments to call severe_wx_diagnostics'
CALL wrf_debug( 0 , message )
so that the prints tell the user they are not going to get these diagnostics.

Wei,
That's a really good idea

phys/module_diag_afwa.F
```diff
@@ -236,28 +306,30 @@ SUBROUTINE afwa_diagnostics_driver (   grid , config_flags     &
...
-          grid % wspd10max(i,j) = 0.
-          IF ( config_flags % afwa_severe_opt == 1 ) THEN
-          grid % afwa_llws(i,j) = 0.
-          ENDIF
+      IF ( config_flags % afwa_severe_opt == 1 ) THEN
+        DO j = jms, jme
+          DO i = ims, ime
+            grid % wspd10max(i,j) = 0.
+            grid % afwa_llws(i,j) = 0.
+          ENDDO
```
weiwangncar: @davegill wspd10max needs to be outside afwa_severe_opt, this one needs to be computed all the time.

weiwangncar: @davegill Or we can move wspd10max to be packaged with afwa_severe_opt in registry.afwa.

davegill:
The warnings are wrf_debug (level=1), so no one will actually see the warnings. Users will not know
why they are getting zero-valued diagnostics.

phys/module_diag_afwa.F
davegill: @weiwangncar 
Wei,
I think that I fixed this. I made the afwa_diag_opt always active if any other AFWA diag is active. So the wspd10max is always available. Let me know if this addresses your concern adequately.

dudhia:
That would just be a fix until we figure out something better. It allows it
to go in V4.1.

davegill:
Folks,
From about line 785 to 885 you can see that we only have a few MP schemes listed.
What do we do here?

phys/module_diag_afwa.F
weiwangncar: @davegill When only afwa_diag_opt is active, then this can still be a problem.

weiwangncar:
Which routine are we talking about here?

phys/module_diag_afwa.F
```diff
@@ -236,28 +311,32 @@ SUBROUTINE afwa_diagnostics_driver (   grid , config_flags     &
...
         ENDDO
-      ENDDO
+      ENDIF
     ENDIF 
 
     ! Calculate the max 10 m wind speed between output times
```
weiwangncar: @davegill The do loop just after these comments need to be put in 
IF ( config_flags % afwa_severe_opt == 1 ) THEN
ENDIF

phys/module_diag_afwa.F
davegill: @weiwangncar 
Wei,
I think that I have _now_ fixed the case where you were still getting incorrect memory access

phys/module_diag_afwa.F
davegill: @weiwangncar 
Wei,
Sorry, put the comment in the wrong place. I think I fixed this now by including an IF test around this larger block that encompasses the wspd10max location.

weiwangncar:
@davegill It WORKS now!

PR #870, opened 2019-04-05
Provide initial radiative effective size for clouds for very first radiation time step

gthompsnWRF:
TYPE:  enhancement

KEYWORDS: radiation, cloud drop size, ice size, initialization

SOURCE: Greg Thompson (gthompsn)

DESCRIPTION OF CHANGES:  Immediately at first time step, subroutine phy_init will provide default very tiny values of radiative effective size (radius) of cloud water, ice, and snow that get used at the first radiation time step.  However, since radiation is called prior to microphysics, the values are far from reality and won't receive the microphysics-based size until the 2nd radiation time step.  So the first time step is quite out of synch with subsequent ones.  Using this new enhancement, the initial sizes are based on microphysics principals similar to other codes to give better initial values for that first radiation step.

ISSUE:  not noted previously

LIST OF MODIFIED FILES:
 dyn_em/start_em.F
 phys/module_physics_init.F

TESTS CONDUCTED:  A test simulation run for 5 minutes with one minute radiation time step and one minute history file output showed the initial radius of cloud variables was better than older defaults.

 - [ ] Needs to have moist vs dry theta if test

### DIFF
--- a/dyn_em/start_em.F
+++ b/dyn_em/start_em.F
@@ -992,6 +992,8 @@ endif
                       grid%w0avg, grid%rainnc, grid%rainc, grid%raincv, grid%rainncv,  &
                       grid%snownc, grid%snowncv, grid%graupelnc, grid%graupelncv,  &
                       z_at_q, grid%alt, grid%qnwfa2d, scalar(ims,kms,jms,1), num_scalar,         & ! G. Thompson
+                      moist(ims,kms,jms,1), num_moist,                  & ! G. Thompson
+                      grid%t_1, grid%p_hyd,                             & ! G. Thompson
                       grid%re_cloud, grid%re_ice, grid%re_snow,         & ! G. Thompson
                       grid%has_reqc, grid%has_reqi, grid%has_reqs,      & ! G. Thompson
                       grid%re_cloud_gsfc, grid%re_ice_gsfc,             & ! Goddard
--- a/phys/module_physics_init.F
+++ b/phys/module_physics_init.F
@@ -1187,7 +1187,7 @@ CONTAINS
                 N_ice = 0.
              ENDIF
              IF (Q_ice .GT. 0) THEN
-                temper = theta(i,k,j) * (pressure(i,k,j)/p00)**RCP
+                temper = theta(i,k,j) * (pressure(i,k,j)/p00)**(R_d/Cp)
                 re_snow(i,k,j) = calc_effectRad_Ice (Q_ice, N_ice, inv_dens(i,k,j), temper)
              ENDIF
           end do
--- a/phys/module_physics_init.F
+++ b/phys/module_physics_init.F
@@ -46,6 +46,8 @@ CONTAINS
                          W0AVG, RAINNC, RAINC, RAINCV, RAINNCV,  &
                          SNOWNC, SNOWNCV, GRAUPELNC, GRAUPELNCV, &
                          z_at_q, inv_dens, qnwfa2d, scalar, num_sc,  & ! G. Thompson
+                         moist, num_qm,                          & ! G. Thompson
+                         theta, pressure,                        & ! G. Thompson
                          re_cloud, re_ice, re_snow,              & ! G. Thompson
                          has_reqc, has_reqi, has_reqs,           & ! G. Thompson
 #if ( EM_CORE == 1 )
@@ -371,6 +373,7 @@ CONTAINS
    REAL, DIMENSION(ims:ime, kms:kme, jms:jme), INTENT(INOUT) :: re_cloud, &
              re_ice, re_snow
    INTEGER, INTENT(INOUT):: has_reqc, has_reqi, has_reqs
+   REAL:: Q_cld, N_cld, Q_ice, N_ice, temper
 #if ( EM_CORE == 1 )
    REAL, DIMENSION(ims:ime, kms:kme, jms:jme), INTENT(INOUT) :: re_cloud_gsfc, &
              re_ice_gsfc, re_snow_gsfc, re_graupel_gsfc, re_hail_gsfc, re_rain_gsfc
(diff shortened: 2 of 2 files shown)


dudhia:
I notice the effective radius routine is added to the physics_init module. Is it sufficiently general for all microphysics options? I wonder about g_ratio and whether that is specific to Thompson in some way.

dyn_em/start_em.F
```diff
@@ -992,6 +992,8 @@ SUBROUTINE start_domain_em ( grid, allowed_to_read &
                       grid%w0avg, grid%rainnc, grid%rainc, grid%raincv, grid%rainncv,  &
                       grid%snownc, grid%snowncv, grid%graupelnc, grid%graupelncv,  &
                       z_at_q, grid%alt, grid%qnwfa2d, scalar(ims,kms,jms,1), num_scalar,         & ! G. Thompson
+                      moist(ims,kms,jms,1), num_moist,                  & ! G. Thompson
```
davegill: @gthompsnWRF 
Greg,
Instead of `moist(ims,kms,jms,1)`, why not just pass in `moist`?
I do not see a reason to do otherwise.

dudhia: This is the first time we have needed to pass the 4d moist array into phy_init, so I am wondering about that. It seems to be a big step.

gthompsnWRF: @davegill  yes, that is correct as well. I just followed the same as the scalar argument, so would you prefer a match or both changed?

@dudhia  while the question you ask about g_ratio is true, how would you suggest a different way?  Both Hugh Morrison and I use this almost identically.  The difference with gamma shape parameter as compared to using 2.5 microns drop size is dramatic. Changing "mu" in gamma distribution would be rather small difference compared to the setting of 2.5 microns. Otherwise, the idea of calling how many different microphysics subroutines to align the cloud sizes with a microphysics scheme is going to get really nasty.  This is part of the larger picture of how future codes have to consider a plethora of connectivities  -none of which are trivial to resolve.  One could argue we need to call separate routines for NSSL, P3, Milbrandt, Goddard, Thompson, the list is crazy long and my own thought is this enhancement is way better than existing code and quite simple and isolated using the functions I built.

davegill:
@gthompsnWRF 
Greg,
In your commit message you say

> similar to other codes

Can you be more specific.

dyn_em/start_em.F
davegill: Greg,
Do moist correctly. Pass it in as just "moist".  After you do your PR, we'll clean up scalar with a second PR. Also, if moist is an intent(in) variable only (which appears to be the case), please modify the declaration.

davegill:
@gthompsnWRF @dudhia 
Can we follow this through please to get to a decision

phys/module_physics_init.F
```diff
@@ -1115,37 +1120,82 @@ SUBROUTINE phy_init ( id, config_flags, DT, restart, zfull, zhalf,     &
...
+             IF (P_QI .GT. 1) THEN
+                Q_ice = MAX(0., moist(i,k,j,P_QI))
+             ENDIF
+             IF (P_QNI .GT. 1) THEN
+                N_ice = MAX(0., scalar(i,k,j,P_QNI))
+             ELSE
+                N_ice = 0.
+             ENDIF
+             IF (Q_ice .GT. 0) THEN
+                temper = theta(i,k,j) * (pressure(i,k,j)/p00)**(R_d/Cp)
```
weiwangncar: @gthompsnWRF @davegill Is this 'theta' always dry, or could it be 'theta_m'?

davegill: @weiwangncar @gthompsnWRF 
Wei,
GREAT catch, I completely skipped past that issue.

Greg two points:
1. The theta array is dummy value associated the argument field grid%t_1. Depending on config_flags%use_theta_m (= 0 (dry), or = 1 (moist)), the grid%t_1 field can be either a moist or a dry theta.
2. The grid%t_1 field is a perturbation, not a full field.

davegill:
@gthompsnWRF @weiwangncar @dudhia
Greg,
I took this code from a diagnostics routine. It will return the model dry theta.
```
               IF ( use_theta_m .EQ. 1 ) THEN
                  potential_t(i,k,j) =  ( t(i,k,j) + t0 )  / ( 1. + (R_v/R_d) * qv(i,k,j) )
               ELSE
                  potential_t(i,k,j) =   t(i,k,j) + t0
               END IF
```
In here, 
1. `t(i,k,j)` comes from grid%t_2. This is either the moist or dry perturbation potential temperature, (K). 
2. `Qv` is the ususal mixing ratio of vapor: `moist(:,:,:,P_Qv)`, (kg/kg).
3. `t0`, `Rv`, and `Rd` come straight from module_module_constants.
4. `use_theta_m` is from `config_flags%use_theta_m`.

davegill:
@gthompsnWRF 
Greg,
A recap:
1. The moist array should be passed in as moist, not moist(ims,kms,jms,1). 
2. We will fix scalar array in a diff PR.
3. Fix the temperature computation using correct theta perturbation.
4. Tweak PR message to address "similar to other codes" meaning
5. Demonstrate what this PR does. Show a left/right plot of some impacted field before/after.

smileMchen:
@dudhia @weiwangncar 
Wei and Jimy,
Would you please take a look at this PR and make sure whether we can include it in WRFV4.1.1 release?

dudhia:
Should be OK.

davegill:
@gthompsnWRF @dudhia @weiwangncar 
Greg,
We are OK in principal. The algorithm needs to include the necessary mods to handle 
1. the fact the grid%t_2 is a perturbation potential temperature
2. the field can be moist or dry

We are probably going to push out the 4.1.1 release by mid next week. If this goes into release-v4.1.2, that is just fine.

davegill:

PR #745, opened 2018-12-28
Improve memory usage for urban models

barlage:
Improve memory usage for urban models

TYPE: enhancement with no answer changes

KEYWORDS: urban model

SOURCE: Michael Barlage (NCAR)

DESCRIPTION OF CHANGES:

Multi-layer urban models pack 4D, 5D, and 6D arrays into 3D. The old code used one dimension (num_urban_layers) which depends on the largest 6D array. The new code creates separate mapping for the different urban arrays so that the arrays are maximally filled for each urban array. 

Compared to the original code, the memory cost (relative to a non-urban run) of using UCM is reduced from 31% to 5%, BEP from 31% to 15% and BEM from 707% to 274%, equivalent to a 20%, 45% and 64% reduction in memory used, respectively.

When users run BEP or BEM, new namelist values for the urban dimensions must be set.

LIST OF MODIFIED FILES: 

M       Registry/Registry.EM_COMMON
M       Registry/registry.dimspec
M       main/depend.common
M       share/module_check_a_mundo.F
M       dyn_em/start_em.F
M       dyn_em/module_first_rk_step_part1.F
M       dyn_nmm/start_domain_nmm.F
M       dyn_nmm/module_PHYSICS_CALLS.F
M       phys/module_physics_init.F
M       phys/module_surface_driver.F
M       phys/module_sf_noahdrv.F
M       phys/module_sf_noahmpdrv.F
M       phys/module_sf_urban.F
M       phys/module_sf_bep.F
M       phys/module_sf_bep_bem.F
M       wrftladj/start_em_ad.F
M       wrftladj/module_physics_init_ad.F
M       wrftladj/start_em_tl.F
M       wrftladj/module_physics_init_tl.F

TESTS CONDUCTED:
1. 1-hour summer/winter simulations produces no answer changes

PR #877, opened 2019-04-09
Require user to specify BEP/BEM run-time dimensions

smileMchen:
TYPE: bug fix

KEYWORDS: urban physics, BEP, BEM

SOURCE: internal

DESCRIPTION OF CHANGES: When BEP or BEM urban physics modules are turned on, values of 
several dimension variables need to be specified correctly in namelist.input. Their default values 
(set to 1) in the Registry only work for the UCM module. These values of these variables are checked 
in the check_a_mundo file when BEP/BEM options are turned on. If these variables are still the 
default values from the Registry (wrong), a fatal error informs the user to refer to 
test/em_real/examples.namelist for correct settings. 

These modifications are required with the commit SHA a500a27c, PR #745 "Improve memory usage 
for urban models".

LIST OF MODIFIED FILES:
M    share/module_check_a_mundo.F
M    test/em_real/examples.namelist

TESTS CONDUCTED: Three test cases have been done.

(1) BEP is turned on with incorrect values of the dimension variables. 

```
taskid: 0 hostname: r2i7n27
Quilting with   1 groups of   0 I/O tasks.
 Ntasks in X           12 , ntasks in Y           12
--- ERROR: num_urban_ndm is wrong
--- Please refer to example.namelist for correct setting of num_urban_ndm
  --- Please refer to example.namelist for correct setting of num_urban_ndm
--- ERROR: num_urban_nz is wrong
--- Please refer to example.namelist for correct setting of num_urban_nz
  --- Please refer to example.namelist for correct setting of num_urban_nz
--- ERROR: num_urban_ng is wrong
--- Please refer to example.namelist for correct setting of num_urban_ng
  --- Please refer to example.namelist for correct setting of num_urban_ng
--- ERROR: num_urban_nwr is wrong
--- Please refer to example.namelist for correct setting of num_urban_nwr
  --- Please refer to example.namelist for correct setting of num_urban_nwr
-------------- FATAL CALLED ---------------
FATAL CALLED FROM FILE:  <stdin>  LINE:    1965
NOTE:       4 namelist settings are wrong. Please check and reset these options
-------------------------------------------
```

(2) BEM is turned on with incorrect values:

```
taskid: 0 hostname: r1i4n0
Quilting with   1 groups of   0 I/O tasks.
 Ntasks in X           12 , ntasks in Y           12
--- ERROR: num_urban_ndm is wrong
--- Please refer to example.namelist for correct setting of num_urban_ndm
[...]

PR #878, opened 2019-04-10
Use urban modules to define run-time configuration dimensions

davegill:
TYPE: enhancement

KEYWORDS: BEP, BEM, urban, dimensions

SOURCE: internal

DESCRIPTION OF CHANGES: 
Motivation: 
With the effort to reduce memory for the urban models (SHA a500a27c, PR #745 "Improve 
memory usage for urban models"), the previously existing hard-coded array dimensions no 
longer exist in the Registry. This PR uses the existing INTEGER PARAMETERs in each of the 
urban models (BEP and BEM) to fill in _derived_ rconfig Registry entries. Users never have to 
be concerned that the namelist sizes of the arrays are inconsistent with the physics scheme.

1. Add access functions for each of BEP and BEM schemes
    a. The access functions pick up the dimensions for urban arrays
    b. The functions are within the files, but not within the modules
2. Change the Registry so that the urban dimensions are `derived`, not `namelist,physics`
3. Remove unused `ulay` from dimension specification Registry
4. Add in a single print out line in the urban driver to explain  what that print-out is for
    a. By the way - this entire print-out construct is not thread safe, but it is only a print
5. For balance in the world, removed some remnants of debugging prints
6. Assign the correct dimensions urban in check_a_mundo, based on sf_urban_physics option

LIST OF MODIFIED FILES:
M Registry/Registry.EM_COMMON
M Registry/registry.dimspec
M phys/module_sf_bep.F
M phys/module_sf_bep_bem.F
M phys/module_sf_urban.F
M test/em_real/examples.namelist

TESTS CONDUCTED: 
1. Tested with sf_urban_physics = 1, and 2, and 3


davegill:
@barlage @dudhia @smileMchen 
Mike and Jimy,
Ming and I have worked together to put together two different PRs (this one and #877). Which do you want to go with?

barlage:
Though I don't completely understand the Fortran, I like #878 better. It removes the namelist requirements, which is a nice feature, and guarantees consistency with the BEP and BEM modules so if users want to change the module dimensions, the memory allocation dimensions are done automatically.

davegill:
@barlage 
Mike,
Thanks for the info

dudhia:
Looks clever. As long as it works, I'm fine with it.

davegill:
@dudhia @barlage @smileMchen 
> Looks clever. As long as it works, I'm fine with it.

Jimy,
Ming is going to do some simple tests to verify that everything is OK.

smileMchen:
@davegill 
Dave,
I run a test case with BEP, which is done successfully.

PR #879, opened 2019-04-10
Force positive definite with small mass and number from RRTMG K, avoids floating exception

davegill:
TYPE: bug fix

KEYWORDS: effective radius, radiation, exception

SOURCE: internal

DESCRIPTION OF CHANGES: 
Problem:
This subroutine computes an effective radius given the mass and the number for the cloud,
snow, and graupel. 
1. When a denominator is zero (in the first time step), this gives a divide-by-0 situation.
2. After a time step, (cloud number / cloud mass < 0) is a problem when using a real exponent.
3. The sum of snow and graupel is used as a denominator. Both can be zero.

Solution:
Use MAX function to make sure that positive definite values are >= zero, and make sure that 
denominators are >= a pre-defined small value.
1. Both the cloud mass and the cloud number now use a MAX function with a previously existing value
(r1 = 1.e-12). The `rqcmps` variable is local, so it is changed directly. The `nc` variable is INTENT(IN),
so the MAX function surrounds this variable (is used in-line) twice. 
2. The sum of qs and qg now has a minimum allowable value (r1 = 1.e-12). In the same loop, set the 
minimum allowable value for each qs and qg (both used in the numerator) to be 0.

LIST OF MODIFIED FILES: 
M module_ra_effective_radius.F

TESTS CONDUCTED: 
 - [x] Without mod, code failed in first time step, at (i,j) = (1,1).
 - [x] With mod, code runs through effective radius calculation OK.
 - [x] With mod, code completes the regression simulation (10 time steps) with full debugging.

MMM Classroom regtest; em_real, nmm, em_chem; GNU only

Individual tests:
SUCCESS_RUN_WRF_d01_em_real_32_em_chem_1
SUCCESS_RUN_WRF_d01_em_real_32_em_chem_2
SUCCESS_RUN_WRF_d01_em_real_32_em_chem_5
SUCCESS_RUN_WRF_d01_em_real_32_em_real_03FD
SUCCESS_RUN_WRF_d01_em_real_32_em_real_07NE
SUCCESS_RUN_WRF_d01_em_real_32_em_real_10
SUCCESS_RUN_WRF_d01_em_real_32_em_real_11
SUCCESS_RUN_WRF_d01_em_real_33_em_real_03FD
SUCCESS_RUN_WRF_d01_em_real_33_em_real_07NE
SUCCESS_RUN_WRF_d01_em_real_33_em_real_10
SUCCESS_RUN_WRF_d01_em_real_33_em_real_11
SUCCESS_RUN_WRF_d01_em_real_34_em_chem_1
SUCCESS_RUN_WRF_d01_em_real_34_em_chem_2
SUCCESS_RUN_WRF_d01_em_real_34_em_chem_5
SUCCESS_RUN_WRF_d01_em_real_34_em_real_03FD
SUCCESS_RUN_WRF_d01_em_real_34_em_real_07NE
SUCCESS_RUN_WRF_d01_em_real_34_em_real_10
SUCCESS_RUN_WRF_d01_em_real_34_em_real_11
SUCCESS_RUN_WRF_d01_nmm_real_32_nmm_nest_01
SUCCESS_RUN_WRF_d01_nmm_real_32_nmm_nest_03
SUCCESS_RUN_WRF_d01_nmm_real_32_nmm_nest_04a
SUCCESS_RUN_WRF_d01_nmm_real_32_nmm_nest_06
SUCCESS_RUN_WRF_d01_nmm_real_34_nmm_nest_01
SUCCESS_RUN_WRF_d01_nmm_real_34_nmm_nest_03
SUCCESS_RUN_WRF_d01_nmm_real_34_nmm_nest_04a
SUCCESS_RUN_WRF_d01_nmm_real_34_nmm_nest_06

Comparison tests:
SUCCESS_RUN_WRF_d01_em_real_32_em_real_03FD vs SUCCESS_RUN_WRF_d01_em_real_33_em_real_03FD status = 0
SUCCESS_RUN_WRF_d01_em_real_32_em_real_03FD vs SUCCESS_RUN_WRF_d01_em_real_34_em_real_03FD status = 0
Files SUCCESS_RUN_WRF_d01_em_real_32_em_real_07NE and SUCCESS_RUN_WRF_d01_em_real_33_em_real_07NE differ
SUCCESS_RUN_WRF_d01_em_real_32_em_real_07NE vs SUCCESS_RUN_WRF_d01_em_real_33_em_real_07NE status = 1
SUCCESS_RUN_WRF_d01_em_real_32_em_real_07NE vs SUCCESS_RUN_WRF_d01_em_real_34_em_real_07NE status = 0
SUCCESS_RUN_WRF_d01_em_real_32_em_real_10 vs SUCCESS_RUN_WRF_d01_em_real_33_em_real_10 status = 0
SUCCESS_RUN_WRF_d01_em_real_32_em_real_10 vs SUCCESS_RUN_WRF_d01_em_real_34_em_real_10 status = 0
SUCCESS_RUN_WRF_d01_em_real_32_em_real_11 vs SUCCESS_RUN_WRF_d01_em_real_33_em_real_11 status = 0
SUCCESS_RUN_WRF_d01_em_real_32_em_real_11 vs SUCCESS_RUN_WRF_d01_em_real_34_em_real_11 status = 0
SUCCESS_RUN_WRF_d01_nmm_real_32_nmm_nest_01 vs SUCCESS_RUN_WRF_d01_nmm_real_34_nmm_nest_01 status = 0
SUCCESS_RUN_WRF_d01_nmm_real_32_nmm_nest_03 vs SUCCESS_RUN_WRF_d01_nmm_real_34_nmm_nest_03 status = 0
SUCCESS_RUN_WRF_d01_nmm_real_32_nmm_nest_04a vs SUCCESS_RUN_WRF_d01_nmm_real_34_nmm_nest_04a status = 0
SUCCESS_RUN_WRF_d01_nmm_real_32_nmm_nest_06 vs SUCCESS_RUN_WRF_d01_nmm_real_34_nmm_nest_06 status = 0
SUCCESS_RUN_WRF_d01_em_real_32_em_chem_1 vs SUCCESS_RUN_WRF_d01_em_real_34_em_chem_1 status = 0
SUCCESS_RUN_WRF_d01_em_real_32_em_chem_2 vs SUCCESS_RUN_WRF_d01_em_real_34_em_chem_2 status = 0
SUCCESS_RUN_WRF_d01_em_real_32_em_chem_5 vs SUCCESS_RUN_WRF_d01_em_real_34_em_chem_5 status = 0

### DIFF
--- a/phys/module_ra_effective_radius.F
+++ b/phys/module_ra_effective_radius.F
@@ -238,8 +238,8 @@
      enddo
 !
      do k = kts,kte_in
-       qsqg(k) = qs(k)+qg(k)
-       re_qs(k) = (re_qs(k)*qs(k)+re_qg(k)*qg(k))/qsqg(k)
+       qsqg(k) = max(r1,qs(k)+qg(k))
+       re_qs(k) = (re_qs(k)*max(0.,qs(k))+re_qg(k)*max(0.,qg(k)))/qsqg(k)
      enddo     
    endif
 !
--- a/phys/module_ra_effective_radius.F
+++ b/phys/module_ra_effective_radius.F
@@ -140,6 +140,7 @@
 !
 ! minimum microphys values
 !
+   real, parameter          :: r0 = 0.
    real, parameter          :: r1 = 1.e-12
    real, parameter          :: r2 = 1.e-6
 !
@@ -181,7 +182,7 @@
      qcmps(k)  = qc(k)-qccps(k)
      rqcmps(k) = max(r1,qcmps(k)*rho(k))
      if(f_qnc) then
-       lammps    = 2.*cdm2*(pidnc*MAX(r1,nc(k))/rqcmps(k))**obmr
+       lammps    = 2.*cdm2*(pidnc*MAX(r0,nc(k))/rqcmps(k))**obmr
      else
        lammps   = (pidnc*nc0/rqcmps(k))**obmr
      endif
@@ -208,7 +209,7 @@
    if (has_qc) then
      do k = kts,kte_in
        if (rqc(k).le.r1.or.rnc(k).le.r2) cycle
-       lamc = 2.*cdm2*(pidnc*(MAX(r1,nc(k))+nccps(k))/rqc(k))**obmr
+       lamc = 2.*cdm2*(pidnc*(MAX(r0,nc(k))+nccps(k))/rqc(k))**obmr
        re_qc(k) =  max(2.51e-6,min(sngl(3.0/lamc),50.e-6))
      enddo
    endif
--- a/phys/module_ra_effective_radius.F
+++ b/phys/module_ra_effective_radius.F
@@ -239,7 +239,7 @@
 !
      do k = kts,kte_in
        qsqg(k) = max(r1,qs(k)+qg(k))
-       re_qs(k) = (re_qs(k)*max(0.,qs(k))+re_qg(k)*max(0.,qg(k)))/qsqg(k)
+       re_qs(k) = (re_qs(k)*max(r0,qs(k))+re_qg(k)*max(r0,qg(k)))/qsqg(k)
      enddo     
    endif
 !
--- a/phys/module_ra_effective_radius.F
+++ b/phys/module_ra_effective_radius.F
@@ -179,9 +179,9 @@
 !
      rqc(k) = max(r1,qc(k)*rho(k))
      qcmps(k)  = qc(k)-qccps(k)
-     rqcmps(k) = qcmps(k)*rho(k)
+     rqcmps(k) = max(r1,qcmps(k)*rho(k))
      if(f_qnc) then
-       lammps    = 2.*cdm2*(pidnc*nc(k)/rqcmps(k))**obmr
+       lammps    = 2.*cdm2*(pidnc*MAX(r1,nc(k))/rqcmps(k))**obmr
      else
        lammps   = (pidnc*nc0/rqcmps(k))**obmr
      endif
@@ -208,7 +208,7 @@
    if (has_qc) then
      do k = kts,kte_in
        if (rqc(k).le.r1.or.rnc(k).le.r2) cycle
-       lamc = 2.*cdm2*(pidnc*(nc(k)+nccps(k))/rqc(k))**obmr
+       lamc = 2.*cdm2*(pidnc*(MAX(r1,nc(k))+nccps(k))/rqc(k))**obmr
        re_qc(k) =  max(2.51e-6,min(sngl(3.0/lamc),50.e-6))
      enddo
    endif


davegill:
@weiwangncar @dudhia 
Wei and Jimy,
I only found this trouble when I turned on floating exceptions. Since we are changing the computation of a variable used by the radiation scheme, we can wait for v4.1.1. Seems late to jam these mods in.

weiwangncar:
Looks ok to me.

phys/module_ra_effective_radius.F
```diff
@@ -179,9 +179,9 @@ subroutine effectRad(t, qc, nc, qi, qs, qg, rho, qmin, t0c,                 &
 !
      rqc(k) = max(r1,qc(k)*rho(k))
      qcmps(k)  = qc(k)-qccps(k)
-     rqcmps(k) = qcmps(k)*rho(k)
+     rqcmps(k) = max(r1,qcmps(k)*rho(k))
      if(f_qnc) then
-       lammps    = 2.*cdm2*(pidnc*nc(k)/rqcmps(k))**obmr
+       lammps    = 2.*cdm2*(pidnc*MAX(r1,nc(k))/rqcmps(k))**obmr
```
dudhia: did this need a MAX? It seems that 0.**obmr should work.

dudhia: your description implies nc was < 0?

davegill: @dudhia @weiwangncar 
Jimy, 
Correct. The nc(k) was < 0 (the first value it flagged was order 10^-13). Would it be better to use zero instread of r1 (1.e-12) for the MAX?
New
```
MAX(r1,nc(k))
```
Newer
```
MAX(0.,nc(k))
```

dudhia:
In the case of just keeping it safe for **obmr, yes zero would be better.

phys/module_ra_effective_radius.F
davegill: @dudhia 
Jimy,
OK, the minimum of the numerator of the positive definite values is not set to zero. These were originally unprotected and could become negative. My first set of mods set them to be no smaller than a scheme-defined small value (1.e-12). Per your suggestions, I have now replaced the nc(k) value with MAX(0.,nc(k)) in both places where this field is used with an exponent.

davegill:
@weiwangncar @dudhia 
Wei and Jimy,
I'll do a short forecast to make sure that everything is OK. 

This code was introduced for v4.0 with the KIAPS RRTMG commit. Therefore, unintended side-effects are unlikely as only a single scheme uses this module. However, I am still reluctant to put this modification in the v4.1 release, and much more comfortable with v4.1.1.

dudhia:
Since it is not curing a particular blow-up, it would be OK with me too.
Meanwhile we can suggest it to the KIAPS people,
Jimy

PR #903, opened 2019-05-14
Fix WRFDA dm_bcast interfaces

jjguerrette:
TYPE: bug fix

KEYWORDS: WRFDA, BCAST, DMPAR, mpi, 4DVAR

SOURCE: Internal, JJ Guerrette

DESCRIPTION OF CHANGES:
Fix dependencies on wrf_dm_bcast_integer/real/string/bytes …
A missing use statement for da_wrf_interfaces in da_obs_io.f90 was causing the following compile-time error for a 4D-Var build:

```
catastrophic error: **Internal compiler error: segmentation violation signal raised** Please report this error along with the circumstances in which it occurred in a Software Problem Report.  Note: File and line given may not be explicit cause of this error.
compilation aborted for da_obs_io.f (code 1)

real    0m19.553s
user    0m18.652s
sys     0m0.668s
da.make:421: recipe for target 'da_obs_io.o' failed
make[1]: [da_obs_io.o] Error 1 (ignored)
```
The interfaces are updated in da_wrf_interfaces, including assumed-size array behavior similar to that originally defined in module_dm.f90.  All known missing use statements related to wrf_dm_bcast_* calls are corrected.

The error arose for intel/17.0.1 and mpt/2.19 on cheyenne.  Of note is that the compile error has not been noticed previously.  An alternative solution could be to notify intel, as the error message suggests.  It seemed more logical to put a non-ambiguous fix in place that references specific subroutine interfaces.

ISSUE: none

LIST OF MODIFIED FILES: 
M       var/build/depend.txt
M       var/da/da_gpseph/da_gpseph.f90
M       var/da/da_obs_io/da_obs_io.f90
M       var/da/da_par_util/da_par_util.f90
M       var/da/da_physics/da_physics.f90
M       var/da/da_radiance/da_radiance.f90
M       var/da/da_radiance/da_radiance1.f90
M       var/da/da_setup_structures/da_scale_background_errors_cv3.inc
M       var/da/da_test/da_test.f90
M       var/da/da_tools/da_wrf_interfaces.f90
M       var/da/da_transfer_model/da_transfer_model.f90
M       var/da/da_transfer_model/da_transfer_wrftoxb.inc

TESTS CONDUCTED: Compile works now.  Have not conducted WRFDA REGTEST. Only affects 4D-Var builds.


davegill:
@jjguerrette 
JJ,
Since this does appear to be a bug, maybe this could be targeted for 4.1.1 or 4.1.2? This is entirely under the purview of the DA group to decide.

jjguerrette:
This PR could be directed at v4.1.1 instead, but would only require this commit:
d96c28c4e9ddc66bcfdb1aad5414ccef70f9c14d

Unfortunately I branched off of develop instead of v4.1.1.  @liujake @davegill let me know what you prefer.

liujake:
@jjguerrette If this bug fix is also applicable to V4.1, it normally should be committed to V4.1.1 and then  merge it from v4.1.1 to develop at later time.

jjguerrette:
Ok, I will fix the PR today.

PR #906, opened 2019-05-15
Enforce conditional use of IEEE Fortran 2003 capabilities in v4.1 Goddard schemes

### LINKED ISSUE #905: ieee_arithmetic usage
WRF 4.1 unconditionally added uses of ieee_arithmetic in some places, e.g. https://github.com/wrf-model/WRF/blob/dd5c4c3afdc01a11f0ee4d37969b0617fd72435f/phys/module_checkerror.F#L72-L73

Pre-4.1 code uses a check:
https://github.com/wrf-model/WRF/blob/dd5c4c3afdc01a11f0ee4d37969b0617fd72435f/frame/module_clear_halos.F#L29-L31

I couldn't find anything in the changelog about dropping support for compiler (versions) without ieee_arithmetic (e.g. gfortran < 5). Can this be made explicit or otherwise checks be used everywhere?

davegill:
TYPE: bug fix
    
KEYWORDS: IEEE, GNU
    
SOURCE: reported by Maik Reichert, fixed internally
    
DESCRIPTION OF CHANGES:
Problem:
With the introduction of the Goddard MP and radiation schemes in release-v4.1, there
are locations where the IEEE capability for Fortran 2003 is unconditionally used. With the 
default GNU compiler on Linux, the WRF code will not build. Updating the compiler, while
free, does require that the user then rebuild libraries from scratch.
    
Solution:
The WRF code has had tests for using IEEE for a number of years (with the svn releases,
starting with v3.6). Since that release, during the `configure` step, if a small test code fails 
to build or run successfully, the `configure.wrf` file sets a cpp compile-time flag. This ifdef
infrastructure is now added to the new Goddard-related files that were recently included in 
the WRF repository (b60a5b9e7729).
    
ISSUE:
Fixes #905 "ieee_arithmetic usage"
    
LIST OF MODIFIED FILES:
M ./configure
M ./phys/module_checkerror.F
M ./phys/module_gocart_coupling.F
M ./phys/module_mp_gsfcgce_4ice_nuwrf.F
M ./phys/module_ra_goddard.F
    
TESTS CONDUCTED:
1. No harm - the code performs as expected with newer GNU compilers.
2. Code builds with an older GNU compiler (4.9.2), with this warning during the `configure` step:
```
************************** W A R N I N G ************************************
There are some IEEE Fortran 2003 features in WRF that your compiler does not
recognize. The IEEE function calls have been removed.
*****************************************************************************
```
    
RELEASE NOTE: The WRF code added unprotected IEEE functions in the source code with release-v4.1. Using existing infrastructure, those declarations and functions are now ifdef'ed out automatically. Note that this does remove some default error checking in the Goddard MP and radiation schemes.

### DIFF
--- a/configure
+++ b/configure
@@ -879,7 +879,7 @@ EOF
   rm -f ${foo} ${foo}.[cfo] 2> /dev/null
 fi
 
-# testing for Fortran 2003 IEEE signaling features
+# testing for Fortran 2003 IEEE features
 make fortran_2003_ieee_test > tools/fortran_2003_ieee_test.log 2>&1
 rm -f tools/fortran_2003_ieee_test.log
 retval=-1
@@ -893,8 +893,8 @@ if [ $retval -ne 0 ] ; then
   echo " "
   echo "************************** W A R N I N G ************************************"
   echo " "
-  echo "There are some Fortran 2003 features in WRF that your compiler does not recognize"
-  echo "The IEEE signaling call has been removed.  That may not be enough."
+  echo "There are some IEEE Fortran 2003 features in WRF that your compiler does not"
+  echo "recognize. The IEEE function calls have been removed."
   echo " "
   echo "*****************************************************************************"
 fi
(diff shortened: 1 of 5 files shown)


davegill:
@letmaik @kkeene44 @smileMchen 
Maik,
Does this PR fix the problem for you? If so, we will refer to this on the WRF forum (forum.mmm.ucar.edu).

letmaik:
Thanks, the fix works.

PR #911, opened 2019-05-16
Support for compiling on cygwin

DWesl:
TYPE: enhancement

KEYWORDS: cygwin, compiler, configure, library search path

SOURCE: DWesl (PSU)

DESCRIPTION OF CHANGES: 

The build configury did not understand the cygwin platform.  Since the platform is similar to Linux and BSD, it was relatively straightforward to add support.

I had to move some settings from the postamble to the preamble for the settings in configure.defaults to be honored.  I also had to allow a larger stack in the registry for that program not to crash.

Some libraries appear to have different names on Cygwin and on Linux, in particular, I cannot find a libhdf5_fortran, but libhdf5_hl appears to work just fine.  I have no idea if these are fully interchangeable on Linux.

I could not find code for libwavelet, so I removed that library from LIB_EXTERNAL

LIST OF MODIFIED FILES: 
arch/Config.pl
arch/configure.defaults
arch/postamble
arch/preamble
configure
external/io_netcdf/makefile
tools/Makefile

TESTS CONDUCTED: 
I succeeded in compiling and running the em_scm_xy case on Cygwin.

### DIFF
--- a/arch/configure.defaults
+++ b/arch/configure.defaults
@@ -1597,8 +1597,7 @@ RLFLAGS		=
 CC_TOOLS        =       $(SCC) 
 
 LIB_EXTERNAL    = \
-                     ../external/io_netcdf/libwrfio_nf.a CONFIGURE_NETCDF_PATH/lib/libnetcdf.lib \
-                     ../external/wavelet/libWavelet.a ../external/wavelet/lib_wavelet.a
+                     ../external/io_netcdf/libwrfio_nf.a CONFIGURE_NETCDF_PATH/lib/libnetcdf.lib
 ESMF_IO_LIB     =    ../external/esmf_time_f90/libesmf_time.a
 LIB_BUNDLED     = \
                      ../external/fftpack/fftpack5/libfftpack.a \
--- a/arch/configure.defaults
+++ b/arch/configure.defaults
@@ -1556,7 +1556,7 @@ CC_TOOLS        =      $(SCC)
 ###########################################################
 #ARCH    CYGWIN_NT i686, PGI compiler on Windows # serial smpar dmpar dm+sm
 #
-DESCRIPTION     =       PGI ($SFC/$SCC): Windows
+DESCRIPTION     =       PGI ($SFC/$SCC): Windows native
 DMPARALLEL      =       # 1
 OMPCPP          =       # -D_OPENMP
 OMP             =       # -mp -Minfo=mp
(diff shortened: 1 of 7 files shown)


configure
```diff
@@ -204,7 +204,7 @@ fi
 
 # If the user asked for classic netcdf, acquiesce to the request.
 
-if [ "`uname`" = "Linux" ] ; then
+if [ "`uname`" = "Linux" -o "`uname -o`" = "Cygwin" ] ; then
```
davegill: Is uname -o correct?

external/io_netcdf/makefile
```diff
@@ -50,7 +50,7 @@ diffwrf:                diffwrf.F90
 	$(FC) -c $(FFLAGS) diffwrf.f
 	@if [ \( -f ../../frame/wrf_debug.o \) -a \( -f ../../frame/module_wrf_error.o \) -a \( -f $(ESMF_MOD_DEPENDENCE) \) -a \( -f ../../frame/clog.o \) ] ; then \
 	  echo "diffwrf io_netcdf is being built now. " ; \
-          if [ \( -f $(NETCDFPATH)/lib/libnetcdff.a -o -f $(NETCDFPATH)/lib/libnetcdff.so \) ] ; then \
+          if [ \( -f $(NETCDFPATH)/lib/libnetcdff.a -o $(NETCDFPATH)/lib/libnetcdff.dll.a -o -f $(NETCDFPATH)/lib/libnetcdff.so \) ] ; then \
```
davegill: Does this require -f in front of the new dll file?

configure
DWesl: `uname` gives result `CYGWIN_NT-10.0`.  I could change that, but I'm not sure when the "10.0" part changes.

external/io_netcdf/makefile
DWesl: Yes, that would.

davegill:
@DWesl @mgduda @micurry
Anything that changes the build system needs to be thoroughly vetted, so be patient. Lets iterate back and forth.

First, I checked, and this is using the develop branch as a base, and you are targeting the develop branch - both what we want with this type of PR.

I would like to see this PR re-done (only 7 files, so not too arduous). Keep this PR active for reference, but make a new one. The new PR will be assembled from very specific commits, in specific order. If you have troubles at any step, clean it up before the actual commit is made to your branch.

Before each of the specific commits, a full test build of the WRF code on a non-CYGWIN machine will be required (clean -a, configure, compile), and a heavier duty test case than SCM. Take a look at:
https://github.com/davegill/wrf-coop/blob/master/README.md#with-the-second-image-build-three-containers
We want you to run ARW, NMM, and Chemistry builds, and one test case for each. For each of these test cases, we need serial and MPI, and a comparison of the results demonstrating bit-for-bit identical results. You will swap in your github repository in the Dockerfile. At each step, the non-CYGWIN results are to be maintained. After the last commit, the CYGWIN results should also work (of course, unavailable from inside a docker container). A single GNU compiler will suffice (handled inside of the docker container).

Here are the commits. Again, with each of these proposed commits, conduct the 6 tests (serial and MPI for each of ARW, NMM, and Chem) and the comparisons (the serial vs MPI for each of the 3 builds). When everything is OK, put that commit on your branch.
1. Change the preamble, postamble, and configure.defaults files
2. Change configure (check syntax question)
3. Change external/io_netcdf/makefile (check syntax question) 
4. Change tools/Makefile - explain in the PR comment what is peflags
5. Change Config.pl

We will likely need the mods in \#3 to also be put into other I/O libs, or we need a build time check that says "we can't use xxxxx" with CYGWIN. 

It seems that \#5 may require an additional IF test to determine if we are using Linux of CYGWIN, if there are troubles with the hdf5 library.

Before starting on these tasks, let's get all of the software engineers to sign off on this initial proposed set of steps.

configure
davegill: The -o option trims the "10.0" and changes the string from "CYGWIN_NT" to "Cygwin"?

DWesl:
On 5/16/2019 6:52 PM, Dave Gill wrote:
> ------------------------------------------------------------------------
> 
> In configure 
> <https://nam01.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fwrf-model%2FWRF%2Fpull%2F911%23discussion_r284926771&data=02%7C01%7Cdfw5129%40psu.edu%7C06d2926f23f745785a4408d6da512cd2%7C7cf48d453ddb4389a9c1c115526eb52e%7C0%7C0%7C636936439511069061&sdata=Rt%2BCJ0cPpT6VKwcTj8MiPPbeKxt%2FgG2T%2Bu1U4Aa%2BGL0%3D&reserved=0>:
> 
>> @@ -204,7 +204,7 @@ fi
>   
>   # If the user asked for classic netcdf, acquiesce to the request.
>   
> -if [ "`uname`" = "Linux" ] ; then
> +if [ "`uname`" = "Linux" -o "`uname -o`" = "Cygwin" ] ; then
> 
> The -o option trims the "10.0" and changes the string from "CYGWIN_NT" 
> to "Cygwin"?
> 

$ uname
CYGWIN_NT-10.0
$ uname -o
Cygwin

I could change it to "CYGWIN_NT-10.0", but "Cygwin" seems like it would 
be around longer, plus it looks more like "Linux".  I have no strong 
preference either way.

davegill:
@DWesl 
Nope, I agree that we want Cygwin

DWesl:
On 5/16/2019 6:45 PM, Dave Gill wrote:
> Anything that changes the build system needs to be thoroughly vetted, so 
> be patient. Lets iterate back and forth.

Definitely.
> 
> First, I checked, and this is using the develop branch as a base, and 
> you are targeting the develop branch - both what we want with this type 
> of PR.

That's a good start.
> 
> I would like to see this PR re-done (only 7 files, so not too arduous). 
> Keep this PR active for reference, but make a new one. The new PR will 
> be assembled from very specific commits, in specific order. If you have 
> troubles at any step, clean it up before the actual commit is made to 
> your branch.

I can do that
> 
> Before each of the specific commits, a full test build of the WRF code 
> on a non-CYGWIN machine will be required (clean -a, configure, compile), 
> and a heavier duty test case than SCM. Take a look at:
> https://github.com/davegill/wrf-coop/blob/master/README.md#with-the-second-image-build-three-containers 
> We want you to run ARW, NMM, and Chemistry builds, and one test case for 
> each. For each of these test cases, we need serial and MPI, and a 
> comparison of the results demonstrating bit-for-bit identical results.

To check bit-for-bit reproducability, that would be:
clean, configure, compile, run
cp output.nc old_output.nc
make changes
clean, configure, compile, run
diff output.nc old_output.nc && echo good

or did you have something else in mind?  Would I also need to check the 
nesting options, or should that be orthogonal to the changes considered 
here?

> You will swap in your github repository in the Dockerfile. At each step, 
> the non-CYGWIN results are to be maintained. After the last commit, the 
> CYGWIN results should also work (of course, unavailable from inside a 
> docker container). A single GNU compiler will suffice (handled inside of 
> the docker container).

This will be a tad harder.  It would be much more straightforward for me 
to do this without Docker.  I may be able to get docker on WSL, but I 
haven't checked that.
> 
> Here are the commits. Again, with each of these proposed commits, 
> conduct the 6 tests (serial and MPI for each of ARW, NMM, and Chem) and 
> the comparisons (the serial vs MPI for each of the 3 builds). When 
> everything is OK, put that commit on your branch.
>

ARW-Chem or NMM-Chem?  Is it only possible to do ARW-Chem?

On a related note, why not exercise the Chemistry code at the same time 
as the dynamical cores?  Ease of debugging?

Would the Chem test need to include every bit of chemistry I can find, 
or would the WRF-GHG configuration be sufficient?

>  1. Change the preamble, postamble, and configure.defaults files
>  2. Change configure (check syntax question)
>  3. Change external/io_netcdf/makefile (check syntax question)
>  4. Change tools/Makefile - explain in the PR comment what is peflags
>  5. Change Config.pl

This is straightforward.

peflags modifies the headers in a Windows-format executable, in this 
case to reserve extra space for the stack.
https://www.cygwin.com/cygwin-ug-net/setup-maxmem.html
It is at least somewhat related to the "-mcmodel=medium" family of options.
https://gcc.gnu.org/onlinedocs/gcc-7.4.0/gcc/x86-Options.html#index-mcmodel_003dmedium-1

> 
> We will likely need the mods in #3 
> to also be put into other I/O libs, or we need a build time check that 
> says "we can't use xxxxx" with CYGWIN.

The link line is exactly the same, just the name of the relevant file is 
different.  I could grep through the code for any mentions of ".so", 
".dylib", ".a", ".lib", or ".dll" and add similar changes to all of those.

The only place I'm seeing that might need adjustment is
external/io_grib_share/build/{library,tcl_script}_rules.mk
so I'll look into making sure I can get grib and grib2 output.
There's also
chem/KPP/kpp/kpp-2.1/int/oldies/ros3.c
but I don't think there's any reason to update the `ldd` output in that 
file.
> 
> It seems that #5 
> may require an additional IF test to determine if we are using Linux of 
> CYGWIN, if there are troubles with the hdf5 library.

It appears that `libhdf5_fortran` is available in upstream HDF5 sources 
but was not built for Cygwin.  In that case, I can compile my own 
version with both thread-safety and the fortran code, then build against 
that.  If it works, I can suggest the changes to the Cygwin maintainer

My understanding was that `libnetcdff` uses only code available in 
`libnetcdf`, which depends on `libhdf5` and possibly `libhdf5_hl`, so we 
might be able to drop this requirement entirely.

To exercise this branch I should be making very sure NetCDF4 output 
works.  Is there other HDF5-based output I should be including in this test?
> 
> Before starting on these tasks, let's get all of the software engineers 
> to sign off on this initial proposed set of steps.
> 

Good plan.

davegill:
@DWesl 
The test cases are small. They run in about 1 minute on my laptop.

arch/configure.defaults
```diff
@@ -1597,8 +1597,7 @@ RLFLAGS		=
 CC_TOOLS        =       $(SCC) 
 
 LIB_EXTERNAL    = \
-                     ../external/io_netcdf/libwrfio_nf.a CONFIGURE_NETCDF_PATH/lib/libnetcdf.lib \
-                     ../external/wavelet/libWavelet.a ../external/wavelet/lib_wavelet.a
```
DWesl: I couldn't find any way to generate an `external/wavelet/libWavelet.a`, so I removed that link here and did not add it for the Cygwin link line.  If that archive was moved somewhere rather than removed, I should change these to point to the new location.

davegill: This wavelet library looks to be part of the data assimilation. It is located in var/external.

DWesl:
On 6/2/2019 12:00 PM, Dave Gill wrote:
> ------------------------------------------------------------------------
> 
> In arch/configure.defaults:
> 
>> @@ -1597,8 +1597,7 @@ RLFLAGS		=
>   CC_TOOLS        =       $(SCC)
>   
>   LIB_EXTERNAL    = \
> -                     ../external/io_netcdf/libwrfio_nf.a CONFIGURE_NETCDF_PATH/lib/libnetcdf.lib \
> -                     ../external/wavelet/libWavelet.a ../external/wavelet/lib_wavelet.a
> 
> This wavelet library looks to be part of the data assimilation. It is 
> located in var/external.
> 

Okay.  I can update that location.  Should I do that just for Cygwin, or 
everywhere I can find it?

DWesl:
I reorganized the commits as requested; the result is #1003.

PR #913, opened 2019-05-17
Correct some instructions in README.namelist

smileMchen:
TYPE: text only

KEYWORDS:  README.namelist, NMM

SOURCE: internal

DESCRIPTION OF CHANGES: 
A few physics options are declared as "NMM only" in README.namelist, whereas they actually work 
for both ARW and NMM. 
Surface Layer: NCEP GFS (`sf_sfclay_physics=3`)
PBL: Hybrid EDMF GFS (`bl_pbl_physics=3`)
This PR corrects the information. 

LIST OF MODIFIED FILES: 
M    run/README.namelist

TESTS CONDUCTED: 
A single ARW case is successfully done with the options of sf_sfclay_physics=3 and bl_pbl_physics=3. Results look fine.

### DIFF
--- a/run/README.namelist
+++ b/run/README.namelist
@@ -656,7 +656,7 @@ Namelist variables for controlling the adaptive time step option:
                                      = 0, no surface-layer
                                      = 1, Revised MM5 Monin-Obukhov scheme (Jimenez, renamed in v3.6)
                                      = 2, Monin-Obukhov (Janjic) scheme
-                                     = 3, NCEP Global Forecast System scheme (NMM only)
+                                     = 3, NCEP Global Forecast System scheme (works for both ARW and NMM )
                                      = 4, QNSE surface layer
                                      = 5, MYNN surface layer
                                      = 7, Pleim-Xiu surface layer (ARW only)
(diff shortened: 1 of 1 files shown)


run/README.namelist
```diff
@@ -656,7 +656,7 @@ Namelist variables for controlling the adaptive time step option:
                                      = 0, no surface-layer
                                      = 1, Revised MM5 Monin-Obukhov scheme (Jimenez, renamed in v3.6)
                                      = 2, Monin-Obukhov (Janjic) scheme
-                                     = 3, NCEP Global Forecast System scheme (NMM only)
+                                     = 3, NCEP Global Forecast System scheme (works for both ARW and NMM )
```
davegill: Ming,
We only seem to comment when the scheme works for only a single dynamical core. In this mod and the next, remove the parenthetical comment.

davegill:
@smileMchen 
Ming,
Are there any problems in check_a_mundo that need to be addressed for this? If not, please add that info to the commit message, so everyone knows that issue was considered.

barlage:
@davegill Maybe this doesn't fit in this PR, but it seems to apply to the title anyway... after we added the functions to the new urban memory (PR #878 ), is the section of README.namelist that includes the urban dimensions (num_urban_ndm, etc.) still necessary?

weiwangncar:
I agree they should be removed. We should open another PR for this.

davegill:
@smileMchen 
Ming
>  Maybe this doesn't fit in this PR, but it seems to apply to the title anyway... after we added the functions to the new urban memory (PR #878 ), is the section of README.namelist that includes the urban dimensions (num_urban_ndm, etc.) still necessary?

 - [ ] Ming would you take a look at this query?

PR #914, opened 2019-05-17
Correct variable in the package of "do_trad_fields"

smileMchen:
TYPE: bug fix

KEYWORDS: do_trad_fields, diag_nwp2, traditional fields

SOURCE: internal

DESCRIPTION OF CHANGES:  
Problem:
The variable `theta` is not a state variable, but it was incorrectly used as a variable in the package 
for "do_trad_fields". 

Solution:
This PR removes `theta` from the package list, and replaces `theta` with the correct variable 
`potential_t`.  

LIST OF MODIFIED FILES:
Registry/registry.trad_fields

TESTS CONDUCTED: 
A single case with the option `diag_nwp2=1` (the option to output the traditional fields) is 
conducted. The model runs fine and the results are reasonable.

### DIFF
--- a/Registry/registry.trad_fields
+++ b/Registry/registry.trad_fields
@@ -25,4 +25,4 @@ state    real   rh            ikj   dyn_em    1  -   h1  "RH"           "relativ
 #  Package declarations
 
 package   skip_trad_fields   diag_nwp2==0  -   -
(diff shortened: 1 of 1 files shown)


davegill:
@smileMchen 
Ming,
It would be easy to do a before vs after. We should see a reasonable field and then the broken one that is (likely) all zeros. Would you do a simple ncview of this to show that it is correct now.

smileMchen:
@davegill 
Dave,
The variables (theta and potential_t) are not really used in the option diag_nwp2 = 1. We save memory if we pack potential_t in the package. There will be no difference in the plots.

PR #918, opened 2019-05-28
Prepare for WRF-v4.1.1 release

smileMchen:
TYPE: text only

KEYWORDS: README, release

SOURCE: internal

DESCRIPTION OF CHANGES: 
Update the top-level README file, and the inc/version_decl file with the correct version.
Now using a string which in no longer a fixed length.

LIST OF MODIFIED FILES: 
M	   README
M	   inc/version_decl

TESTS CONDUCTED:  A test case is run with the correct printout
```
taskid: 0 hostname: r5i2n23
Quilting with   1 groups of   0 I/O tasks.
 Ntasks in X           12 , ntasks in Y           12
WRF V4.1.1 MODEL
 Parent domain
 ids,ide,jds,jde            1         401           1         261
 ims,ime,jms,jme           -4          41          -4          29
 ips,ipe,jps,jpe            1          34           1          22
```
 - [x] Do a run and demonstrate that the printout is correct.

### DIFF
--- a/inc/version_decl
+++ b/inc/version_decl
@@ -1,1 +1,1 @@
-   CHARACTER (LEN=10), PARAMETER :: release_version = 'V4.1.1'
+   CHARACTER (LEN=*), PARAMETER :: release_version = 'V4.1.1'
--- a/inc/version_decl
+++ b/inc/version_decl
@@ -1,1 +1,1 @@
-   CHARACTER (LEN=10) :: release_version = 'V4.1.1      '
+   CHARACTER (LEN=10) :: release_version = 'V4.1.1'
--- a/inc/version_decl
+++ b/inc/version_decl
@@ -1,1 +1,1 @@
-   CHARACTER (LEN=10) :: release_version = 'V4.1.1'
+   CHARACTER (LEN=10), PARAMETER :: release_version = 'V4.1.1'
--- a/inc/version_decl
+++ b/inc/version_decl
@@ -1,1 +1,1 @@
-   CHARACTER (LEN=10) :: release_version = 'V4.1      '
+   CHARACTER (LEN=10) :: release_version = 'V4.1.1      '
--- a/README
+++ b/README
@@ -1,4 +1,4 @@
-WRF Model Version 4.1
+WRF Model Version 4.1.1
 
 http://www2.mmm.ucar.edu/wrf/users/
 


inc/version_decl
```diff
@@ -1 +1 @@
-   CHARACTER (LEN=10) :: release_version = 'V4.1      '
+   CHARACTER (LEN=10) :: release_version = 'V4.1.1      '
```
mgduda: Here we're assigning a string of length 12 to a variable of declared length 10.

mgduda: I think it's valid Fortran to write something like this:
```fortran
CHARACTER (LEN=*), PARAMETER :: release_version = 'V4.1.1' 
```
This way, we don't need to worry about putting in the right number of spaces. However, I don't know whether there is downstream code that depends on `release_version` being exactly 10 characters or that needs to change `release_version`.

smileMchen: @mgduda 
Michael,
Thanks a lot. i have changed the code as suggested.

inc/version_decl
```diff
@@ -1 +1 @@
-   CHARACTER (LEN=10) :: release_version = 'V4.1      '
+   CHARACTER (LEN=10) :: release_version = 'V4.1.1'
```
davegill: Ming,
Michael suggested to also change the `LEN=10` to `LEN=*`.

Would you make this change, and then run the WRF model and look at the print out. We should see something like:
```
WRF V4.1.1 MODEL
```
Update the commit message that you have conducted this test.

davegill: @jamiebresch 
... and with a parameter statement, too

davegill:
@smileMchen 
Ming,
Do a run and demonstrate that the printout is correct.
Dave

smileMchen:
@davegill 
I run a test case and the printout is correct. Please approve this PR.
Ming

davegill:
@kkeene44 @smileMchen 
Kelly,
I _want_ to approve this, but the global owner is required.
I asked for a test, and Ming conducted it. Ming has modified the code to reflect Michael's suggestions.

smileMchen:
TYPE: text only

KEYWORDS: README, release

SOURCE: internal

DESCRIPTION OF CHANGES:
Update the top-level README file, and the inc/version_decl file with the correct version.
Now using a string which in no longer a fixed length.

LIST OF MODIFIED FILES:
M	README
M	inc/version_decl

TESTS CONDUCTED: A test case is run with the correct printout

taskid: 0 hostname: r5i2n23
Quilting with   1 groups of   0 I/O tasks.
 Ntasks in X           12 , ntasks in Y           12
WRF V4.1.1 MODEL
 Parent domain
 ids,ide,jds,jde            1         401           1         261
 ims,ime,jms,jme           -4          41          -4          29
 ips,ipe,jps,jpe            1          34           1          22

PR #919, opened 2019-06-02
nwp_diagnostics made compatible with adaptive time step

meteoadriatic:
TYPE: enhancement

KEYWORDS: nwp_diagnostic, adaptive time step

SOURCE: Ivan Toman (University of Zadar, Croatia)

DESCRIPTION OF CHANGES: 
Previously, the nwp_diagnostics feature worked properly only if fixed time step was used. 
If the adaptive time step option was used, resetting max arrays after output time failed 
because the original code expected a fixed number of time steps between output times. 

For this enhancement, the tracking of time after the output is used to determine when the
reset time has been reached. Resetting max arrays is performed after the output at the next 
time step by checking if 0 < if time from output < max_time_step. The details are given in 
code comments.

LIST OF MODIFIED FILES:
M       phys/module_diag_misc.F
M       phys/module_diagnostics_driver.F

TESTS CONDUCTED: 
1. Code compiled fine and produces expected results with single domain and with nested 
domain.
2. Regression is done successfully.
3. A test case with nwp_diagnostics = 1 and adaptive time step option turning on works fine. The 
figure attached displays the maximum 10m wind speed as one example.
WSPD10MAX

### DIFF
--- a/phys/module_diag_misc.F
+++ b/phys/module_diag_misc.F
@@ -56,9 +56,9 @@ CONTAINS
                      ,qg_curr                                         &
                      ,ng_curr,qh_curr,nh_curr,qr_curr,nr_curr         & !  Optional (gthompsn)
                      ,rho,ph,phb,g                                    &
+                     ,max_time_step,adaptive_ts                       &
                                                                       )
 !----------------------------------------------------------------------
-
   USE module_dm, ONLY: wrf_dm_sum_real, wrf_dm_maxval
   USE module_state_description, ONLY :                                  &
       KESSLERSCHEME, LINSCHEME, SBU_YLINSCHEME, WSM3SCHEME, WSM5SCHEME, &
@@ -263,6 +263,11 @@ CONTAINS
       REAL:: mvd_r, xslw1, ygra1, zans1
       INTEGER:: ng, n
 
+   REAL                                       :: time_from_output
+   INTEGER                                    :: max_time_step
+   LOGICAL                                    :: adaptive_ts
+   LOGICAL                                    :: reset_arrays = .FALSE.
+
 !-----------------------------------------------------------------
 ! Handle accumulations with buckets to prevent round-off truncation in long runs
 ! This is done every 360 minutes assuming time step fits exactly into 360 minutes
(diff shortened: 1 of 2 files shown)


davegill:
@meteoadriatic 
Ivan,
If you look in the dyn_em/solve_em.F routine, there are a few alarm variables (history, restart, boundary, and one for SST update). Should we consider using an alarm-type of logic for the diagnostic output? When we introduce more constructs with mod(real_1, real_2), the regional climate people complain that these do not work after the model advances a few years.

I have not reviewed the code carefully, but how does a 1.05 factor multiplied by the maximum possible time step work? We could be in an unstable period and stay entirely at the minimum time step.

meteoadriatic:
@davegill 
Dave, About first question. I believe that this should not be a problem because I retained exact the same logic if adaptive step is not used. So in that case, the original code is executed as before. The modified logic actually kicks in only if adaptive time step is chosen. So the change should not introduce new problems. Not sure if I completely understood that part of question and if this answers it.

For the second part, well that factor 1.05 is a workaround I came up with after first attemp that was not successful because of rounding error. In the module, we already have xtime variable (in minutes) that I use to determine how much seconds elapsed from output:

time_from_output = mod( xtime, REAL(history_interval) )

However, xtime variable is rounded and if it is rounded up in last decimals this leads to a problem.
This leads to situation, where time_from_output can be actually larger than max_time_step variable when time step is maximized (for example, time_from_output = 0.500032 instead of = 0.5 if time step is = 30s). So to fix that I just multiplied max_time_step a bit which shown to fix the problem. The same can be achieved by dividing time_from_output with similar value.

Now, when actual time steps are small what happens is this. For example if max_time_step = 30, and actual time step = 10 seconds, we will have arrays resetting on first three time steps after output. But, I don't think this is big inaccuracy, because regardless resets will occur multiple times after output, they still won't occur after max_time_step (times 1.05) time whatsoever so resets are still limited to small time window after output.

I hope this helps,
Best regards,
Ivan

smileMchen:
@meteoadriatic @davegill 
(1) Regression is done successfully.
(2) A test case with nwp_diagnostics = 1 and adaptive time step option turning on works fine.

The figure attached displays the maximum 10m wind speed as one example.
WSPD10MAX

meteoadriatic:
@smileMchen Sure,

I've set above:
`time_from_output = mod( xtime, REAL(history_interval) )`

So time_from_output is amount of time that elapsed after wrf wrote output into last wrfout file.

Now, in order to reset arrays that keep nwp_diagnostics max values, we need to check its value and see if it is larger than 0 (because if it is = 0, then wrf is at timestep of outputting to wrfout and if we reset them at that point, all nwp_diagnostics will be = 0.

Now, as explained in previous paragraph, we need to wait with resetting arrays up to the point when wrf writes them down (time_from_output > 0), but after wrf writes them down, we need to reset them at first opportunity we have. 

So what I came up with is to check if time_from_output is less than max_time_step value, this means, that just wrf did outputting to wrfout so this is perfect time to reset arrays.

However, now why * 1.05. Well that is bit hackish but I could not came up with better solution. This is needed because when code rounds time, the time_from_output in first timestep after outputting can actually become slightly larger than max_time_step and then resetting will not happen in that case. So I just englarged max_time_step by 5 percent in order to catch those situations also.

Does this explanation helps? Let me know if I can address it more.

PR #925, opened 2019-06-07
Jensen ISHMAEL microphysics: Bug fixes

graupely:
TYPE: bug fixes

KEYWORDS: code changes/bug fixes

DESCRIPTION OF CHANGES:
1) Added allocation for single-precision lookup tables which are defined from the double-precision ones. Without this, certain versions of PGI throw an error. 
2) Moved the calculation of the inherent growth ratio to prevent if from being undefined in certain instances.
3) Storing dsold (old deltastr value, or shape parameter), use it to diagnose shape during nucleation and aggregation (transfer between ice categories), and ensure that this value stays in bounds. This prevents the updated shape from going out of bounds in certain instances.
4) Moved the mass check for aggregation to occur to around the aggregation subroutine call to prevent the aggregate diagnostics, calculated just above, from being undefined. 
5) Added a fix to ensure that nucleation won't turn planar particles into columnar ones by limiting this shape to spherical (deltastr(cc)=1).  
6) Set a low limit on nucleation size to 2 microns.

LIST OF MODIFIED FILES:

TESTS CONDUCTED:
Tested serial intel, gnu, pgi.
Builds and runs correctly dmpar (ifort), and I get the same answer on a restart. 🎉

### DIFF
--- a/phys/module_mp_jensen_ishmael.F
+++ b/phys/module_mp_jensen_ishmael.F
@@ -20,7 +20,7 @@ module module_mp_jensen_ishmael
   ! See Jensen et al. 2017 (JAS), Jensen et al. 2018 (MWR), and Jensen et al. 2018 (JAS) for details.    !
   !                                                                                                      !
   ! Author:        Anders A. Jensen, NCAR-RAL, ajensen@ucar.edu, 303-497-8484                            ! 
-  ! Last modified: 29 March 2019                                                                         !
+  ! Last modified: 07 June 2019 (ISHMAEL version 0.1)                                                    !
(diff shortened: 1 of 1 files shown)


davegill:
@weiwangncar 
Wei,
Would you take a look at this PR from Anders. I am approving, but a physics person must review.

1. The description indicates that _most_ of these are bug fixes: single precision vs double precision causing errors, undefined values, etc.
2. The only changes are internal to a single file.
3. The modifications do not touch any subroutine interfaces.

weiwangncar:
I'm ok with the changes.

Issue #922, opened 2019-06-05
Restart runs not reproducible in nest when topo_shading is turned on


emilycollier:
Summary: Restart runs in a nest are not reproducible when topo_shading is turned on, because shadowmask has a different value after restart compared to the continuous simulation.

Build: WRF 4.1; compiler choice: 66

Bug Fix: Editing Registry.EM_COMMON to add shadowmask to restart files seems to fix the problem.

davegill:
@emilycollier 
Emily,
Would you be interested in putting a pull request together to get this fix into the code? The problem that you have identified seems fairly contained (likely only a change in the Registry, as you mention). 

If so, use the release-v4.1.2 branch as your starting point.

Then do a checkout -b to get to your shadow-fix branch. You can take a look at https://github.com/wrf-model/WRF/wiki/Workflow-for-WRF-Code-Modification for lots more info on the mechanics of the commit specifically for a bug-fix

We will likely iterate with you on the contents of the commit message (partially we iterate since "more eyes" is a good idea, and partially as you are relatively new to our pull request requirements). More info on the way to build a good message is available at https://github.com/wrf-model/WRF/wiki/Making-a-good-pull-request-message.

emilycollier:
@davegill 
Hi Dave,

Sure, I can do that. Note that I had actually tested adding shadowmask & ht_shad to the restart file, and assumed that the former was the fix. I checked, and it is actually ht_shad that makes restarts reproducible, while shadowmask makes no difference.

Emily

PR #931, opened 2019-06-14
Fix to make restarts reproducible in nests when topo_shading is used

emilycollier:
TYPE: bug fix

KEYWORDS: topographic shading, radiation, restart runs, nested domains

SOURCE: Emily Collier (FAU Erlangen-Nürnberg)

DESCRIPTION OF CHANGES: 

1. Added ht_shad to the restart files

Restart runs aren't reproducible in nested domains when topographic shading is turned on, because ht_shad is not initialised in grid points adjacent to the lateral boundary of the nest.

Zero values for ht_shad at affected grid points (all located one grid point away from the eastern boundary in the same patch) lead to different values of shadowmask and swnorm. Here is a diff of swnorm in D2 in the first affected time step for a restart at 04 UTC:
Screenshot 2019-06-14 at 15 59 08
Screenshot 2019-06-14 at 15 59 30

Here is swdown for reference:
Screenshot 2019-06-14 at 16 24 18

Adding ht_shad to the restart files makes restart runs reproducible for all domains when topo_shading is used.

ISSUE: #922

LIST OF MODIFIED FILES: 

TESTS CONDUCTED: WRF 3.9.1.1 and 4.1 for two different configurations and study areas.


dudhia:
Looks fine to me.

PR #933, opened 2019-06-17
Fix to allow proper QVAPOR analysis when mp_physics is not set in namelist.input

jamiebresch:
TYPE: bug fix

KEYWORDS: WRFDA, mp_physics, package, QVAPOR, physics_suite

SOURCE: Jamie Bresch (NCAR)

DESCRIPTION OF CHANGES:
Add back the packaging for mp_physics=-1 (added in commit 93a17919, but dropped
unintentionally in commit 81ca2ff0) to allow WRFDA to allocate qv in the cases when
(1) physics_suite is set and mp_physics is not set,
(2) both physics_suite and mp_physics are not set.
Commit 7427eac9 contains some information about physics_suite in WRFDA.

When mp_physics is not set in the namelist.input, its default -1 leads to no moist variable
(even qv) being allocated. Even if the user is able to identify the incorrect QVAPOR analysis,
it is almost impossible for the user to figure out why and how to solve the problem.
The problem can be avoided by setting proper mp_physics. For dual-resolution EnVar, my_physics
in both column 1 and column 2 must be set.

NOTE that this PR handles qv only.
More other consideration and tests are needed for cloud variables.

LIST OF MODIFIED FILES:
M       Registry/registry.var

TESTS CONDUCTED:
(1) with missing mp_physics setting, the fixed code gets proper QVAPOR analysis for one simple test case.

RELEASE NOTE: Fix to allow proper QVAPOR analysis when mp_physics is not set in namelist.input.

### DIFF
--- a/Registry/registry.var
+++ b/Registry/registry.var
@@ -526,6 +526,7 @@ package   nomoist_ad   mp_physics_ad==98            -             -
 package   warmrain_ad  mp_physics_ad==99            -             -
 
 #do "grep package Registry/Registry.EM_COMMON | grep mp_physics==", then remove non-moist lists
+package   mpnotset        mp_physics==-1               -             moist:qv
 package   passiveqv       mp_physics==0                -             moist:qv
 package   kesslerscheme   mp_physics==1                -             moist:qv,qc,qr
 package   linscheme       mp_physics==2                -             moist:qv,qc,qr,qi,qs,qg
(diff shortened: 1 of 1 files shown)


jjguerrette:
Thank you for catching this Jamie.  There is a "DO NOT MERGE" tag.  Is the testing all that is missing?

jamiebresch:
The PR message has been updated and is ready for review.

jjguerrette:
Were the hydrometeors supposed to be included in registry.var?  I see Jamie has a commit that adds them on 21 June, but they are not contained in the release-v4.1.2 branch.

PR #945, opened 2019-07-01
Better bucketing: remove as many buckets as possible, not just one by one

davegill:
TYPE: enhancement

KEYWORDS: buckets

SOURCE: internal

DESCRIPTION OF CHANGES: 
1. There is no need to initialize variables to zero if time==0. That is already done automatically.
2. Instead of removing a single bucket per 6 h period, why not remove as many buckets as are
possible? This allows for using smaller bucket sizes, reducing the risk of loss of significance 
due to floating point precision during the bucket accumulation period.

For each of the bucketed variables, the old code
```
   aclwdnbc(i,j) = aclwdnbc(i,j) - bucket_J 
   i_aclwdnbc(i,j) =  i_aclwdnbc(i,j) + 1
```
has been replaced with
```
   bc = INT(aclwdnbc(i,j)/bucket_J)
   aclwdnbc(i,j) = aclwdnbc(i,j) - bc*bucket_J
   i_aclwdnbc(i,j) =  i_aclwdnbc(i,j) + bc
```

LIST OF MODIFIED FILES: 
M phys/module_diag_misc.F

TESTS CONDUCTED: 
 - [ ] Will do bit-for-bit comparison once the release v4.1.2 has been merged into develop. A PR there fixes the use of the radiation buckets.

### DIFF
--- a/phys/module_diag_misc.F
+++ b/phys/module_diag_misc.F
@@ -262,7 +262,8 @@ CONTAINS
       DOUBLE PRECISION:: lamr, N0min
       REAL:: mvd_r, xslw1, ygra1, zans1
       INTEGER:: ng, n
-      INTEGER:: bc ! largest number of integer buckets that can be removed
+      INTEGER:: bc ! largest number of integer buckets that can be removed,
+                   ! this is used locally within about a dozen computations
 
 !-----------------------------------------------------------------
 ! Handle accumulations with buckets to prevent round-off truncation in long runs
--- a/phys/module_diag_misc.F
+++ b/phys/module_diag_misc.F
@@ -294,43 +294,43 @@ CONTAINS
       DO i=i_start(ij),i_end(ij)
         IF(acswupt(i,j) .gt. bucket_J)THEN
           bc = INT(acswupt(i,j)/bucket_J)
-          acswupt(i,j) = acswupt(i,j) - nb*bucket_J
+          acswupt(i,j) = acswupt(i,j) - bc*bucket_J
           i_acswupt(i,j) =  i_acswupt(i,j) + bc
         ENDIF
         IF(acswuptc(i,j) .gt. bucket_J)THEN
           bc = INT(acswuptc(i,j)/bucket_J)
-          acswuptc(i,j) = acswuptc(i,j) - nb*bucket_J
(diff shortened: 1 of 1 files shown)


dudhia:
I don't think it is every 6 hours, but only when it exceeds a bucket that it subtracts one. Or maybe that's just rain. Anyway that was the intended behavior. It is not going to jump 2 buckets between checks.

davegill:
@dudhia 
> I don't think it is every 6 hours, but only when it exceeds a bucket that it subtracts one. Or maybe that's just rain. Anyway that was the intended behavior. It is not going to jump 2 buckets between checks.

Jimy,
It is every 6 h that the check is performed in `phys/module_diag_misc.F`:
```
! This is done every 360 minutes assuming time step fits exactly into 360 minutes
   IF(bucket_mm .gt. 0. .AND. MOD(NINT(XTIME),360) .EQ. 0)THEN
```

And now that I look at this logic, you _must_ activate the rain bucket. If you only activate the radiation flux bucket, the code is never activated. 

So, another commit is about to land.

davegill:
@dudhia 
Jimy,
So I remember what I was thinking, here's likely what happens to the new logic:
```
   IF( ( bucket_mm .gt. 0. ) .OR. ( bucket_J .gt. 0. ) .AND. &
       ( MOD(NINT(XTIME),360) .EQ. 0 ) ) THEN
```

dudhia:
Typically the bucket values suggested only get exceeded with monthly
accumulations,
so it would be an extreme event that exceeds it in 6 hours, and for
radiation it should be
impossible.

davegill:
@dudhia 
Jimy,
The problem with a large bucket size concerns the original purpose of the buckets in the first place: loss of data. We do not want the individual value that is added to be more than 7 orders of magnitude smaller than the bucket size. 

While it is unlikely that _per time step_ accumulations of precip < 10E-5 mm are important, it is still roundoff. The value for fluxes is probably much too large. 
```
 bucket_mm                           = 100.
 bucket_J                            = 1.e9
```
A smaller bucket that can be emptied multiple times is a better solution.

dudhia:
The values are both based on monthly accumulations where the size for
radiation is probably about 1.e9.
I don't think a higher accuracy is needed.
Jimy

davegill:
Not required at this time.

PR #946, opened 2019-07-02
README.namelist: Adding new option, correct changed option

weiwangncar:
TYPE: text only

KEYWORDS: oml_hml0, and dfi_savehydmeteors

SOURCE: Jamie Bresch (NCAR/MMM), internal

DESCRIPTION OF CHANGES:
Two changes for README.namelist 
1. The namelist variable `oml_hml0` used to be used for constant initial ocean mixed layer depth 
(> 0) and  real-time ocean mixed layer depth (such as from HYCOM) (< 0). But since version 3.9, 
an option was added to use climatology ocean mixed layer depth by setting this variable to 0. This 
information is now added to the README file. 
2. The other correction is to namelist variable `dfi_radar`. It has been renamed to 
`dfi_savehydmeteors` since version 3.5, but that change was not reflected in the README file. 

Both are namelist options are correctly updated with this PR.

LIST OF MODIFIED FILES: 
run/README.namelist

TESTS CONDUCTED: 
No test needed.

### DIFF
--- a/run/README.namelist
+++ b/run/README.namelist
@@ -463,7 +463,8 @@ Namelist variables for controlling the adaptive time step option:
  dfi_fwdstop_hour                    = 12       ! two-digit month of stop time for forward DFI integration
  dfi_fwdstop_minute                  = 00       ! two-digit month of stop time for forward DFI integration
  dfi_fwdstop_second                  = 00       ! two-digit month of stop time for forward DFI integration
- dfi_radar                           = 0        ! DFI radar da switch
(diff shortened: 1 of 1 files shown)


dudhia:
Does the climatology come through geogrid? Is it monthly?

weiwangncar:
@dudhia This is a global dataset in intermediate format. We obtained this dataset from SY Hong.

cmattocks:
Perhaps another use case for Fortran 2003 enums? Demo code attached.
[enum.txt](https://github.com/wrf-model/WRF/files/3356390/enum.txt)

Issue #949, opened 2019-07-03
domain decomposition check fails on an unused domain


kwman44:
I run WRF v4.1 with a domain shut off using the grid_allowed namelist option.  Even though the domain is inactive, the domain decomposition is checked, and this fails (and stops the run) if a tile for this domain does not have enough grid points in X or Y.

I submit that this domain decomposition check (and perhaps other checks, too) should not be performed (or should not be fatal) on domains for which grid_allowed is set to False.

PR #950, opened 2019-07-03
Namelist checks only to be conducted on "allowed" grids

davegill:
TYPE: bug fix

KEYWORDS: grid_allowed, check_a_mundo

SOURCE: Kevin Manning (NCAR/MMM), internal

DESCRIPTION OF CHANGES:

The WRF model has a construct to permit users to bypass specified nested domains (and all of that 
nest's children) using the `grid_allowed` option. In the example below, the 3-domain run is really
only a 2-domain run.
```
&domains
 grid_allowed = T, F, T
 parent_id    = 0, 1, 1,
 max_dom      = 3,
/
```
Problem:
If a user activated this option, the namelist checker in `module_check_a_mundo.F` previously did not consider whether a domain was active. For an unused domain, the WRF model would return a fatal error if, for example, the number of grid cells for the unused domain was too small for the decomposition.

Solution:
For each of the domain-indexed namelist checks in `module_check_a_mundo.F`, the DO loop 
block has been modified to include a Fortran `CYCLE` statement. If that domain is inactive, no 
domain-indexed namelist checks are conducted.

Original loop
```
DO i = 1, model_config_rec % max_dom
   IF ( ( model_config_rec % e_we(i) /  model_config_rec % nproc_x .LT. 10 ) .OR. &
        ( model_config_rec % e_sn(i) /  model_config_rec % nproc_y .LT. 10 ) ) THEN
      WRITE ( wrf_err_message , * ) 'For domain ',i,', the domain size is too small for this many processors, ', &
                                    'or the decomposition aspect ratio is poor.'
```

New loop
```
DO i = 1, model_config_rec % max_dom
   IF ( .NOT. model_config_rec % grid_allowed(i) ) CYCLE
   IF ( ( model_config_rec % e_we(i) /  model_config_rec % nproc_x .LT. 10 ) .OR. &
        ( model_config_rec % e_sn(i) /  model_config_rec % nproc_y .LT. 10 ) ) THEN
      WRITE ( wrf_err_message , * ) 'For domain ',i,', the domain size is too small for this many processors, ', &
                                    'or the decomposition aspect ratio is poor.'
```

ISSUE:
Closes issue #949 "domain decomposition check fails on an unused domain"

LIST OF MODIFIED FILES:

TESTS CONDUCTED:
1. Kevin confirms that this has the expected behavior for the initially identified failure case.
    a. With `grid_allowed` set to `.false.` for the unused domain, the problematic checks are skipped 
for that grid, and the model runs as expected (correct behavior).
   b. With the older version of the code, without the fix, we get fatal error alerting to a grid too small for the decomposition (incorrect behavior).
    c. With the new version of the code, with `grid_allowed = 6*.true.`, we get fatal errors alerting 
to a grid too small for the decomposition (correct behavior).
2. Set up namelist with inconsistent physics for 3 domain run, with domain 2 not being run. The 
second domain has the wrong PBL (typically each domain has the same PBL scheme selected). 
Also, domain 2 has an incompatible request for using Deng shallow cu (given the PBL scheme).
```
&domains
max_dom        = 3,
grid_allowed   = T, F, T
/
```
```
&physics
bl_pbl_physics =  2,     1,     2,
shcu_physics    = 5,     5,     5,
/
```
- [x] Original code fails when inconsistent physics schemes are selected, even though the domain is 
not active.
```
--- ERROR: Deng shallow convection can only work with MYJ or MYNN PBL
  --- ERROR: Fix shcu_physics or bl_pbl_physics in namelist.input.
--- ERROR: bl_pbl_physics must be equal for all domains (or = zero)
  --- Fix bl_pbl_physics in namelist.input
FATAL CALLED FROM FILE:  <stdin>  LINE:    1986
NOTE:       2 namelist settings are wrong. Please check and reset these options
```
 - [x] New code runs OK

### DIFF
--- a/arch/Config.pl
+++ b/arch/Config.pl
@@ -1,6 +1,6 @@
 #!/usr/bin/perl
 #
-# Configuration Script for WRF prototype code
+# Configuration script for WRF prototype code
 # 
 # Be sure to run as ./configure (to avoid getting a system configure command by mistake)
 #
--- a/arch/Config.pl
+++ b/arch/Config.pl
@@ -1,6 +1,6 @@
 #!/usr/bin/perl
 #
-# Configuration script for WRF prototype code
+# Configuration Script for WRF prototype code
 # 
 # Be sure to run as ./configure (to avoid getting a system configure command by mistake)
 #
--- a/share/module_check_a_mundo.F
+++ b/share/module_check_a_mundo.F
@@ -320,6 +320,7 @@
 ! each of those respective modules.
 !-----------------------------------------------------------------------
       DO i = 1, model_config_rec % max_dom
+         IF ( .NOT. model_config_rec % grid_allowed(i) ) CYCLE
          IF ( model_config_rec % sf_urban_physics(i) == bepscheme     ) THEN
            model_config_rec % num_urban_ndm  = bep_ndm()
            model_config_rec % num_urban_nz   = bep_nz_um()
@@ -341,6 +342,7 @@
 ! Check that mosiac option cannot turn on when sf_urban_physics = 2 and 3
 !-----------------------------------------------------------------------
       DO i = 1, model_config_rec % max_dom
+         IF ( .NOT. model_config_rec % grid_allowed(i) ) CYCLE
          IF ( model_config_rec % sf_surface_mosaic .EQ. 1 .AND. &
               (model_config_rec % sf_urban_physics(i) .EQ. 2 .OR. &
                model_config_rec % sf_urban_physics(i) .EQ. 3 ) ) THEN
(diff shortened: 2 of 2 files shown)


davegill:
@kwman44 
Kevin,
See if this fixes the problem that you noted.

kwman44:
@davegill 
I can confirm that this has the expected behavior for my failure case.
With grid_allowed=.false., the problematic checks are skipped for that grid, and the model runs as expected.
With grid_allowed=.true. we get the (expected) fatal error alerting me to a grid too small for the decomposition.

Tangentially related, would it be reasonable for real to use grid_allowed and not produce a wrfinput file where grid_allowed=.false.?

davegill:
@weiwangncar @dudhia @kkeene44 @smileMchen 
Folks,
Kevin said this fixes his problem, and the code works for my tests. This is ready for a review.

PR #952, opened 2019-07-04
Fix incorrect use of whereis and which to find nf-config command

yriveiro:
TYPE: bug fix

KEYWORDS: Linux, Darwin, nc-config, configure

SOURCE: Yago Riveiro (Air Quality and Odor Management - AQOM), internal

DESCRIPTION OF CHANGES: 
This pull fixes the incorrect use of `whereis` when we try to locate `nc-config` 
command. 
 - Only the `which` command is valid. 
 - The return code is used, not the value of the `which nc-config` command.
 - There is no need to determine the OS prior to this test.

LIST OF MODIFIED FILES: 
M       configure

TESTS CONDUCTED: 
 - [x] Re-run configure script, the warning disappears
 - [x] Still works on Darwin (netcdf/3.6.3 and netcdf/4.5.0)
 - [x] Still works on Linux desktop Centos 7.6 (netcdf/4.7.0)
 - [x] Still works on Linux NCAR supercomputer SUSE 12 SP4 (netcdf/4.6.3)

### DIFF
--- a/configure
+++ b/configure
@@ -205,9 +205,9 @@ fi
 # If the user asked for classic netcdf, acquiesce to the request.
 
 if [ "`uname`" = "Linux" ] ; then
-  ans="`whereis nf-config`"
-elif [ "`uname`" = "Darwin" ] ; then
   ans="`which nf-config`"
+elif [ "`uname`" = "Darwin" ] ; then
+  ans="`whereis nf-config`"
 else
   ans=""
 # echo "Add in your architecture's uname and the command to find executables in the path"
--- a/configure
+++ b/configure
@@ -204,16 +204,9 @@ fi
 
 # If the user asked for classic netcdf, acquiesce to the request.
 
-if [ "`uname`" = "Linux" ] ; then
-  ans="`which nf-config`"
-elif [ "`uname`" = "Darwin" ] ; then
-  ans="`which nf-config`"
-else
-  ans=""
-# echo "Add in your architecture's uname and the command to find executables in the path"
-# exit 1
-fi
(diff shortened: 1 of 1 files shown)


davegill:
@yriveiro 
Yago,
Would you please include your affiliation. That way we can give you proper credit when release 4.1.3 comes out.

yriveiro:
@davegill

Affiliation included.

davegill:
@yriveiro 
Yago,
Take a look at:
```
https://superuser.com/questions/40301/which-whereis-differences
```
I am thinking that we should always and only use `which`. The `whereis` command could end up giving location information on man pages, etc.

davegill:
@yriveiro 
Yago,
Here is a quick output I generated:
```
> which nf-config
nf-config: Command not found.
> set OK = $status
> echo $OK
1
> which ls
/bin/ls
> set OK = $status
> echo $OK
0
```
We are going to ONLY use `which`, but we are going to check the return code of the command.

yriveiro:
Consistency is good for code, allows code to be simple to read and maintain.

This approach seems to be better.

davegill:
@yriveiro @weiwangncar 
Yago,
Would you please verify that this latest modification fixes the build troubles you had.

yriveiro:
lgtm

davegill:
@kkeene44 
Kelly,
Would you please review.
1. Positive tests:
   * The original poster says that this fixes his troubles.
2. Negative tests:
   * I have conducted several tests on different architecture / OS combinations which used to previously work, and still do. There are no indications of breaking any builds.

PR #958, opened 2019-07-18
New DA namelist alpha_vertloc_opt to choose how to get the alpha_vertloc info

jamiebresch:
TYPE: enhancement

KEYWORDS: WRFDA, alpha_vertloc, alpha_vertloc_opt

SOURCE: Jamie Bresch (NCAR/MMM)

DESCRIPTION OF CHANGES:

PR #850 discussed about the issues in current var/gen_be/gen_be_vertloc.f90.
Instead of modifying the stand-alone var/gen_be/gen_be_vertloc.f90 utility that generates be.vertloc.dat, this PR proposes a simpler and better solution inside WRFDA.

1. A new namelist alpha_vertloc_opt is added to replace alpha_vertloc to choose how to handle
vertical localization for EnVar applications.
alpha_vertloc_opt=1, the behavior is the same as before, except that the new code now gives proper error message when be.vertloc.dat does not exist.
alpha_vertloc_opt=2 (default, recommended), let WRFDA calculate logP-based vertical localization. be.vertloc.dat will be written out.
alpha_vertloc_opt = 0, no vertical localization.
alpha_vertloc_opt  is only used when ensdim_alpha>0, so the alpha_vertloc_opt setting does not matter when ensdim_alpha=0.
2. frame/module_configure.F is modified to notify users about the obsolete alpha_vertloc option.
3. var/README.namelist is updated.

LIST OF MODIFIED FILES:
M       Registry/registry.var
M       frame/module_configure.F
M       var/README.namelist
M       var/build/depend.txt
A       var/da/da_setup_structures/da_get_alpha_vertloc.inc
M       var/da/da_setup_structures/da_setup_be_nmm_regional.inc
M       var/da/da_setup_structures/da_setup_be_regional.inc

TESTS CONDUCTED:
A case to test the new code works as expected.

RELEASE NOTE: For EnVar DA applications, logical alpha_vertloc namelist switch is replaced by new integer option alpha_vertloc_opt. See var/README.namelist.


jjguerrette:
I am thinking that only `alpha_vertloc_opt` is needed and not `alpha_vertloc`.  @jamiebresch, do you think the logical on/off will help users very much?

jamiebresch:
I did not touch alpha_vertloc in order to keep the change to a minimum.
The only benefit is that if a non-EnVar user has alpha_vertloc in the namelist, the new code still works.
alpha_vertloc can certainly be removed. I have no preference.

jjguerrette:
I think alpha_vertloc can be removed.  If alpha_vertloc_opt is set to a non-zero value for a non-EnVar application, WRFDA should quit with an informative error message.

jjguerrette:
@jamiebresch, sorry for the delay on this one.  The previous default was to have vertical localization turned off.  I can see why that was not ideal.  Should the release note state that the new default is to turn on the localization?  Other than that, this looks ready to me.

jjguerrette:
@liujake, note that the following REGTEST namelist entries need to be removed or changed to `alpha_vertloc_opt=2`:

`namelist.input.4denvar_large_conus: alpha_vertloc                       = .true.,`
`namelist.input.dual_res_hybrid: alpha_vertloc                       = .true.,`
`namelist.input.realtime_hybrid: alpha_vertloc=.true.`

The reference output will change as well.

PR #960, opened 2019-07-23
YSU PBL: check on flag_qi instead of present(rqiblten)

dudhia:
TYPE: bug fix

KEYWORDS: YSU, WSM3, Kessler, ice check

SOURCE: internal

DESCRIPTION OF CHANGES: 
YSU checks on PRESENT(RQIBLTEN) to determine whether to fill local ice part of qv2d, but 
RQIBLTEN is always present, so this does not avoid copy from ice array when WSM3/Kessler MP 
is used. Better to check flag_qi which is also available and true except for WSM3 and Kessler.

LIST OF MODIFIED FILES: 
phys/module_bl_ysu.F

TESTS CONDUCTED: 
 - [x] Check code runs and uses flag_qi for simple MP scheme.

RELEASE NOTE: 
Using the YSU PBL scheme in combination with either the WSM3 or Kessler MP schemes (no ice) is now fixed to not use ice array.

### DIFF
--- a/phys/module_bl_ysu.F
+++ b/phys/module_bl_ysu.F
@@ -229,7 +229,7 @@ contains
         do i = its,ite
           qv2d(i,k) = qv3d(i,k,j)
           qv2d(i,k+kte) = qc3d(i,k,j)
-          if(present(rqiblten)) qv2d(i,k+kte+kte) = qi3d(i,k,j)
+          if(flag_qi) qv2d(i,k+kte+kte) = qi3d(i,k,j)
         enddo
       enddo
 !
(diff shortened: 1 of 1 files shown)


davegill:
@lfowler58 @dudhia @weiwangncar 
Jimy,
Laura is working on a newer YSU scheme as part of CPF (so eventually to be shared in MPAS and WRF). Her modifications include a cleaned up way to handle the number of moist arrays (using input flag variables, as you are doing). May we hold off with this PR until her shared code is available for consideration?

dudhia:
This is for 4.1.3 as that release is likely to go out before we do any
merging,
so it can be seen as a temporary fix.
Jimy

weiwangncar:
I'm ok with the change.

davegill:
@dudhia 

> This is for 4.1.3 as that release is likely to go out before we do any
merging,
so it can be seen as a temporary fix.

Jimy,
I'm OK with that plan

davegill:
@weiwangncar 
Wei,
Looking over these proposed changes, I am OK with them. You probably need to explicitly approve them.

PR #961, opened 2019-07-23
Revert "YSU PBL: check on flag_qi instead of present(rqiblten)"

dudhia:
Reverts wrf-model/WRF#960


dudhia:
need to do it again because the message was wrong. Need approval.

dudhia:
I made the pull request again

PR #962, opened 2019-07-23
YSU fix for WSM3 and Kessler no-ice test

dudhia:
TYPE: bug fix

KEYWORDS: YSU, WSM3, Kessler, ice check

SOURCE: internal

DESCRIPTION OF CHANGES:
YSU checks on PRESENT(RQIBLTEN) to determine whether to fill local ice part of qv2d, but
RQIBLTEN is always present, so this does not avoid copy from ice array when WSM3/Kessler MP
is used. It is better to check flag_qi which is also available and true (except when using either
WSM3 and Kessler MP schemes).

LIST OF MODIFIED FILES:
phys/module_bl_ysu.F

TESTS CONDUCTED:
Check code runs and uses flag_qi for simple MP scheme.

RELEASE NOTE:
Using the YSU PBL scheme in combination with either the WSM3 or Kessler MP schemes (no ice) is now fixed to not use ice array.

### DIFF
--- a/phys/module_bl_ysu.F
+++ b/phys/module_bl_ysu.F
@@ -229,7 +229,7 @@ contains
         do i = its,ite
           qv2d(i,k) = qv3d(i,k,j)
           qv2d(i,k+kte) = qc3d(i,k,j)
-          if(present(rqiblten)) qv2d(i,k+kte+kte) = qi3d(i,k,j)
+          if(flag_qi) qv2d(i,k+kte+kte) = qi3d(i,k,j)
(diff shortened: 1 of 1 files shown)


dudhia:
need approval again

dudhia:
I am undoing reversion instead - sorry for confusion
please approve reverting reversion

dudhia:
this just needs approval again

davegill:
@kkeene44 
I am good with this (no permission to be "official").
You can approve.

PR #969, opened 2019-08-02
MYNN PBL (icloud) restart  bug fix 

kkeene44:
TYPE: bug fix

KEYWORDS: bug, MYNN, PBL, cldfra_bl, icloud_bl, Registry

SOURCE: internal

DESCRIPTION OF CHANGES: When using the MYNN PBL scheme, restarts did not give bit-for-bit results. When using this scheme, the default setting for icloud_bl is 1 (on), but the variable used (cldfra_bl) was not put in the restart stream. Simply adding this variable to the restart stream in Registry.EM_COMMON corrects the problem.

LIST OF MODIFIED FILES: 

TESTS CONDUCTED: Confirmed that the code builds, and the change corrects the non-bit-for-bit restart problem.

RELEASE NOTE: When using the MYNN PBL scheme, with icloud_bl = 1 (which is default), restarts did not give bit-for-bit results when compared to a non-restart run. This problem has been corrected.


dudhia:
Looked OK. Why was it closed?

dudhia:
OK, I see now

PR #970, opened 2019-08-02
bug fix for mynn PBL restarts - not giving bit-for-bit results

kkeene44:
TYPE: bug fix

KEYWORDS: bug, MYNN, PBL, cldfra_bl, icloud_bl, Registry

SOURCE: internal

DESCRIPTION OF CHANGES: When using the MYNN PBL scheme, restarts did not give bit-for-bit 
results. When using this scheme, the default setting for icloud_bl is 1 (on), but the variable used 
(cldfra_bl) was not put in the restart stream. Simply adding this variable to the restart stream in 
Registry.EM_COMMON corrects the problem.

LIST OF MODIFIED FILES:
M Registry/Registry.EM_COMMON

TESTS CONDUCTED: 
Confirmed that the code builds, and the change corrects the non-bit-for-bit restart problem.

RELEASE NOTE: When using the MYNN PBL scheme, with icloud_bl = 1 (which is default), restarts did not give bit-for-bit results when compared to a non-restart run. This has been a problem since the option was introduced in V3.8, but is now corrected.


davegill:
@kkeene44
The line you fixed has been there since this option was first put in, Jan 2016. Would you add that info to the release message.

kkeene44:
@davegill I updated the release note.

dudhia:
Looks fine to me.

weiwangncar:
I'm ok with this too.

PR #977, opened 2019-08-23
max_ts_level should be .LE. the number of half levels

### LINKED ISSUE #976: Need check_a_mundo test: max_ts_level must be <= number of znu half layers
From Pedro Jimenez

I found that max_ts_level should not be set to a number larger than the number of half levels. If the user accidentally set this namelist entry to a larger value, then the wrf_timeseries.F tries to use w_2  and other vars at higher levels than possible. The model would not crash but I do not know what would happen to the time series results. I found this compiling WRF in debugging mode since I set up max_ts_level to the number of full levels. Unfortunately I removed the log file with the error.

There should be a check maybe in check_a_mundo to ensure that max_ts_level is not set up to a number higher than the number of half levels.

smileMchen:
TYPE: bug fix

KEYWORDS: max_ts_level

SOURCE: Pedro Jimenez Munoz (RAL/MMM), internal

DESCRIPTION OF CHANGES: 
The value of `max_ts_level` should be set less than or equal to the number of half levels. If 
`max_ts_level` is accidentally set to a larger value, the wrf_timeseries.F would try to access 
the vertical velocity (`w_2`) and other variables that are defined on full levels with an index 
outside of the variables' extent.  A check is added to check_a_mundo to ensure that if 
`max_ts_level` is  set up to a number larger than the number of half layers, then it is reset to 
the number of half layers.

ISSUE:
Fixes #976 "Need check_a_mundo test: max_ts_level must be <= number of znu half layers"

LIST OF MODIFIED FILES:
M    run/README.tslist 

TESTS CONDUCTED: 
1. A single test is run to ensure the check works as expected. Below is the output:

```
Quilting with   1 groups of   0 I/O tasks.
 Ntasks in X            6 , ntasks in Y            6
   max_ts_level must be <= number of znu half layers
   max_ts_level is reset to the number of znu half layers
```

### DIFF
--- a/share/module_check_a_mundo.F
+++ b/share/module_check_a_mundo.F
@@ -1906,13 +1906,13 @@
 !-----------------------------------------------------------------------
 !   Check that max_ts_level should be smaller than the number of half levels
 !-----------------------------------------------------------------------
-   IF ( model_config_rec % max_ts_level .gt. model_config_rec %e_vert(1)-1 )  then
+      IF ( model_config_rec % max_ts_level .gt. model_config_rec %e_vert(1)-1 )  then
         wrf_err_message = ' max_ts_level must be <= number of znu half layers '
         CALL wrf_debug ( 0, TRIM( wrf_err_message ) )
         wrf_err_message = ' max_ts_level is reset to the number of znu half layers '
         CALL wrf_debug ( 0, TRIM( wrf_err_message ) )
         model_config_rec % max_ts_level = model_config_rec %e_vert(1)-1
-   ENDIF
+      ENDIF
 
 !-----------------------------------------------------------------------
 !  Consistency checks between vertical refinement and radiation
--- a/share/module_check_a_mundo.F
+++ b/share/module_check_a_mundo.F
@@ -1904,7 +1904,7 @@
       ENDDO
 
 !-----------------------------------------------------------------------
-!   Check that max_ts_level should be smaller than the number of half levels
+!   Check that max_ts_level is smaller than the number of half levels
 !-----------------------------------------------------------------------
       IF ( model_config_rec % max_ts_level .gt. model_config_rec %e_vert(1)-1 )  then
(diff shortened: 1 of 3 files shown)


jamiebresch:
Should max_ts_locs be max_ts_level?

dudhia:
oops,yes

share/module_check_a_mundo.F
```diff
@@ -1903,6 +1903,15 @@ END FUNCTION bep_bem_nbui_max
         ENDIF
       ENDDO
 
+!-----------------------------------------------------------------------
+!   Check that max_ts_level should be smaller than the number of half levels
+!-----------------------------------------------------------------------
+   IF ( model_config_rec % max_ts_locs .gt. model_config_rec %e_vert(1)-1 )  then
```
davegill: @smileMchen 
This should be max_ts_level, right?

davegill:
@jamiebresch 
🦅 👁

share/module_check_a_mundo.F
smileMchen: @davegill 
Yes it should be max_ts_level. I will recommit the code.

jamiebresch:
How about also updating max_ts_level description in run/README.tslist to clarify that it is number of model half levels?

smileMchen:
@jamiebresch 
Jamie,
I prefer not to change it. In this way, one can determine how many levels to be  included in the profile. For example, sometimes one may only need to look at profiles in the PBL.

davegill:
@smileMchen 
> How about also updating max_ts_level description in run/README.tslist to clarify that it is number of model half levels?

Ming,
Here is the line in the README.tslist file that refers to the number of vertical levels:
```
 * max_ts_level = number of model levels for time series vertical
   profiles (default is 15)
```
We should add extra info for the users:
```
 * max_ts_level = number of model levels for time series vertical
   profiles, the default is 15, the maximum number of vertical levels
   is e_vert-1 (the number of half layers in the model run)
```

PR #978, opened 2019-08-28
Feature/sfclay z0nested fix

domingom:
This code commit fixes a WRF issue affecting some surface layer variables on a parent domain at the time of initializing the nested domain. The attached figure demonstrates the problem and provides verification of the fix.

initialization_d02_effect_on_d01_fix


dudhia:
OK, but if there is a fix for nested diagnostics we would be interested.

davegill:
Accidental PR to wrong repo

PR #983, opened 2019-09-10
Print separate numerator and denominator in lightning driver

zxdawn:
TYPE: bug fix
KEYWORDS: lightning driver

SOURCE: J. Wong, NCAR

DESCRIPTION OF CHANGES: 
Correct a print statement for the IC:CG ratio in lightning driver. Originally, this was printed as
a separate numerator and denominator.

LIST OF MODIFIED FILES:
phys/module_lightning_driver.F

TESTS CONDUCTED:
The message prints correctly:
```
lightning_driver: using user-prescribed IC:CG ratio = xx
```


davegill:
@zxdawn 
Xin,
Since this is a really a bug, and SO small, please make this PR base branch the release-v4.1.3. This does not need to be on develop, where users will not see this for another 5-6 months.

zxdawn:
@davegill Should I just change the `base` to `release-v4.1.3`?
Because I see this warning:
```
Some commits from the old base branch may be removed from the timeline, and old review comments may become outdated.
```

zxdawn:
@davegill BTW, the modification didn't *Print separate numerator and denominator in lightning driver*. Instead, I just keep the message of printing the correct value of IC:CG ratio.

davegill:
@zxdawn 
Xin,
1. This is such a small commit that you can go ahead and try to just change the base. If that does not work (for example, afterwards it says you are changing 237 files, for example), then just close this and do another PR from the release-v.4.1.3 base branch.
2. Thanks for pointing out the oops in the title modification!
3. Please build the code and run the code, then copy the actual print out and place that in the PR commit message.

zxdawn:
@davegill Thanks! I will create the new PR after building the code.

PR #985, opened 2019-09-17
Fix conflict between obs nudging and Thompson microphysics

brianreen:
TYPE: bug fix

KEYWORDS: observation nudging, Thompson microphysics

SOURCE: Brian Reen (Army Research Lab)

DESCRIPTION OF CHANGES: 
When observation nudging was used on domain 3, Thompson microphysics was used, and the .dat files used by Thompson microphysics were already present WRF would crash.

The problem originates from the same unit (63) being used for the Thompson microphysics .dat files (qr_acr_qg.dat, qr_acr_qs.dat, and freezeH2O.dat) and the domain 3 observation nudging input file (OBS_DOMAIN301).  The Thompson microphysics code does not explicitly close qr_acr_qg.dat and freezeH2O.dat, and when observation nudging attempts to read from unit 63, it is freezeH2O.dat that is open rather than OBS_DOMAIN301.

This problem was earlier found and corrected in 80b22b316df5a24b13bf42e1d686d0b11e18dca0 but was reintroduced in 6cea5a0fd74af2abdd78551e68ca9f75ffd0d8dc.  The code earlier included to correct this problem by explicitly closing unit 63 for qr_acr_qg.dat and freezeH2O.dat was been reintroduced.

LIST OF MODIFIED FILES: 
M phys/module_mp_thompson.F

TESTS CONDUCTED: 
A 3-domain simulation with obs nudging and the *.dat files pre-existing using the unmodified code leads to the following messages in rsl.out:

<pre>
OBS NUDGING: Reading new obs for time window TBACK =   -1.500 TFORWD =    1.500 for grid =  3
 File open on obs nudging unit (          63 ) with wrong name
 File open on obs nudging unit is named freezeH2O.dat but it should be named OBS_DOMAIN301
 This likely means this unit number was opened elsewhere in WRF
FATAL CALLED FROM FILE:  wrf_fddaobs_in.b  LINE:     460
wrf_fddaobs_in: in4dob STOP Obs nudging file name mismatch
</pre>

A 3-domain simulation with obs nudging and the *.dat files pre-existing using the modified code finishes without error.  This simulation creates bit-identical wrfout files to an experiment using the unmodified code in which the *.dat files do not exist. 

Tested 3-domain simulation without obs nudging with and without this modification and found bit-identical results.

### DIFF
--- a/phys/module_mp_thompson.F
+++ b/phys/module_mp_thompson.F
@@ -3611,7 +3611,7 @@
               ENDIF
             ENDIF
           ELSE
-            INQUIRE(63,opened=lopen)
+            INQUIRE(63,OPENED=lopen)
             IF (lopen) THEN
               CLOSE(63)
             ENDIF
--- a/phys/module_mp_thompson.F
+++ b/phys/module_mp_thompson.F
@@ -3610,6 +3610,11 @@
                 CALL wrf_error_fatal("Error opening qr_acr_qg.dat. Aborting because force_read_thompson is .true.")
               ENDIF
             ENDIF
+          ELSE
+            INQUIRE(63,opened=lopen)
+            IF (lopen) THEN
+              CLOSE(63)
+            ENDIF
           ENDIF
         ELSE
           IF( force_read_thompson ) THEN
@@ -4082,6 +4087,11 @@
               IF( force_read_thompson ) THEN
                 CALL wrf_error_fatal("Error opening freezeH2O.dat. Aborting because force_read_thompson is .true.")
               ENDIF
+            ENDIF
+          ELSE
+            INQUIRE(63,opened=lopen)
+            IF (lopen) THEN
+              CLOSE(63)
             ENDIF
           ENDIF
         ELSE


davegill:
@gthompsnWRF 
Greg,
Please take a look at this PR

weiwangncar:
@brianreen I thought something like this has been fixed before...

brianreen:
@weiwangncar  Yes, it was corrected in 80b22b3 but was reintroduced in 6cea5a0 when code fixing the problem was removed.

weiwangncar:
@brianreen I'm sorry about it... And thanks for fixing it again...

gthompsnWRF:
> @gthompsnWRF
> Greg,
> Please take a look at this PR

I assume the proper fix is just adding a CLOSE(63) just prior (or after) the good=1 line?
And it appears that Subroutine table_ccnAct needs to CLOSE the open file also, which it isn't.

I was always slightly concerned when John M. put in this chunk of code and hardwired file unit numbers rather than searching for an empty slot first.

How do others want to handle the fact that a CLOSE is also needed in the other subroutine? Separate PR or just include it with this one?

brianreen:
@gthompsnWRF 
It does appear that adding a CLOSE(63) just prior (or after) the good=1 line should produce identical results to the changes I made.  The changes I made were just to reinsert deleted code to restore this section of the code to the previous state that worked.  When I originally made the changes I think I probably did it this way because the already-extant close statements explicitly checked whether the unit was open and were within the IF statement regarding the status of the variable good.  Thus, I was trying to put the close statement where the other close statements were and to follow the pattern of the code already in place to close unit 63.  However, in the case where good=1 it seems that the file must be open and so checking if it is open seems extraneous.  If adding the CLOSE(63) statement at the good=1 line is the preferred solution I can do that for the two reversions I made in this pull request (for qr_acr_qg.dat, and freezeH2O.dat), but for consistency should probably also make the equivalent change for qr_acr_qs.dat.  The equivalent CLOSE statement for that file was not removed and so I did not need to revert it.  Is this the preferred solution?

File unit numbers are a bit tricky, since it is unclear how someone can be sure to avoid stepping on a unit number used by another part of WRF.  Since it seems that obs nudging would not have opened the obs nudging files by the time microphysics is first accessed, even if one looked for available unit numbers there is no guarantee that one later used by obs nudging would not be chosen as available and used.  Searching the code to find unit numbers hardcoded elsewhere may not work since for obs nudging the unit number is 60 + the current domain number, and so one would not find an explicit reference to unit 63.

davegill:
@gthompsnWRF @brianreen @weiwangncar @dudhia 
Greg and Brian,
We are trying to get these mods in. What is the final decision?

brianreen:
@gthompsnWRF
Greg,
Since the code currently in this pull request fixes the identified problem and does so by reverting parts of the code to earlier versions that worked, and given the desire to get the bug fix finalized, perhaps use the code in the pull request for this release and consider the further modifications you suggested for a future release?

davegill:
@brianreen @gthompsnWRF @weiwangncar @dudhia 
> perhaps use the code in the pull request for this release and consider the further modifications you suggested for a future release

Folks,
This is the direction that I am leaning

PR #991, opened 2019-09-26
Release v4.1.3 nssl mp update

MicroTed:
TYPE: bug fix

KEYWORDS: snow reflectivity; ice effective radius

SOURCE: Ted Mansell (NSSL microphysics developer/maintainer)

DESCRIPTION OF CHANGES: 

 Bug fixes:
  - Effective radius calculations were only done at history times. Now is done every time step. [impact only when used with RRTMG radiation]
  - Snow reflectivity: Previous "fix" was incorrect and yields snow dBZ that is too low. Reverted to old version which was correct
  - Incorrectly updated a state value in the reflectivity code. (Could cause small differences if reflectivity is not calculated)

 Updates:
  - Put limit on snow volume (2 cm diameter) in aggregation rate [enhancement]
  - Graupel and hail drag coefficients are returned from fall speed subroutine to use in ventilation coeffs. for consistency (minor change) [enhancement]
  - Added code hints to use the "axtra2d" array to communicate rates from the microphysics routine into any 3d arrays that are passed in to the driver. [no impact]
  - Added (compile) option flag icracr to turn off rain self-collection [no impact: user must change code]
  - Added compile options 'depfac' and 'meltfac' to adjust deposition/sublimation and melting (not freezing) rates of graupel/hail by a constant factor (for experimentation). Default values are 1.0 [no impact: user must change code]

LIST OF MODIFIED FILES: 
phys/module_mp_nssl_2mom.F

TESTS CONDUCTED: Tested ideal cases to be sure nothing got broken.

RELEASE NOTE: Snow reflectivities were too low and were reverted to a previous version. Effective radius calculations for droplets/ice (used only by RRTMG) were only calculated at history output times and are now done every time step.

### DIFF
--- a/phys/module_mp_nssl_2mom.F
+++ b/phys/module_mp_nssl_2mom.F
@@ -71,7 +71,7 @@
 !---------------------------------------------------------------------
 ! Sept. 2019:
 ! Bug fixes:
-!  - effective radius calculation was only done at history times. Now every time step (though should be just before radiation is called)
+!  - Effective radius calculation was only done at history times. Now every time step (though should be just before radiation is called)
 !  - Snow reflectivity: Previous "fix" was incorrect and yields snow dBZ that is too low. Reverted to old version which was correct
 !  - Incorrectly updated a state value in the reflectivity code. (Could cause small differences if reflectivity is not calculated)
 ! Updates:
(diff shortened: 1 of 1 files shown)


davegill:
@weiwangncar @dudhia 
Would you please provide a review

weiwangncar:
@MicroTed I'm ok with the changes. In the future, it might be good to separate bug fixes and enhancement for easy documentation.
@davegill We should do a regtest that test this scheme.

MicroTed:
@weiwangncar That's a good suggestion about separating the bug fixes. I'll try to remember that next time, especially if it is more than a minor issue.
It might be nice to introduce a flag or counter to know when the RRTMG is about to be called, so that the effective diameters only get calculated when needed. But it's not a very expensive calculation, either.

PR #972, opened 2019-08-13
Add a scale-adaptive 3DTKE parameterization scheme as a subgrid mixing option 

zhangx-3dtke:
TYPE: new feature

KEYWORDS: 3DTKE scheme, scale-adaptive

SOURCE: Xu Zhang (Shanghai Typhoon Institute/CMA, China)

DESCRIPTION OF CHANGES: 
This PR adds a new scale-adaptive 3DTKE scheme (SMS-3DTKE, Shanghai Meteorological Service) 
for subgrid turbulent mixing. The new scheme is self-adaptive to the grid-size between the 
large-eddy simulation (LES) and mesoscale limits. The new 3DTKE scheme is added as a new 
diffusion option km_opt = 5. This option extends original LES subgrid model 1.5-order TKE closure 
(km_opt=2) to the mesoscale, and can be used in LES, mesoscale, and the gray zone resolutions in 
between. In the horizontal diffusion, the new scheme blends horizontal Smagorinsky first-order 
closure (km_opt=4) and 1.5-order TKE-based horizontal diffusivity. In the mesoscale limit, 
horizontal diffusion recovers to the 2D Smagorinsky form (km_opt=4), and the new vertical 
diffusion adds a non-local term and is made implicit to allow for longer time steps and thin levels. 
The new scheme can therefore replace km_opt=2 and conventional PBL schemes. In the LES limit, 
km_opt=5 recovers km_opt=2. The added 3D arrays are packaged in Registry.

Several newly added subroutines:
1. SUBROUTINE nonlocal_flux: prescribes the nonlocal heat flux profile based on LES analysis 
and computes the nonlocal momentum gamma term.
2. SUBROUTINE free_atmos_length: calculates mixing length in the free atmosphere.
3. SUBROUTINE meso_length_scale: calculates mesoscale length scale.
4. SUBROUTINE vertical_diffusion_implicit: solves the vertical diffusion equations for u, v, w, th, 
tke, moist, chem, scalar, tracer using an implicit method.
5. SUBROUTINE update_tke_implicit: solves the TKE equation using implicit method. The vertical 
TKE diffusion and dissipation are implicitly treated.

From a user perspective, when the new scheme is turned on (diff_opt=2 and km_opt=5), PBL 
schemes must be turned off (bl_pbl_physics = 0).

This scheme can be used with sf_sfclay_physics = 1, 5, 91.

LIST OF MODIFIED FILES: 
Registry/Registry.EM_COMMON
dyn_em/module_diffusion_em.F
[...]

PR #1012, opened 2019-11-15
SMS3dTKE: Remove broken packaging for km_opt=5

davegill:
TYPE: bug fix

KEYWORDS: packaging, sms3dtke

SOURCE: Internal

DESCRIPTION OF CHANGES:

The existing packaging in combination with the logic added in the
code for PR #972 (Add a scale-adaptive 3DTKE parameterization scheme as a subgrid 
mixing option) d059afe1e646f7 causes segmentation faults for most model
simulations that use km_opt=2.

The packaging is important, but we have a broken repository. The
packaging will be removed, developers will fix the problem, and
then put the packaging back into the Registry file for this option.

Jan Mandel states:
> "found that l_scale, th_h_tend, and tke_diffusion_h_tend made my km_opt=2 test case fail, 
and with them removed from the package it did not fail. But I do not know if sms_3dtke even 
with the packaging removed does not break something in a different way."

Perhaps these are the variables to start the seg fault investigation to re-introduce 
packaging for this option.

LIST OF MODIFIED FILES:

TESTS CONDUCTED:
 - [x] With original packaging, all tests with km_opt=2 failed
 - [x] With mods (removing km_opt=5 packaging), km_opt=2 and km_opt=5 tests pass

### DIFF
--- a/Registry/Registry.EM_COMMON
+++ b/Registry/Registry.EM_COMMON
@@ -2915,7 +2915,8 @@ package   mynn_dmp_edmf  bl_mynn_edmf==1            -              state:edmf_a,
 package   pbl_cloud      icloud_bl==1                -             state:cldfra_bl,qc_bl
 
 #SMS-3DTKE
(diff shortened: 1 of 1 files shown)


weiwangncar:
@davegill This looks ok to me.

davegill:
Kelly,
Would you please review this PR. You have "super powers", so you can be an infrastructure person. Wei and Jimy both are OK with this.

PR #1013, opened 2019-11-15
Add missing "IF km_opt==5" tests for diag computations

davegill:
TYPE: bug fix

KEYWORDS: sms3dtke, km_opt==5

SOURCE: Internal

DESCRIPTION OF CHANGES:
Each location where the optional diagnostics for km_opt==5 are computed,
instead of only using a local IF test, now a combination of the same local
IF test (T/F do the computation) is used in conjunction with and a test
for if the diagnostic is permitted (is km_opt==5).

The km_opt==5 scheme originally had (and hopefully, will eventually again have) 
specified arrays via the package capability in the Registry. Those fields need to be 
protected from accidental use when km_opt==2.

LIST OF MODIFIED FILES:
modified:   dyn_em/module_diffusion_em.F

TESTS CONDUCTED:
 - [x] Without mods, there are attempts to compute km_opt==5 diags when km_opt==2.
 - [x] With mods, no attempts to compute km_opt==5 diags when km_opt==2.

### DIFF
--- a/dyn_em/module_diffusion_em.F
+++ b/dyn_em/module_diffusion_em.F
@@ -3344,7 +3344,7 @@ SUBROUTINE horizontal_diffusion_u_2( tendency, config_flags,              &
    ENDDO
 
 ! XZ
-   IF (output_tend) THEN
+   IF ( ( output_tend ) .AND. ( config_flags%km_opt .EQ. 5 ) ) THEN
       DO j = j_start, j_end
       DO k = kts,ktf
       DO i = i_start, i_end
@@ -3564,7 +3564,7 @@ SUBROUTINE horizontal_diffusion_v_2( tendency, config_flags,              &
    ENDDO
 
 ! XZ
-   IF (output_tend) THEN
+   IF ( ( output_tend ) .AND. ( config_flags%km_opt .EQ. 5 ) ) THEN
       DO j = j_start, j_end
       DO k = kts,ktf
       DO i = i_start, i_end
@@ -3783,7 +3783,7 @@ SUBROUTINE horizontal_diffusion_w_2( tendency, config_flags,              &
    ENDDO
 
 ! XZ
-   IF (output_tend) THEN
+   IF ( ( output_tend ) .AND. ( config_flags%km_opt .EQ. 5 ) ) THEN
       DO j = j_start, j_end
       DO k = kts+1,ktf
       DO i = i_start, i_end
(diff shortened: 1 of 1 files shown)


dudhia:
Were arrays only allocated for km_opt=5 via packages used? If so,
that could be added to the description. Or else what was the nature of the
failure?

davegill:
@dudhia @smileMchen @weiwangncar @janmandel
Jimy,
With the top of the repository on the develop branch, every ideal ideal job will fail that uses km_opt=2. I started removing individual variables in the Registry.EM_COMMON file package for sms3dtke, to see if I could find "the one" problematic variable. When that was not successful, I tried looking through the code to see if there were obvious missing IF tests. Both of these activities became too much of a time drain. 

The only solution for a working develop branch is to unpackage everything that is new for the sms3dtke option (more than a dozen new arrays) and keep that memory and those fields allocated. The original developer needs to address this problem.

dudhia:
OK, so it possibly means that there are areas in the code that have not
been found yet
that use these packaged arrays when the scheme is off. Thanks for taking
the time on this.

weiwangncar:
@davegill This makes sense too.

dudhia:
Does this fix the previous issue?
    On Thursday, November 14, 2019, 07:38:17 PM MST, Dave Gill <notifications@github.com> wrote:  
 
 
TYPE: bug fix

KEYWORDS: sms3dtke, km_opt==5

SOURCE: Internal

DESCRIPTION OF CHANGES:
Each location where the optional diagnostics for km_opt==5 are computed,
instead of only using a local IF test, now a combination of the same local
IF test (T/F do the computation) is used in conjunction with and a test
for if the diagnostic is permitted (is km_opt==5).

LIST OF MODIFIED FILES:
modified: dyn_em/module_diffusion_em.F

TESTS CONDUCTED:
   
   -  Without mods, there are attempts to compute km_opt==5 diags when km_opt==2.
   -  With mods, no attempts to compute km_opt==5 diags when km_opt==2.

You can view, comment on, or merge this pull request online at:

  https://github.com/wrf-model/WRF/pull/1013

Commit Summary
   
   - Add missing "IF km_opt==5" tests for diag computations

File Changes
   
   -  M dyn_em/module_diffusion_em.F (12) 

Patch Links:
   
   - https://github.com/wrf-model/WRF/pull/1013.patch
   - https://github.com/wrf-model/WRF/pull/1013.diff

PR #1009, opened 2019-11-10
Fixing sms_3dtke km_opt=5 which breaks km_opt=2

janmandel:
TYPE: bug fix

KEYWORDS: 3DTKE, Fire

SOURCE: Jan Mandel

DESCRIPTION OF CHANGES:

Removed variables  l_scale, th_h_tend, and tke_diffusion_h_tend included in registry package sms_3dtke for km_opt=5 which are used outside of km_opt=5 but should not be and broke km_opt=2. 

When running with bound checks  on, I also found and fixed an index overrun in vertical staggering I originally made in https://github.com/openwfm/WRF-Fire-merge/commit/be6d402373151a5f66d7757bc6321a984c63fc99  The memory allocated in the vertical grid direction is by one one less than what I thought. The fix is, do not allow the index exceed the end memory index.

LIST OF MODIFIED FILES: 
Registry/Registry.EM_COMMON | 7 ++++++-
 phys/module_fr_fire_util.F  | 6 +++---

TESTS CONDUCTED: Jenkins, test/em_fire/two_fires serial (build option 13) only 

RELEASE NOTE: 

Symptoms: /compile em_fire; cd test/em_fire/two_fires; ./ideal.exe; ./wrf.exe failed with error message FIRE:crash: NaN detected

Tested with 
-fpe0 -check noarg_temp_created,bounds,format,output_conversion,pointers,uninit -ftrapuv -unroll0 -u
uncommened in configure.wrf

It is possible that more variables need to be removed from sms_3dtke for other values of km_opt.

PR #1014, opened 2019-11-15
Fire: fix vert index for half-levels

davegill:
TYPE: bug fix

KEYWORDS: fire, bounds check, vertical index

SOURCE: Internal

DESCRIPTION OF CHANGES:
A bounds check indicated a half-level variable was accessed with
the vertical index at the top full-level. Only a single array is
in the DO loop. The vertical index is reduced by one.

LIST OF MODIFIED FILES:
modified:   phys/module_fr_fire_util.F

TESTS CONDUCTED:
 - [x] Without mod, the WRF Fire code built with bounds checking activated
throws an error.
 - [x] With mod, the WRF Fire code built with bounds checking activated
completes the simulation without any troubles.

### DIFF
--- a/phys/module_fr_fire_util.F
+++ b/phys/module_fr_fire_util.F
@@ -1370,7 +1370,7 @@ js=ifval(jstag.ne.0,1,0)
 
 lsum=0
 do j=jps,jpe1
-  do k=kps,kpe1
+  do k=kps,kpe1-1
     do i=ips,ipe1
       rel=a(i,k,j)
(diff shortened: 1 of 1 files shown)


weiwangncar:
@davegill The change looks ok to me.

janmandel:
@davegill 
Yes more elegant than https://github.com/wrf-model/WRF/pull/1009/commits/90d3aa46375f613a088bc86dead0112b86d53275

PR #1029, opened 2019-12-10
sfclayrev fix for divide by zero

dudhia:
TYPE: bug fix

KEYWORDS: sfclayrev scheme, divide by zero

SOURCE: internal (reported by CH Liu)

DESCRIPTION OF CHANGES: 
During iterative solution for z/L there are rare cases where the same bit-for-bit value is returned for slightly different inputs to function zolri2 resulting in a divide by zero. This appears to be rare.
Fix is to return when these are found equal because solution is converged already.

LIST OF MODIFIED FILES: 
phys/module_sf_sfclayrev.F 

TESTS CONDUCTED: 
Fix works for case that stopped.
Test on standard June case is bit-for-bit as expected
No regtests yet (Jenkins?)

RELEASE NOTE: 
Fix for occasional divide-by-zero error in sfclayrev option.

### DIFF
--- a/phys/module_sf_sfclayrev.F
+++ b/phys/module_sf_sfclayrev.F
@@ -1119,6 +1119,8 @@ CONTAINS
       fx1=zolri2(x1,ri,z,z0)
       fx2=zolri2(x2,ri,z,z0)
       Do While (abs(x1 - x2) > 0.01)
+! check added for potential divide by zero (2019/11)
+      if(fx1.eq.fx2)return
       if(abs(fx2).lt.abs(fx1))then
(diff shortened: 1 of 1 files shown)


weiwangncar:
@dudhia If we do a test, say the Jan 2000 case, would you get bit-for-bit results with this change?

dudhia:
Very likely yes. It should only affect runs that stopped before.

weiwangncar:
@dudhia Thanks. Could you do a test to verify this?

dudhia:
test of June 2001 case gives bit4bit same results with fix as expected

weiwangncar:
@dudhia That's good to know. I'm ok with the fix.

PR #1038, opened 2019-12-14
Remove pop/push communicator on non MPI runs

davegill:
TYPE:bug fix

KEYWORDS: fire, pop, push, communicator

SOURCE: internal

DESCRIPTION OF CHANGES: 
Problem:
WRF gets into a state where the code tries to pop or push a stack of communicators. This should 
not happen when doing a Serial or OpenMP build.

Here is the example error message we get with a push on OpenMP:
```
FATAL CALLED FROM FILE:  <stdin>  LINE:    5489
pop_communicators_for_domain on empty stack
```

And here is the example error message with a pop with OpenMP:
```
FATAL CALLED FROM FILE:  <stdin>  LINE:    5464
push_communicators_for_domain would excede stacksize
```

Solution:
This problem only occurs with Serial or OpenMP builds. The communicators are only required for 
distributed memory MPI jobs. Therefore, put an ifdef around the routine that handles pops 
and pushes of the the communicator stack.

LIST OF MODIFIED FILES: 
modified:   external/RSL_LITE/module_dm.F

TESTS CONDUCTED: 
 - [x] Auto regression is HAPPY!

### DIFF
--- a/external/RSL_LITE/module_dm.F
+++ b/external/RSL_LITE/module_dm.F
@@ -2528,6 +2528,7 @@ END MODULE module_dm
    SUBROUTINE pop_communicators_for_domain
       USE module_dm
       IMPLICIT NONE
+#if ( defined( DM_PARALLEL ) && ( ! defined( STUBMPI ) ) )
       IF ( communicator_stack_cursor .LT. 1 ) CALL wrf_error_fatal("pop_communicators_for_domain on empty stack")
       current_id = id_stack(communicator_stack_cursor)
       local_communicator = local_communicator_stack( communicator_stack_cursor )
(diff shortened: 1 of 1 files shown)


davegill:
@mgduda 
No review required, but this PR made me think of you.

Screen Shot 2019-12-20 at 4 56 25 PM

kkeene44:
@davegill 
I modified the commit message. In this line:
"WRF gets into a state where the code tries to pop or push a stack of communicators."
You originally said "pop or pop." I assume it should be as it is now stated above. This is causing a repo test, though, so it currently says we are waiting on some checks - not sure if you want to squash/merge this yet.

PR #1049, opened 2019-12-27
HWRF fixes: build = replicated symbol; run = link in look-up tables

davegill:
TYPE: bug fix

KEYWORDS: HWRF, build, run

SOURCE: internal

DESCRIPTION OF CHANGES: 
To get the HWRF code to build and run the regression test cases, two changes are required:

### Build
There is a replicated symbol in subroutine feedback_domain_nmm_part2: SMOOTHER. This
named symbol cannot be both a LABEL and the name of a called SUBROUTINE within 
a single subprogram unit (subroutine or function). Just a quick misspelling of SMOOTHR works
wonders.

### Run
Link in (ln -sf) all of the look-up tables to the test/nmm_real directory in the last step of the build
process. Most of the run-time-required links to physics data files were not originally in place.

LIST OF MODIFIED FILES:
modified:   Makefile
modified:   external/RSL_LITE/module_dm.F

TESTS CONDUCTED: 
 - [x] Without build mod (replicated symbol), code does not build with GNU
 - [x] Without run mod (linked-in look-up tables), the code does not run regression tests
 - [x] With both mods, the code builds and runs.
 - [x] Regression test is now in place for HWRF.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -880,7 +880,7 @@ nmm_real : nmm_wrf
 	( cd test/nmm_real ; /bin/rm -f ETAMPNEW_DATA.expanded_rain ETAMPNEW_DATA RRTM_DATA ;    \
 	     ln -sf ../../run/ETAMPNEW_DATA . ;                     \
 	     ln -sf ../../run/ETAMPNEW_DATA.expanded_rain . ;       \
-	     ln -sf ../../run/RRTM_DATA . ;                         \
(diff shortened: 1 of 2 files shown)


llpcarson:
Thanks for updating these, the run-dir links are likely overkill, since many of those options are not used/tested with HWRF, but this is a simple fix.

davegill:
@llpcarson 
Laurie,
FYI: these mods get us a working HWRF in the repo.

PR #1056, opened 2020-01-22
Changing defaults for damp_opt and dampcoef 

weiwangncar:
TYPE: enhancement

KEYWORDS: damp_opt, dampcoef

SOURCE: internal

DESCRIPTION OF CHANGES: 
In the spirit of providing best model options to users, we want to change the default for damp_opt and dampcoef. These options have been used so we shouldn't expect issues. But if you don't use it, it will change results.

TESTS CONDUCTED: 
Tests?

RELEASE NOTE: 
Defaults for damp_opt has been changed from 0 to 3, or w-Rayleigh damping, and its coefficient dampcoef to 0.2. (Ref: Klemp, J. B., J. Dudhia, and A. D. Hassiotis, 2008, MWR)


davegill:
@weiwangncar 
Can you just list examples where the new settings were utilized? I don't think that we need any plots.

weiwangncar:
@davegill Not sure what you mean?

weiwangncar:
@davegill I have updated the testing bit.

PR #1059, opened 2020-01-23
Add year dependent CO2 function as default for RRTMG only

dudhia:
TYPE: enhancement

KEYWORDS: CO2, RTTMG longwave and shortwave

SOURCE: internal

DESCRIPTION OF CHANGES: 
CO2 is changing as a function of year. The default RRTMG value is 379 ppm valid in 2005. 
Current values exceed 400 ppm. A simple exponential function fits the past century to about 1%.
The function applied is CO2(ppm) = 280 + 90 exp (0.02 (year-2000)). Below are two plots 
using this function. 

The first figure shows the analytic function and measured values, as a function of time.
Screen Shot 2020-01-23 at 9 44 40 AM

The second figure shows the difference in the analytic and measured values for CO2.
Screen Shot 2020-01-23 at 10 10 25 AM

This PR partially replaces #705 "User defined values for CO2 content in RRTMG and NOAH-MP".
The modifications are only to RRTMG (and the fast version).

LIST OF MODIFIED FILES: 
module_ra_rrtmg_lw.F
module_ra_rrtmg_sw.F
module_ra_rrtmg_lwf.F
module_ra_rrtmg_swf.F

TESTS CONDUCTED: 
With and without change. 2001 test case has 372 ppm versus 379 ppm default. T2 difference.
Screen Shot 2020-01-23 at 12 00 56 PM

RELEASE NOTE: RRTMG (and fast version) longwave and shortwave CO2 value is now a function of year, which replaces the outdated constant value of 379 ppm (applies to 2005).

### DIFF
--- a/phys/module_ra_rrtmg_lw.F
+++ b/phys/module_ra_rrtmg_lw.F
@@ -11744,9 +11744,9 @@ CONTAINS
 #else
 
 ! Set trace gas volume mixing ratios, 2005 values, IPCC (2007)
-! carbon dioxide (379 ppmv)
+! carbon dioxide (379 ppmv) - this is being replaced by an annual function in v4.2
     real :: co2
-    data co2 / 379.e-6 / 
+!   data co2 / 379.e-6 / 
 ! methane (1774 ppbv)
     real :: ch4
     data ch4 / 1774.e-9 / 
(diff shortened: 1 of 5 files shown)


weiwangncar:
The changes look ok to me...

PR #1060, opened 2020-01-23

MicroTed:
REPLACED with #1169 "Efficiency fix for nwp_diag reflectivity calculation"
TYPE:  enhancement

SOURCE: Ted Mansell

DESCRIPTION OF CHANGES: At some point, solve_em was changed so that diag_flag was always set to true when nwp_diagnostics=1. This caused the full 3D radar reflectivity to be calculated on each time step, whereas the diagnostic only needs the lowest level (k=1).  Here, I added a variable 'ke_diag' in solve_em to tell the reflectivity calculation whether to do the full depth (for history output) or only the lowest level (for nwp_diagnostics=1). This avoids unnecessary computation of dBZ for nwp_diag=1, which is expensive for a number of schemes. ke_diag is passed to microphysics driver and from there to individual microphysics schemes. The NSSL and Thompson schemes have been updated to use it, and others could be done in a very similar manner.

LIST OF MODIFIED FILES: 
 dyn_em/solve_em.F
 phys/module_mp_nssl_2mom.F
 phys/module_mp_thompson.F

### DIFF
--- a/dyn_em/solve_em.F
+++ b/dyn_em/solve_em.F
@@ -295,9 +295,9 @@ SUBROUTINE solve_em ( grid , config_flags  &
 !
 ! Set diagnostic flag value history output time
 !-----------------------------------------------------------------------------
-!  if ( Is_alarm_tstep(grid_clock, grid_alarms(HISTORY_ALARM)) ) then
-      ke_diag = kms ! default to ke_diag=1 in case of nwp_diagnostics == 1
-      diag_flag = .false.
+
+   ke_diag = kms ! default to ke_diag=1 in case of nwp_diagnostics == 1
+   diag_flag = .false.
    if ( Is_alarm_tstep(grid%domain_clock, grid%alarms(HISTORY_ALARM)) ) then
       diag_flag = .true.
       ke_diag = min(k_end,kde-1) ! set depth to full domain for reflectivity field
--- a/dyn_em/solve_em.F
+++ b/dyn_em/solve_em.F
@@ -127,6 +127,7 @@ SUBROUTINE solve_em ( grid , config_flags  &
    
    ! Flag for producing diagnostic fields (e.g., radar reflectivity)
    LOGICAL                        :: diag_flag
+   INTEGER                        :: ke_diag ! tells reflectivity calculation whether to do full depth or only k=1
    
 #if (WRF_CHEM == 1)
    ! Index cross-referencing array for tendency accumulation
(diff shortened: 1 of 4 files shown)


MicroTed:
@gthompsnWRF please also review my changes to module_mp_thompson.F

MicroTed:
Second update to module_mp_thompson.F fixes the minor loop issue by adding a loop variable instead of clobbering k_0. The original change caused small differences because k_0 is used to reference an array. Now I see precisely the same values of REFL_MAX in the modified code.

davegill:
@MicroTed @gthompsnWRF @dudhia @weiwangncar 
Ted,
Are only these two MP schemes impacted?

Greg,
Please take a look at the proposed mods to your scheme.

davegill:
@MicroTed 
I am agreeing that this should go on the v414 branch. While not technically a bug, this is certainly something that needs to get fixed.

MicroTed:
@davegill 
Dave,

These are the only two MP schemes that I modified. All scheme are impacted by nwp_diagnostics=1, but especially the ones that use the "rayleigh_soak_wetgraupel" routine (gsfcgce/4ice, lin, morr, wsmX/wdmX). I haven't touched those schemes, but their maintainers could follow the example in Greg's scheme, since they all seem to use the same method of finding the melting level (k_0).

The NSSL scheme doesn't see much difference in timing because the dBZ calc is not so expensive. In the Thompson scheme, I got about a 30% reduction in the wall clock per time step for the test I did, and I think the slowdown could be worse than that with wide-spread convection.

-- Ted 

> @MicroTed @gthompsnWRF @dudhia @weiwangncar
> Ted,
> Are only these two MP schemes impacted?

weiwangncar:
@MicroTed What does 'ke' stand for?
I just want to add a note here so that we know how this diagnostics has evolved. Long time ago when we received these diagnostic code from NSSL, some way to compute radar reflectivity at every model time step was there. The reason was simple, the option wanted to compute maximum radar reflectivity in a model output interval. Some time later, when Greg's radar calculation was added, we realized it was slow so we were simply not computing it at every model time step. By doing so the nwp_diagnostics option lost this useful variable, even though the variable remained in the output. Recently there were some questions about this variable, and this has prompt us to fix it in 4.1, but we didn't consider the efficiency issue.

MicroTed:
@weiwangncar 
> What does 'ke' stand for?

It is for k-end (in the line of kme, kte, etc.) It could be kte_diag or whatever you think would be most easily interpreted.

weiwangncar:
@MicroTed OK. In WRF the first model level is actually kts, not kte.

davegill:
@weiwangncar @dudhia @MicroTed @gthompsnWRF 
Folks,
Are there any "big picture" issues with this PR? Is this a candidate for v4.1.4?

MicroTed:
>  OK. In WRF the first model level is actually kts, not kte.
@weiwangncar 
Yes. Maybe it is clearer if I say that ke_diag gives the vertical loop end limit for the refl_10cm calculation (either 1 or kte.). The loop start is assumed to always be 1 (kts).

weiwangncar:
@MicroTed That's ok.

weiwangncar:
If no one requires this change to be applied to other microphysics option, then this is ok to go.

MicroTed:
@weiwangncar 
Would you want to ask the other developers if they want to add this to their schemes? They may be unaware of the efficiency issue. Of course they can do it later. And I would wait for Greg to return and have a chance to double-check the changes. My refl_10cm calculation is cheap enough that it doesn't have a large impact like some other schemes, so I am happy to wait.

weiwangncar:
@MicroTed It's probably more important to get Greg's approval. We want this to be in 4.1.4. But it could wait for 4.2.

davegill:
REPLACED with #1169 "Efficiency fix for nwp_diag reflectivity calculation"

PR #1072, opened 2020-02-04
Handle possible locations for rpc/types.h

davegill:
TYPE: enhancement

KEYWORDS: type.h, rpc, tirpc

SOURCE: Reported on WRF Forum, fixed internally

DESCRIPTION OF CHANGES:
Problem:
The C header file `rpc/types.h` may be located in a directory other than 
`/usr/include`. If the WRF code cannot find this file, the source code for 
`landread.c` does not compile, and the WRF code fails to build.

Mentioned on WRF Forum as "Problem Compiling WRFV4.0 on Fedora 28"
https://forum.mmm.ucar.edu/phpBB3/viewtopic.php?f=37&t=61&p=15718#p15718

Solution:
Refer to the discussion:
https://forums.gentoo.org/viewtopic-t-1076270-start-0.html

The likely other location for this header file is `/usr/include/tirpc`. During
the `configure` stage, a test is made to determine which of these two 
candidate directories is the correct location of this header file. A cpp flag 
is added to `ARCH_LOCAL` so that the correct location is `#include`d into the 
source code of `landread.c`.

LIST OF MODIFIED FILES:
modified:   configure
modified:   share/landread.c

TESTS CONDUCTED:
1. Automatic jenkins regression test OK, with expected failures due to fixes that are currently
only on the release-v4.1.4 branch (not yey merged to develop):
```
output_2:2 = STATUS test_002m hwrf nmm_real 34 1NE
output_2:2 = STATUS test_002m hwrf nmm_real 34 2NE
output_9:9 = STATUS test_009o fire em_fire 33 01
```
2. Tested with a simple print statement that the code still builds as expected
for OS / glibc versions that did not exhibit this trouble (no harm done).

### DIFF
--- a/configure
+++ b/configure
@@ -983,6 +983,24 @@ if [ $retval -ne 0 ] ; then
   echo "*****************************************************************************"
 fi
 
+# testing for location of rpc/types.h file, used in landuse
+if [ -f /usr/include/rpc/types.h ] ; then
+  sed -e '/^ARCH_LOCAL/s/$/ -DRPC_TYPES=1/' configure.wrf > configure.wrf.edit
+  mv configure.wrf.edit configure.wrf
+echo standard location of RPC
+elif [ -f /usr/include/tirpc/rpc/types.h ] ; then
+  sed -e '/^ARCH_LOCAL/s/$/ -DRPC_TYPES=2/' configure.wrf > configure.wrf.edit
+  mv configure.wrf.edit configure.wrf
+echo newer location of RPC
+else
+  echo "************************** W A R N I N G ************************************"
+  echo " "
+  echo "The moving nest option is not available due to missing rpc/types.h file."
+  echo "Copy landread.c.dist to landread.c in share directory to bypass compile error."
+  echo " "
+  echo "*****************************************************************************"
+fi
+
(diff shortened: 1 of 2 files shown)


davegill:
@kkeene44 
Kelly,
Once I get a success from the reg test, would you please try to get the Forum users to test this PR? I'll send you a note in about an hour - when the test should be done.

shijin-aws:
Hello, I hit the location of `rpc/` issue on CentOS 8 as well. I installed `libtirpc-devel` and we have `/usr/include/tirpc/rpc` available. Then I applied this commit's patch to WRFV-3.9.1.1, but I still get errors when compiling as 

```
landread.c:130:8: error: unknown type name 'XDR'
landread.c:410:11: error: 'XDR' undeclared (first use in this function)
landread.c:410:16: error: expected expression before ')' token
landread.c:411:29: error: 'XDR_DECODE' undeclared (first use in this function)
```
, which looks strange to me because I checked both `XDR` and `XDR_DECODE` is defined in my `/usr/include/tirpc/rpc/xdr.h` file. Could you provide any ideas for this issue?

PR #1074, opened 2020-02-05
A fix for conservation of enthalpy in BMJ cumulus scheme

weiwangncar:
TYPE: bug fix

KEYWORDS: BMJ, conservation of enthalpy

SOURCE: Tieh-Yong Koh (Earth Observatory of Singapore, Nanyang Technological University, Singapore); Ricardo Fonseca (Earth Observatory of Singapore, Nanyang Technological University, Singapore); Chee-Kiat Teo (Temasek Laboratories, Nanyang Technological University, Singapore); Tengfei Zhang (Earth Observatory of Singapore, Nanyang Technological University, Singapore)

DESCRIPTION OF CHANGES: 
This PR separates out a bug fix from PR#921 so that even if a user would not want to use the precipitation convective cloud (PCC) option, the benefit of bug fix can be utilized. The fix corrects an error with conservation of enthalpy. The DHDT term is corrected, which should be an integral from p_bot to p_top and not computed at every level as done in the original version of the code. For more detail, refer to the following paper:
https://www.geosci-model-dev.net/8/2915/2015/

The plot below shows the difference in convective rainfall for a mid-latitude convection case. The left plot is with the fix, and the right one is the diff plot, all at forecast hour 12. The fixed code produces less rain.

diff_bugfix_12h

The original bug fix used part of an array (APESK) that is not defined. This is corrected here. The plot above is also updated.

LIST OF MODIFIED FILES: 
M       phys/module_cu_bmj.F

TESTS CONDUCTED: 
Individual case test. Passed jenkins test which tests the BMJ cumulus scheme.

RELEASE NOTE: The bug fix corrects an error with conservation of enthalpy and more detail can be found in the following paper: https://www.geosci-model-dev.net/8/2915/2015/. The fixed code produces less surface rain.

### DIFF
--- a/phys/module_cu_bmj.F
+++ b/phys/module_cu_bmj.F
@@ -18,7 +18,7 @@
      &                 ,EPSNTP=.0001,EPSNTT=.0001,EPSPR=1.E-7           &
      &                 ,EPSUP=1.00                                      &
      &                 ,FR=1.00,FSL=0.85,FSS=0.85                       &
-     &                 ,FUP=0.                                          &
+     &                 ,FUP=0.,CRMN=0.14                                &
      &                 ,PBM=13000.,PFRZ=15000.,PNO=1000.                &
      &                 ,PONE=2500.,PQM=20000.                           &
      &                 ,PSH=20000.,PSHU=45000.                          &
@@ -126,6 +126,7 @@ CONTAINS
       INTEGER :: LBOT,LPBL,LTOP
 !
       REAL :: DTCNVC,LANDMASK,PCPCOL,PSFC,PTOP
+      REAL :: PWCOL,DQCOL,DQCOLMIN
 ! 
       REAL,DIMENSION(KTS:KTE) :: DPCOL,DQDT,DTDT,PCOL,QCOL,TCOL
 !
@@ -169,9 +170,12 @@ CONTAINS
             DTDT(K)=0.
           ENDDO
 !
+          DQCOL=0.
+          PWCOL=0.
+          PCPCOL=0.
+          DQCOLMIN=0.
           RAINCV(I,J)=0.
           PRATEC(I,J)=0.
-          PCPCOL=0.
           PSFC=PINT(I,LOWLYR(I,J),J)
           PTOP=PINT(I,KTE+1,J)      ! KTE+1=KME
 !
@@ -212,6 +216,7 @@ CONTAINS
           CALL BMJ(ITIMESTEP,I,J,DTCNVC,LMH,LANDMASK,CLDEFI(I,J)        &
      &            ,DPCOL,PCOL,QCOL,TCOL,PSFC,PTOP                       &
      &            ,DQDT,DTDT,PCPCOL,LBOT,LTOP,LPBL                      &
+     &            ,PWCOL,DQCOL,DQCOLMIN                                 &
      &            ,CP,R,ELWV,ELIV,G,TFRZ,D608                           &   
      &            ,PRINT_DIAG                                           &   
      &            ,IDS,IDE,JDS,JDE,KDS,KDE                              &     
@@ -288,6 +293,7 @@ CONTAINS
      & (ITIMESTEP,I,J,DTCNVC,LMH,SM,CLDEFI                              &
      & ,DPRS,PRSMID,Q,T,PSFC,PT                                         &
      & ,DQDT,DTDT,PCPCOL,LBOT,LTOP,LPBL                                 &
+     & ,PWCOL,DQCOL,DQCOLMIN                                            &
      & ,CP,R,ELWV,ELIV,G,TFRZ,D608                                      &
      & ,PRINT_DIAG                                                      &   
      & ,IDS,IDE,JDS,JDE,KDS,KDE                                         &
@@ -309,7 +315,7 @@ CONTAINS
 !
       REAL,DIMENSION(KTS:KTE),INTENT(IN) :: DPRS,PRSMID,Q,T
 !
-      REAL,INTENT(INOUT) :: CLDEFI,PCPCOL
+      REAL,INTENT(INOUT) :: CLDEFI,PCPCOL,PWCOL,DQCOL,DQCOLMIN
 !
       REAL,DIMENSION(KTS:KTE),INTENT(INOUT) :: DQDT,DTDT
 !
@@ -348,7 +354,7 @@ CONTAINS
      &            ,PART1,PART2,PART3,PBOT,PBOTFC,PBTK                   &
      &            ,PK0,PKB,PKL,PKT,PKXXXX,PKXXXY                        &
      &            ,PLMH,PELEVFC,PBTmx,plo,POTSUM,PP1,PPK,PRECK          &
-     &            ,PRESK,PSP,PSUM,PTHRS,PTOP,PTPK,PUP                   &
+     &            ,PRWD,PDQD,PRESK,PSP,PSUM,PTHRS,PTOP,PTPK,PUP         &
      &            ,QBT,QKL,QNEW,QOTSUM,QQ1,QQK,QRFKL                    &
      &            ,QRFTP,QSP,QSUM,QUP,RDP0T                             &
      &            ,RDPSUM,RDTCNVC,RHH,RHL,RHMAX,ROTSUM,RTBAR,RHAVG      &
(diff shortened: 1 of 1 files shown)


weiwangncar:
@rmfonsecaweb Please review this PR and see if it got the bug-fix portion of your PCC PR correctly. Like I indicated in the PR, with these changes, I still get different forecasts, although the differences now are small.

rmfonsecaweb:
We thank the reviewer for creating this PR. As we now state in the reply to the reviewer's question in the other PR, the bug we mentioned in the definition of "DHDT" does not explain all the differences between the two simulations. The additional differences may be related to the initialization of local arrays and hence are not a bug of the code. 
If the reviewer wishes to have a separate PR for this bug, then please just highlight that it is related to the conservation of enthalpy, in particular to the way the variable "DHDT" is defined. The changes are the ones referred in the other PR pasted below:

    On Thursday, February 6, 2020, 03:49:01 AM GMT+4, weiwangncar <notifications@github.com> wrote:  
 
 
@rmfonsecaweb Please review this PR and see if it got the bug-fix portion of your PCC PR correctly. Like I indicated in the PR, with these changes, I still get different forecasts, although the differences now are small.

weiwangncar:
@rmfonsecaweb It would be nice if the bug fix is separated from the PCC change. In addition to the code pointed out by you for the two loops, is the code related to how CLDEFI is computed part of the fix? I know without it, the model produced even more different convective rainfall. If you can let me know which arrays are initialized in your code that I should include in my test, please let me know.

rmfonsecaweb:
Noted. We can keep the bug fix separated from the PCC change.

    On Thursday, February 6, 2020, 11:38:23 PM GMT+4, weiwangncar <notifications@github.com> wrote:  
 
 
@rmfonsecaweb It would be nice if the bug fix is separated from the PCC change. In addition to the code pointed out by you for the two loops, is the code related to how CLDEFI is computed part of the fix? I know without it, the model produced even more different convective rainfall. If you can let me know which arrays are initialized in your code that I should include in my test, please let me know.

davegill:
@weiwangncar 
Wei,
Could you provide a bit more info in the PR commit message. I do realize that this is a bit of a moving target right now.

weiwangncar:
@rmfonsecaweb If the code change for CLDEFI isn't part of bug fix, is that part of PCC change? If I remove the CLDEFI change in your PCC code (about 3 places), I get bit-for-bit results between bug-fixed BMJ run and PCC run with bmj_rad_feedback set to false.

rmfonsecaweb:
No, CLDEFI is not part of the bug fix. As stated before, we only made that change so that we can distinguish when the deep convection scheme was called but was then aborted because the cloud was too thin and/or no CAPE was present, or whether the shallow convection scheme was called directly instead. This change is not part of the PCC scheme, it was a separate change done just for plotting purposes.

If the reviewer finds that reverting back to the old definition of CLDEFI gives identical results between the bug-fixed BMJ run and the PCC run with "bmj_rad_feedback" set to false (again, we did not reach this conclusion in our sensitivity runs, perhaps we missed something), can the reviewer revert back to the old CLDEFI code (change in 3 places)? Thank you!


    On Saturday, February 8, 2020, 10:15:37 AM GMT+4, weiwangncar <notifications@github.com> wrote:  
 
 
@rmfonsecaweb If the code change for CLDEFI isn't part of bug fix, is that part of PCC change? If I remove the CLDEFI change in your PCC code (about 3 places), I get bit-for-bit results between bug-fixed BMJ run and PCC run with bmj_rad_feedback set to false.

rmfonsecaweb:
We would like to thank the reviewer for taking the time to conduct sensitivity experiments and finding the reason for the discrepancy between the simulations without the PCC scheme and with the PCC scheme with the flag "bmj_rad_feedback" set to FALSE, and for updating the description of this bug fix. We do not have anything to add to the latter. Thank you!

davegill:
@weiwangncar 
Wei,
Great detective work.
:shipit:

weiwangncar:
@rmfonsecaweb It turns out removing CLDEFI change does not solve all the problems. So I turned to debugger. When I compiled the code with debug mode (using 'configure -D' using gfortran), I encountered an immediate failure with floating-point exception. I then went back the the bug-fix code, and looked at code a bit more closely. Here is what I find: the floating point exception error occurs on the line where DHDT is computed. It turns out that when you loop through LTOP to LB, the array APESK isn't always defined. This can be seen in the loop just above the enthalpy conservation loop: it is only defined when PK(L)>PQM. One way to fix this is to move DHDT computation out of the current loop further done after LCOR is defined:

!***
!***  ABOVE LQM CORRECT TEMPERATURE ONLY
!***
          IF(LCOR<=LQM)THEN
            DO L=LCOR,LQM
              TREFK(L)=TREFK(L)+HCORR*RCP
            ENDDO
            LCOR=LQM+1
          ENDIF
!
!***  COMPUTE DHDT AS A COLUMN INTEGRAL
!
          DHDT =0.
          DO L=LCOR,LB
            DHDT=(QREFK(L)*A23M4L/((TREFK(L)*APEK(L)/APESK(L))-A4)**2+CP)*DPRS(L) &
     &            +DHDT
          ENDDO
          DHDT=DHDT/(SUMDP-DPRS(LCOR))

Let me know what you think, and maybe you would like to fix this differently. This issue, I'm sure, is the cause for non-deterministic outcome. It needs to be fixed.

rmfonsecaweb:
We again thank the reviewer for taking the time to look into this. We can also confirm that reverting back to the old CLDEFI does not fix all the problems. As stated in the PDF below (equation A18), we need to loop from LTOP to LB to compute DHDT. As LCOR=LTOP+1, we are missing one level.
https://atmospheres.research.ltu.se/owncloud/index.php/s/tP6EhuPnf6nOZJQ
We have to APESK at LTOP, probably by getting it out of the loop in the piece of code above. We will look into this and will get back to the reviewer later today. Thank you! 
Ricardo
    On Monday, February 10, 2020, 09:49:56 AM GMT+4, weiwangncar <notifications@github.com> wrote:  
 
 
@rmfonsecaweb It turns out removing CLDEFI change does not solve all the problems. So I turned to debugger. When I compiled the code with debug mode (using 'configure -D' using gfortran), I encountered an immediate failure with floating-point exception. I then went back the the bug-fix code, and looked at code a bit more closely. Here is what I find: the floating point exception error occurs on the line where DHDT is computed. It turns out that when you loop through LTOP to LB, the array APESK isn't always defined. This can be seen in the loop just above the enthalpy conservation loop: it is only defined when PK(L)>PQM. One way to fix this is to move DHDT computation out of the current loop further done after LCOR is defined:

!***
!*** ABOVE LQM CORRECT TEMPERATURE ONLY
!***
IF(LCOR<=LQM)THEN
DO L=LCOR,LQM
TREFK(L)=TREFK(L)+HCORRRCP
ENDDO
LCOR=LQM+1
ENDIF
!
!** COMPUTE DHDT AS A COLUMN INTEGRAL
!
DHDT =0.
DO L=LCOR,LB
DHDT=(QREFK(L)*A23M4L/((TREFK(L)*APEK(L)/APESK(L))-A4)**2+CP)*DPRS(L) &
& +DHDT
ENDDO
DHDT=DHDT/(SUMDP-DPRS(LCOR))

Let me know what you think, and maybe you would like to fix this differently. This issue, I'm sure, is the cause for non-deterministic outcome. It needs to be fixed.

rmfonsecaweb:
@weiwangncar
We can confirm that when the aforementioned change is made to the codes we get exactly the same result in the runs without the PCC scheme and with the PCC scheme with the flag "bmj_rad_feedback" set to FALSE! A HUGE thank you for the amount of time and effort you have spent working on this and on the PCC scheme. THANK YOU!!!

Before changing the codes in this pull request and in the other one, can you please confirm on your side that the model-predicted convective precipitation is indeed the same in both simulations? Just to double-check before we update the files. Again, the only change required is to move the lines

            PSK(L)=PK(L)+DSP
            APESK(L)=(1.E5/PSK(L))**CAPA

outside the IF-statement. They are still in the L=LTOP,LBOT loop meaning that APESK will be defined at all levels required for its use in the computation of DHDT below. Thank you!

rmfonsecaweb:
I could not agree more!
    On Sunday, February 9, 2020, 08:00:21 PM GMT+4, Dave Gill <notifications@github.com> wrote:  
 
 
@weiwangncar
Wei,
Great detective work.

weiwangncar:
@rmfonsecaweb Yes I can confirm that with new changes made to the bug-fix code (so that APESK is defined for looping from LTOP to LB) I'm getting bit-for-bit results using PCC code with bmj_rad_feedback set to false when compared with the bug-fix code. Also removing CLDEFI changes is required for getting bit-for-bit results.

weiwangncar:
@rmfonsecaweb I also checked the new code change (i.e. moving PSK and APESK out of the if block) gives bit-for-bit results when compared with the original BMJ code. So all is good now. Very happy!

PR #1077, opened 2020-02-07
Output standard U10/V10 from MYJSFC and QNSESFC schemes

weiwangncar:
TYPE: enhancement

KEYWORDS: u10, v10, myjsfc, qnsesfc

SOURCE: reported by Rob Fovell and others, fixed internally

DESCRIPTION OF CHANGES: 
The U10/V10 output when MYJ and QNSE surface layer schemes is used is non-standard over land, which produces higher values. This PR retains the value computed using standard MO 
theory, and moves the modified U10/V10 to U10E/V10E. 

LIST OF MODIFIED FILES: 
M       Registry/Registry.NMM
M       dyn_em/start_em.F
M       dyn_nmm/module_PHYSICS_CALLS.F
M       dyn_nmm/start_domain_nmm.F
M       phys/module_sf_myjsfc.F
M       phys/module_sf_qnsesfc.F

TESTS CONDUCTED: 
1. Compiled. 
2. Tested with MYJSFC on and off. 
   * When MYJSFC is on, both U10/U10E, V10/V10E are in output.
   * When MYJSFC is not used, only U10/V10 are in output.
3. Has passed jenkins test

RELEASE NOTE: 
The U10/v10 diagnostic variables output from the MYJ and QNSE surface layer schemes have been replaced. The standard fields from MO theory are now output as U10/V10, and the original MYJ SFC scheme diagnostic values of U10/V10 are output with the names U10E/V10E. For those using the MYJ surface layer scheme, this change will modify the resultant U10/V10 values that are used inside of the diagnostics and the time series options.

### DIFF
--- a/Registry/Registry.EM_COMMON
+++ b/Registry/Registry.EM_COMMON
@@ -2862,7 +2862,7 @@ package   nosfcscheme        sf_sfclay_physics==0        -             -
 package   sfclayrevscheme    sf_sfclay_physics==1        -             -
 package   myjsfcscheme       sf_sfclay_physics==2        -             state:tke_pbl,u10e,v10e
 package   gfssfcscheme       sf_sfclay_physics==3        -             -
-package   qnsesfcscheme      sf_sfclay_physics==4        -             -
+package   qnsesfcscheme      sf_sfclay_physics==4        -             state:u10e,v10e
 package   mynnsfcscheme      sf_sfclay_physics==5        -             -
 package   pxsfcscheme        sf_sfclay_physics==7        -             -
 package   temfsfcscheme      sf_sfclay_physics==10       -             state:wm_temf
--- a/Registry/Registry.EM_COMMON
+++ b/Registry/Registry.EM_COMMON
@@ -965,6 +965,9 @@ state   real   QSFC             ij     misc        1         -      r        "QS
 state   real   AKHS             ij     misc        1         -      r        "AKHS"                  "SFC EXCH COEFF FOR HEAT"                      "m s-1"    
 state   real   AKMS             ij     misc        1         -      r        "AKMS"                  "SFC EXCH COEFF FOR MOMENTUM"                  "m s-1"    
 state   integer KPBL            ij     misc        1         -     r         "KPBL"                  "LEVEL OF PBL TOP"                             ""
+state   real   U10E             ij     misc        1         -     irhdu     "U10E"                  "Special U at 10 M from MYJSFC"         "m s-1"
+state   real   V10E             ij     misc        1         -     irhdu     "V10E"                  "Special V at 10 M from MYJSFC"         "m s-1"
+
 #BSINGH - For CuP
 #~wig: added real pbl level index for output tests
 state   real   AKPBL            ij     misc        1         -      r        "AKPBL"                  "LEVEL OF PBL TOP"                             ""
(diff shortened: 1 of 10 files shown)


Registry/Registry.EM_COMMON
```diff
@@ -965,6 +965,9 @@ state   real   QSFC             ij     misc        1         -      r        "QS
 state   real   AKHS             ij     misc        1         -      r        "AKHS"                  "SFC EXCH COEFF FOR HEAT"                      "m s-1"    
 state   real   AKMS             ij     misc        1         -      r        "AKMS"                  "SFC EXCH COEFF FOR MOMENTUM"                  "m s-1"    
 state   integer KPBL            ij     misc        1         -     r         "KPBL"                  "LEVEL OF PBL TOP"                             ""
+state   real   U10E             ij     misc        1         -     irhdu     "U10E"                  "Special U at 10 M from MYJSFC"         "m s-1"
```
davegill: @weiwangncar 
Wei,
We can address this now or later. For a diagnostic computed in the MYJ SFC scheme:
1. We should not need "i" - this is not an input from the real program (no mods to module_intialize_real)
2. We should not need "r" - this diagnostic should have no impact on restarts
3. We should not need "d" - as a diagnostic, we do not need to get an initial value from the parent domain

To be fair, these are all on the U10/V10 fields. So, we can address this now or later.

davegill:
@weiwangncar @dudhia 
Folks,
Wei separated this out cleanly enough that I could follow what was going on. 
1. The values of U10 and V10 are no longer overwritten by the MYJ SFC scheme.
2. The new fields, U10E and V10E, are packaged
3. ARW and NMM mods are present for physics changes

Since U10 and V10 are initialized in the real program, those fields have valid values at the model initialization time. Should we add a loop in phy_init that assigns U10E and V10E?
```
if (myj sfc) then
   u10e = u10
   v10e = v10
end if
```

This would require mods to phy_init, and both start_*.F files.

No biggy either way to me.

weiwangncar:
@davegill Is start_em called before the first history file is written? If so, it is probably good to fill u10e/v10e for 0 h output.

Registry/Registry.EM_COMMON
weiwangncar: @davegill Agreed that i and r may not be needed. but d may be if the option is used with moving nests.

davegill: @weiwangncar 
Wei,
Let's leave the "d" and check a moving nest case when we have more time. The "d" nesting option _should_ be unnecessary for a physics diagnostic.

davegill:
@weiwangncar 
>Is start_em called before the first history file is written?

Wei,
Yes. 
1. We use to have differences between the output from the real program and the initial output from the model due to computations that went on inside of the start_em.F file. 
2. There are calls to a few diagnostic routines from start, just so that the 0 h values of those diags are non-zero.

The phy_init routine is called from start, so any changes to U10E/V10E from phy_init would be seen in the initial output time from WRF.

dudhia:
Yes, good to set u10e and v10e at init.

dudhia:
Is u10e even different from u10 for sea ice? If not, that part of the change is not needed. It doesn't seem to make sense that there is a shelter level on sea ice.

dudhia:
Even if it is different for sea ice, I think we should just set it to u10 and v10, because it makes no sense to keep that part.

weiwangncar:
@dudhia we can remove the part for seaice correction.

dudhia:
files changed list is not complete yet

weiwangncar:
@dudhia one thing at a time...

PR #1079, opened 2020-02-07
Replace module interfaces for netcdf and mpi

davegill:
TYPE: enhancement

KEYWORDS: netcdf, mpi, module

SOURCE: internal

DESCRIPTION OF CHANGES: 
There are two similar types of changes. Most of the WRF modeling system uses the 
`netcdf.inc` file instead of `use netcdf`, and most of the WRF modeling system uses 
`mpif.h` instead of `use mpi`. The Fortran module constructs are removed for consistency 
with the rest of the WRF system and also to lower the bar for porting the WRF code to small 
Linux boxes. 

With this PR, the WRF code now builds and runs with newer GNU compilers (starting to 
become important with Fortran 2003 and Fortran 2008 constructs) but still using the 
default installed libraries on the small desktop Linux machines (where the libs are built with 
GNU 4.8-ish). In addition to assisting the user base that has access to user-managed 
Linux systems, these mods simplify building containers for WRF which can be used for 
automated testing or for cloud based purposes. Lastly, this modification allows the 
Windows-based CYGWIN environment to use a fairly recent version of the GNU compiler that 
was not necessarily used to build the automatically installed netcdf or mpi libraries.

Within the WRF source code, the majority of netcdf and mpi calls are through the use of
cpp include directives. This PR brings consistency to the WRF source code.

MPI Differences
When the `use mpi` construct is removed, it is replaced by a Fortran include statement 
`include mpif.h`.

NETCDF Differences
When the `use netcdf` construct is removed, it would typically be replaced by a Fortran 
include statement `include netcdf.inc` (similar to `include mphf.h`). However, in both 
instances where `use netcdf` was removed by this PR, no netcdf code was used within the 
scoping unit covered by the Fortran USE statement. Therefore, the two unnecessary
`use netcdf` statements were removed without any required source code replacement.

LIST OF MODIFIED FILES: 
MPI
M dyn_nmm/module_HIFREQ.F
M dyn_nmm/module_initialize_real.F
M frame/module_bdywrite.F
M phys/module_ra_cam_support.F

NETCDF
M phys/module_gocart_coupling.F
M phys/module_physics_init.F

TESTS CONDUCTED: 
There were two types of modifications. 

1. The MPI changes required replacing the USE module construct with a Fortran include. For 
the MPI changes, it is necessary to make sure that the parallel vs serial, and the before vs 
after results are the same. Those tests were conducted.
    * For ARW, a 12-h Jan 2000, single domain: ARW12
    * For ARW, a 1-h Jan 2000, nested domain: ARW01
    * For NMM, a 3-h, 3-domain run: NMM03
 - [x] ARW12 bit-for-bit: serial, MPI 1pe, MPI 3pe
 - [x] ARW01 bit-for-bit for both domains: serial, MPI 1pe, MPI 3 pe
 - [x] NMM03 bit-for-bit for three domains: MPI 1pe, MPI 3pe
 - [x] ARW12 bit for bit: original vs new
 - [x] NMM03 bit for bit: original vs new

2. The NETCDF modifications were simpler. This entailed removing unnecessary code. In this 
case, removing required code would have broken the compilation process (since both of the 
impacted files have `implicit none`). In this case, the model "running" is sufficient indication 
that the code modifications were OK. Since one of the locations where the NETCDF changes 
were introduced was inside a WRF_CHEM ifdef, a WRF Chem build and test was conducted.
    * For Chem, a 10-timestep, single domain: CHEM
 - [x] CHEM completes simulation


weiwangncar:
I'm ok with the changes.

davegill:
@kkeene44 
Kelly,
Wei is OK with this. Would you please proxy approve.

PR #792, opened 2019-02-18
Fuel moisture model

janmandel:
TYPE: new feature/enhancement

KEYWORDS: WRF-Fire, fuel moisture, coupled model, equilibrium

SOURCE: Jan Mandel, Angel Farguell (University of Colorado Denver), Adam Kochanski (University of Utah)

DESCRIPTION OF CHANGES:   
Fuel moisture content (FMC) is now maintained in each fire model mesh cell separately; it was a single constraint for the whole model before. A FMC model is coupled with the coupled fire-atmosphere model. FMC has a strong effect on fire spread, including the diurnal cycle, and previously was currently a constant in the fuel description. The FMC model runs a differential equation for the FMC in each fuel class (10h, 100h, 1000h) in each atmospheric mesh cell. The FMC evolves towards an equilibrium, computed from the atmospheric state, with a speed dependent on the fuel class. The FMC is interpolated to the fire mesh and mixed in proportions given in fuel description to obtain the FMC on the fire mesh, which  is then used by the fire model. The code also supports import of the FMC map on the fire mesh. Two FMC test cases were added.

Reference: J. Mandel, S. Amram, J.D. Beezley, G. Kelman, A.K. Kochanski, V.Y. Kondratenko, B.H. Lynn, B. Regev, M. Vejmelka, Recent advances and applications of WRF–SFIRE. Natural Hazards and Earth System Science, 14, 829-2845, 2014, https://doi.org/10.5194/nhess-14-2829-2014

**An ideal test problem added**

The ideal example in test/em_fire/rain was added, which will run coupled fuel moisture and fire. The directory and the soft links there are created by ./compile em_fire and deleted by ./clean -a. Variables required to initialize surface physics are set in ideal.exe when built by ./compile em_fire.

**Flags required to be set in namelist.input**

See the file test/em_fire/namelist.input_rain for an example with annotations.

**sf_sfclay_physics**         in &physics section
A nonzero value to turn on a surface physics model is required in order for the variables T2, Q2, PSFC to be populated with values, which are needed to drive the FMC model. Other appropriate surface physics flags and input data are up to the user. The ideal.exe generated by compile em_fire sets the appropriate arrays for
**sf_sfclay_physics              =1**
**sf_surface_physics           =1**
which are therefore required in an ideal run if the fuel moisture model is turned on.

**Flags added to file namelist.input**

The following variables were added  to the &fire section:
**fmoist_run**  default    .false. :          Run the fuel FMC model (on the atmospheric grid).
[...]

PR #1080, opened 2020-02-08
Fix build and run of ARW caused Fire Fuel Model mods

davegill:
TYPE: bug fix

KEYWORDS: Fire, fuel model, GNU

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
1. Cannot build with GNU
2. Seg-faults with standard non-fire runs.

These problems occurred with commit 53d63b053fcb, PR #792 "Fuel moisture model".

Solution:
1. The use of LOGICALs in the option choices for some of the fire variables
is inapropriate for eventual usage when those namelist options are part of
a Registry package statement. The WRF code assumes that the packages are
INTEGER values - not LOGICAL.
2. The number of elements in a few DATA statements needed to be increased by 1.
The "no fuel" entry (currently set to zero) is defined in column 14 of the DATA statement.

LIST OF MODIFIED FILES:
modified:   Registry/registry.fire
modified:   phys/module_fr_fire_phys.F

TESTS CONDUCTED:
 - [x] Without mods, code does not build; with mods, code builds.
 - [x] Without mods, code seg faults; with mods, non-fire code runs.


davegill:
@janmandel
Jan,
Please take a look at this mods

janmandel:
@davegill
I will change fmoist_run and fmoist_interp to integer everywhere and submit a new PR as a bug fix based on 975f0c6e994985d99ac3cbbf4a458ee0bca4c463.

davegill:
@janmandel 
This is ALREADY merged. The integer vs logical issue is handled.

janmandel:
@davegill
I see your changes in share/module_check_a_mundo.F that patch this.
Do you want me to just let it be?

janmandel:
OK I'll let it be.

janmandel:
@davegill 
OK, will let it be.
Thanks for fixing the DATA one column short. I see it caused copying uninitialized memory later on.

PR #1025, opened 2019-11-28
Adding a fast surface shortwave parameterization (FARMS)

pedro-jm:
TYPE: new feature

KEYWORDS: Fast surface shortwave irradiance parameterization

SOURCE: FARMS code provided by Yu Xie (NREL). Pedro Jimenez (NCAR/RAL) coupled FARMS to WRF.

DESCRIPTION OF CHANGES: 
Due to the large computation cost, radiation parameterizations are not typically called every time 
step. As a result, the surface irradiances time series show discontinuities that depend on the 
radiation calls frequency. This new parameterization, FARMS, provides fast calculation of the 
surface shortwave irradiance allowing to call it every model time step. 

The scheme is activated by setting `swint_opt` in the WRF namelist.input file:
```
 &physics
 swint_opt  = 2
 /
```

LIST OF MODIFIED FILES: 
M       Registry/Registry.EM_COMMON
M       dyn_em/module_first_rk_step_part1.F
M       phys/Makefile
M       phys/module_ra_aerosol.F
A       phys/module_ra_farms.F
M       phys/module_radiation_driver.F
M       run/README.namelist

TESTS CONDUCTED: 
1. WRF run shows updates in the shortwave irradiances from FARMS every time step. The figure below shows the irradainces from RRTMG called every 15 min (left), and from FARMS (right) which is always called every model time step.

Screen Shot 2019-12-02 at 3 55 26 PM
This shows results of downward shortwave at some US surface sites. RRTMG is swint_opt=1. Generally FARMS has more cloud effect (thanks to Ju-Hye Kim for this plot).
Screen Shot 2020-02-06 at 10 02 01 AM

RELEASE NOTE: A new parameterization, FARMS (Fast All-sky Radiation Model for Solar applications), provides fast calculation of the surface shortwave irradiance allowing the WRF model to call the scheme every model time step, resulting in a temporally smooth irradiance. The scheme is activated by setting `swint_opt = 2` in the WRF namelist.input file.

PR #1081, opened 2020-02-09
Fix broken NMM build caused by FARMS mods

davegill:
TYPE: bug fix

KEYWORDS: FARMS, NMM

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
The NMM HWRF code would not build after the FARMS commit. A few variables passed 
into the radiation driver are ARW specific, so the call from the NMM core did not have:
MUT or DNW (both input associated with ARW only), or the six additional computed 
radiation output fields (sw*2).

Solution:
1. Make the ARW-specific input fields and the additional radiation output fields OPTIONAL.
2. Use an IF(PRESENT(xxx)) before using any of the fields.

These problems were introduced with commit 3c71dd1de6e0df0, PR #1025 "Adding a fast 
surface shortwave parameterization (FARMS)".

LIST OF MODIFIED FILES:
modified:   phys/module_radiation_driver.F

TESTS CONDUCTED:
1. Without mods, NMM HWRF cannot build.
2. With mods, NMM HWRF builds and runs successfully.
3. With mods, ARW still works.

### DIFF
--- a/dyn_em/module_first_rk_step_part1.F
+++ b/dyn_em/module_first_rk_step_part1.F
@@ -308,10 +308,13 @@ BENCH_START(rad_driver_tim)
      &        ,CAM_ABS_FREQ_S=grid%cam_abs_freq_s                         &
      &        ,XTIME=grid%xtime                                                &
               ,CURR_SECS=curr_secs, ADAPT_STEP_FLAG=adapt_step_flag       &
+!dave
+#if ( EM_CORE == 1)
      &        ,mut=grid%mut,dnw=grid%dnw                                  & ! FARMS coupling
      &        ,swdown2=grid%swdown2, swddni2=grid%swddni2                 & ! FARMS coupling
      &        ,swddif2=grid%swddif2, swddir2=grid%swddir2                 & ! FARMS coupling
      &        ,swdownc2=grid%swdownc2, swddnic2=grid%swddnic2             & ! FARMS coupling
+#endif
 !BSINGH - For WRFCuP scheme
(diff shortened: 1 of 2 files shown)


davegill:
@weiwangncar @dudhia @llpcarson 
Folks,
I just realized that I can probably remove the ifdef mods inside of first_rk_step_part1. I declare the variables as OPTIONAL, and they would always be present with ARW builds.

dyn_em/module_first_rk_step_part1.F
```diff
@@ -308,10 +308,12 @@ SUBROUTINE first_rk_step_part1 (   grid , config_flags              &
      &        ,CAM_ABS_FREQ_S=grid%cam_abs_freq_s                         &
      &        ,XTIME=grid%xtime                                                &
               ,CURR_SECS=curr_secs, ADAPT_STEP_FLAG=adapt_step_flag       &
+#if ( EM_CORE == 1)
      &        ,mut=grid%mut,dnw=grid%dnw                                  & ! FARMS coupling
      &        ,swdown2=grid%swdown2, swddni2=grid%swddni2                 & ! FARMS coupling
      &        ,swddif2=grid%swddif2, swddir2=grid%swddir2                 & ! FARMS coupling
      &        ,swdownc2=grid%swdownc2, swddnic2=grid%swddnic2             & ! FARMS coupling
+#endif
```
weiwangncar: @davegill Why is this needed since this is only called from EM_CORE?

PR #1094, opened 2020-02-12
Fixing extra biomass burning mass aloft

kayeekayee:
TYPE: enhancement

KEYWORDS: biomass burning, plume rise

SOURCE: Ravan Ahmadov (NOAA/ESRL)

DESCRIPTION OF CHANGES: 
This bug results in adding extra biomass burning mass aloft, when the thickness of the 
vertical grid (dz) increases with altitude. The fix does not change the WRF-Chem model 
results significantly.

LIST OF MODIFIED FILES:
M chem/module_chem_plumerise_scalar.F

TESTS CONDUCTED: 
1. Jenkins test has passed.

### DIFF
--- a/chem/module_chem_plumerise_scalar.F
+++ b/chem/module_chem_plumerise_scalar.F
@@ -154,7 +154,7 @@ subroutine plumerise(m1,m2,m3,ia,iz,ja,jz,firesize,mean_fct   &
 	   call set_flam_vert(ztopmax,k1,k2,nkp,zzcon,W_VMD,VMD)
 	   
 	   !- espessura da camada vertical 
-	   dz_flam=zzcon(k2)-zzcon(k1-1)
+	   dz_flam=zzcon(k2+1)-zzcon(k1)   ! fixed version
 	   
 	   !- distribui a emissao flaming entre os niveis k1 e k2	 
 !          print *,'distribui, k1,k2,dz_flam',k1,k2,dz_flam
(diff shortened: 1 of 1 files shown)


weiwangncar:
@kayeekayee This PR should trigger an automatic regression test. Within 30 min or so of making this PR, this regression test should complete, and report should be sent to you with sender address jenkins@no-reply.scalacomputing.com. Look for your junk folder if you don't see it in your inbox. Let us know the test results. 

jenkins@no-reply.scalacomputing.com
--

ravanah:
We informed other WRF-Chem developers about this bug. They agreed that it should be fixed.

kayeekayee:
@weiwangncar I haven't received any email from jenkins@no-reply.scalacomputing.com, including my spam folder. What should I do in this case?

weiwangncar:
@kayeekayee I just made a change in your modified file, which will trigger the test and test results should come back to me. I will let you know.

kayeekayee:
@weiwangncar Thank you for stepping in, Wei.

weiwangncar:
@kayeekayee I may have to take that back... It looks like the automated test is not active at this time. No sure why. I also noted that you made your change based on master branch. It needs to be changed to base on develop branch. Can you do that? This can be found in the middle of the line just under the title of this PR - right now it says 'wrf-model:master. Click on it and it will have a pull-down menu and then select 'develop' from it.

kayeekayee:
@weiwangncar I need to close this PR and open a new one for the develop branch, correct?

dudhia:
I think you can just change the target on the PR directly by clicking on
the "into" link.

kayeekayee:
Thank you @weiwangncar & @dudhia. I am able to change it by pressing the "edit" button next to the title and then change it to "develop". Thanks again!

davegill:
@kayeekayee @ravanah 
> The fix does not change the WRF-Chem model results significantly.

Kayee,
Could you provide a simple "before vs after" figure to give a more objective measure of the impact?

PR #1095, opened 2020-02-14
Separating new_diagnostics from the rest in diag_misc module

weiwangncar:
TYPE: enhancement

KEYWORDS: diagnostics, misc, nwp diagnostics

SOURCE: internal

DESCRIPTION OF CHANGES: 
The module_diag_misc.F is getting long and complicated and it makes adding additional 
diagnostics (such as accumulated tendencies) difficult. This PR pulls the code related to 
new_diagnostics out to a new module: module_diag_nwp.F, and keeps the rest (currently only 
three items: bucket calculations for long simulations, interval rainfall, and surface pressure 
changes) in the original module: module_diag_misc.F.

M       main/depend.common
M       phys/Makefile
M       phys/module_diag_misc.F
A       phys/module_diag_nwp.F
M       phys/module_diagnostics_driver.F

TESTS CONDUCTED: 
 - [x] Passed Jenkins test
 - [x] Tested the code before and after the change for a 3 hour run with nwp_diagnostics turned on, and prec_acc_dt set to 60.
1. The output in wrfout files (which include PREC_ACC_C, PREC_ACC_NC, UP_HELI_MAX, 
WSPD10MAX, W_MEAN, REFD_MAX and others) are bit-for-bit identical. 
2. The option diag_print is also turned on which generates some prints (such as domain-averaged dpsdt, dmudt) in rsl.out.0000, and the output from both runs are also identical.

### DIFF
--- a/main/depend.common
+++ b/main/depend.common
@@ -760,6 +760,7 @@ module_sf_ocean_driver.o : \
 module_diagnostics_driver.o: \
 		module_lightning_driver.o 		\
 		module_diag_misc.o 			\
+		module_diag_nwp.o 			\
 		module_diag_cl.o			\
 		module_diag_pld.o 			\
 		module_diag_zld.o 			\
--- a/phys/Makefile
+++ b/phys/Makefile
@@ -207,6 +207,7 @@ DIAGNOSTIC_MODULES_EM = \
 	module_diag_functions.o \
         module_diag_hailcast.o \
 	module_diag_misc.o \
+	module_diag_nwp.o \
         module_diag_rasm.o \
 	module_diag_pld.o \
 	module_diag_zld.o \
--- a/phys/module_diag_misc.F
+++ b/phys/module_diag_misc.F
@@ -9,8 +9,6 @@ END MODULE module_diag_misc
 !
 
 MODULE module_diag_misc
-      PRIVATE :: WGAMMA
-      PRIVATE :: GAMMLN
 CONTAINS
    SUBROUTINE diagnostic_output_calc(                                 &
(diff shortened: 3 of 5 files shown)


kkeene44:
@weiwangncar @davegill @smileMchen 
Looks okay to me!

davegill:
@weiwangncar 
For the initial output, do you want to put a call to this new routine in start_em.F? It stops the non-zero values.

Here is the location of the pressure level diag call in start_em:
```
   1955    IF (config_flags%p_lev_diags == PRESS_DIAGS ) THEN
   1956     CALL wrf_debug ( 200 , ' PLD: pressure level diags' )
   1957     CALL pld (                                  &
```

Here is the location of the height level diag call in start_em:
```
   2002    IF (config_flags%z_lev_diags == Z_DIAGS ) THEN
   2003      CALL wrf_debug ( 200 , ' ZLD: height level and AGL diags' )
   2004      CALL zld (                                  &
```

weiwangncar:
@davegill Most variables computed in these two modules are diagnostics from model, and they don't have values at 0 h. For example, variables such as prec_acc_*, dpsdt, and so on in the remaining of module_diag_misc.F, and wspd10max, up_heli_max etc in the new module_diag_nwp.F.

davegill:
@weiwangncar 

>Most variables computed in these two modules are diagnostics from model, and they don't have values at 0 h. 

Wei, 
OK, I am fine with that issue.

Please complete the PR message (maybe show the same diagnostic is the same value before vs after) and I'll be OK with this PR.

weiwangncar:
@davegill PR message is updated.

PR #1096, opened 2020-02-19
Just a Jenkins test on develop

davegill:
## TEST

If this is OK, then the release 414 is OK too

modified:   dyn_em/solve_em.F


davegill:
Yep, all is right with the world

PR #1100, opened 2020-02-21
Fix feedback for odd nest ratios > 3:1

davegill:
TYPE: bug fix

KEYWORDS: nest feedback

SOURCE: Reported by Kevin Manning (NCAR), fixed internally

DESCRIPTION OF CHANGES:
Problem:
All nonmasked, odd-ratio (with ratios greater than 3:1) feedback was incorrect. The fine
grid feedback was incorrectly shifted down and to the left.

Solution:
A fixed constant value of 1 was correct for a 3:1 ratio since the term was
supposed to be ratio/2. For a 5:1 ratio the value needed to be 2. Replacing
the fixed value with the correct term fixes the problem. This solution works
for all odd ratios.

LIST OF MODIFIED FILES:
modified:   share/interp_fcn.F

TESTS CONDUCTED:
The test involves generating a known block of values on the fine grid that
should all feedback to their respective parent grid cells. By choosing the
fine grid cells so that only full parent cells are selected, and by filling
in the fine grid with a constant value (in our case: 1.0), then the coarse
grid should visually reflect the fine grid after feedback by appearing as a
square with sharp edges, rather than a smeared image.

1. Example of original code with wrong indexing. This is wrong:
Screen Shot 2020-02-21 at 10 56 19 AM

2. Example of corrected code with hard-coded assignment of 25 fine-grid points. This
is correct:
Screen Shot 2020-02-21 at 11 50 58 AM

3. Example with fixed but now standard code. This is correct:
Screen Shot 2020-02-21 at 2 03 16 PM

4. Even ratios of 2:1 and 4:1 are correct

RELEASE NOTE: ALL of the odd nest ratios (excluding 3:1) had an incorrect feedback. The indexing was off by (ratio/2 -1). For example, a 5:1 ratio would be off by a single fine grid point. Even ratio feedback was not impacted.

### DIFF
--- a/share/interp_fcn.F
+++ b/share/interp_fcn.F
@@ -1465,11 +1465,11 @@ END MODULE module_interp_info
 
         IF      ( ( .NOT. xstag ) .AND. ( .NOT. ystag ) ) THEN
            DO cj = MAX(jpos+spec_zone,cjts),MIN(jpos+(njde-njds)/nrj-jstag-spec_zone,cjte)
-              nj = (cj-jpos)*nrj + jstag + 1
+              nj = (cj-jpos)*nrj + nrj/2 + 1
               DO ck = ckts, ckte
                  nk = ck
                  DO ci = MAX(ipos+spec_zone,cits),MIN(ipos+(nide-nids)/nri-istag-spec_zone,cite)
-                    ni = (ci-ipos)*nri + istag + 1
+                    ni = (ci-ipos)*nri + nri/2 + 1
                     cfld( ci, ck, cj ) = 0.
                     DO ijpoints = 1 , nri * nrj
                        ipoints = MOD((ijpoints-1),nri) + 1 - nri/2 - 1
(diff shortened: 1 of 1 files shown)


davegill:
@jamiebresch @kwman44 
Jamie and Kevin,
This should fix the odd ratio errors that you were seeing, lo those many months ago.

This is a bug fix, and will get into v4.2 without any troubles. There is no real time pressure on this PR. If you do want to conduct a test, within the next few weeks would be fine to post comments here.

dudhia:
Interesting. I have seen user problems with feedback, probably a 5:1 ratio, where waves show up at the domain boundary sometimes at LES resolutions. I wonder if this is related. They got past it by turning off feedback.

weiwangncar:
@davegill So it will fix feedback for all other larger odd ratios? Perhaps it can be stated as so.

davegill:
> I have seen user problems with feedback, probably a 5:1 ratio, where waves show up at the domain boundary sometimes at LES resolutions. I wonder if this is related. They got past it by turning off feedback.

Jimy,
This could certainly be the trouble.

Ming and Kelly,
If anyone has mentioned nest ratios (4:1, 5:1) troubles on the forum, this might be a PR to consider for them to test.

rsarthur:
I am experiencing potential processor communication/halo issues related to this PR for a (admittedly large) nest ratio of 9:1. I have not fully dug into this but the problem appears to go away when I run in serial. I'm wondering if anyone has tested this fix for nest ratios larger than 5:1?

PR #1101, opened 2020-02-24
Check if STEPWTD==0 in NoahMP before mod(timestep,STEPWTD)

davegill:
TYPE: bug fix

KEYWORDS: NoahMP, STEPWTD, mod

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
There were two places in the WRF code that tested for mod(itimestep,STEPWTD).
If STEPWTD==0, then this is an error.

Solution:
The correct test order is defined as
1. Is this opt_run == 5? If true, then the STEPWTD has been previously defined, and can be used.
2. Is this the correct time step?

LIST OF MODIFIED FILES:
modified:   Registry/Registry.NMM
modified:   phys/module_surface_driver.F

TESTS CONDUCTED:
1. All tests with NoahMP fail immediately with STEPWTD==0 due to a divide by 0.
2. All tests pass with this fix in place.
3. Regression test underway.

### DIFF
--- a/dyn_em/module_first_rk_step_part1.F
+++ b/dyn_em/module_first_rk_step_part1.F
@@ -449,7 +449,7 @@ BENCH_START(surf_driver_tim)
 
 !gmm halo of wtd and riverflow for leafhydro
 #ifdef DM_PARALLEL
-  IF ( config_flags%sf_surface_physics.eq.NOAHMPSCHEME ) THEN
+  IF ( ( config_flags%sf_surface_physics.eq.NOAHMPSCHEME ) .and. ( grid%STEPWTD .ne. 0 ) )THEN
        IF ( config_flags%opt_run.eq.5.and.mod(grid%itimestep,grid%STEPWTD).eq.0 )  THEN
 #     include "HALO_EM_HYDRO_NOAHMP.inc"
        ENDIF
--- a/dyn_em/module_first_rk_step_part1.F
+++ b/dyn_em/module_first_rk_step_part1.F
@@ -449,8 +449,8 @@ BENCH_START(surf_driver_tim)
 
 !gmm halo of wtd and riverflow for leafhydro
 #ifdef DM_PARALLEL
-  IF ( ( config_flags%sf_surface_physics.eq.NOAHMPSCHEME ) .and. ( grid%STEPWTD .ne. 0 ) )THEN
-       IF ( config_flags%opt_run.eq.5.and.mod(grid%itimestep,grid%STEPWTD).eq.0 )  THEN
+  IF ( ( config_flags%sf_surface_physics.eq.NOAHMPSCHEME ) .and. ( config_flags%opt_run.eq.5 ) ) THEN
+       IF ( mod(grid%itimestep,grid%STEPWTD).eq.0 )  THEN
 #     include "HALO_EM_HYDRO_NOAHMP.inc"
        ENDIF
   ENDIF
--- a/phys/module_surface_driver.F
+++ b/phys/module_surface_driver.F
@@ -3101,6 +3101,7 @@ CONTAINS
 	 ENDIF 
 
   if(iopt_run.eq.5)then
+     if (mod(itimestep,STEPWTD).eq.0)then
            CALL wrf_debug( 100, 'calling WTABLE' )
 
 !gmm update wtable from lateral flow and shed water to rivers
@@ -3115,6 +3116,7 @@ CONTAINS
                                   ims,ime, jms,jme, kms,kme,                             &
                                   i_start(ij),i_end(ij), j_start(ij),j_end(ij), kts,kte )
 
+     endif
   endif
 
          call seaice_noah( SEAICE_ALBEDO_OPT, SEAICE_ALBEDO_DEFAULT, SEAICE_THICKNESS_OPT, &
--- a/phys/module_surface_driver.F
+++ b/phys/module_surface_driver.F
@@ -3100,8 +3100,8 @@ CONTAINS
 	 
 	 ENDIF 
 
-  if(iopt_run.eq.5)then
-     if (mod(itimestep,STEPWTD).eq.0)then ! STEPWTD always and only defined for iopt_run == 5
+  IF ( iopt_run .EQ. 5 ) THEN
+     IF ( MOD(itimestep,STEPWTD) .EQ. 0 ) THEN ! STEPWTD always and only non-zero for iopt_run == 5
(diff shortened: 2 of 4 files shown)


davegill:
@barlage @dudhia @weiwangncar 
Mike,
Do these safety "IF" tests seem OK?

dyn_em/module_first_rk_step_part1.F
```diff
@@ -449,7 +449,7 @@ SUBROUTINE first_rk_step_part1 (   grid , config_flags              &
 
 !gmm halo of wtd and riverflow for leafhydro
 #ifdef DM_PARALLEL
-  IF ( config_flags%sf_surface_physics.eq.NOAHMPSCHEME ) THEN
+  IF ( ( config_flags%sf_surface_physics.eq.NOAHMPSCHEME ) .and. ( grid%STEPWTD .ne. 0 ) )THEN
```
barlage: @davegill could we instead check (e.g., in check_a_mundo) that STEPWTD > 0?

barlage: I guess check_a_mundo is not the right place either since this is not coming through the namelist, but is calculated in init

davegill: Mike,
This is a bug fix, so we have time to get this right for v4.2. If I just turn sf_surface_physics from 2 to 4, with no noahmp namelist record, I get this error. Is there a default value for stepwtd that is better than zero?

barlage: Dave - 1 is a better default value since this is how often to run groundwater [in timestep units], it should not be <= 0. There is a line in noahmp_init:
STEPWTD = max(STEPWTD,1)
but when groundwater is not active this line is skipped.

barlage: is this a recent compiler change? I had this happen in some other code recently, i.e., it seems to check all the conditionals instead of checking sequentially.

barlage: I also don't know why these are implemented as OPTIONAL since they are not packaged separately, but maybe that's an NMM thing.

davegill: Mike,
I was testing with gnu 9

davegill: Mike,
Here's the code:
```
       if(iopt_run.eq.5) then
          IF ( PRESENT(smoiseq)     .AND. &
            PRESENT(smcwtdxy)    .AND. &
            PRESENT(rechxy)      .AND. &
            PRESENT(deeprechxy)  .AND. &
            PRESENT(areaxy)      .AND. &
            PRESENT(dx)          .AND. &
            PRESENT(dy)          .AND. &
            PRESENT(msftx)       .AND. &
            PRESENT(msfty)       .AND. &
            PRESENT(wtddt)       .AND. &
            PRESENT(stepwtd)     .AND. &
            PRESENT(dt)          .AND. &
            PRESENT(qrfsxy)      .AND. &
            PRESENT(qspringsxy)  .AND. &
            PRESENT(qslatxy)     .AND. &
            PRESENT(fdepthxy)    .AND. &
            PRESENT(ht)          .AND. &
            PRESENT(riverbedxy)  .AND. &
            PRESENT(eqzwt)       .AND. &
            PRESENT(rivercondxy) .AND. &
            PRESENT(pexpxy)      .AND. &
            PRESENT(rechclim)    ) THEN

             STEPWTD = nint(WTDDT*60./DT)
             STEPWTD = max(STEPWTD,1)

          ELSE
             CALL wrf_error_fatal ('Not enough fields to use groundwater option in Noah-MP')
          END IF
       endif
```

I would have expected stepwtd to be 1 after this. I'll run this down.

barlage: I think the problem is that this block isn't being called when iopt_run /= 5 so STEPWTD is getting the compiler default initialization, which might be 0.

davegill: Mike,
Yes, STEPWTD is a state integer, so initialized to zero.

So the logic would be 
```
if iopt_run /= 5 then
   stepwtd = 1
```
?

barlage: I think what I've done recently in this case is change:
`IF ( config_flags%opt_run.eq.5.and.mod(grid%itimestep,grid%STEPWTD).eq.0 )  THEN`
to:
```
IF ( config_flags%opt_run.eq.5) THEN
  IF (mod(grid%itimestep,grid%STEPWTD).eq.0 )  THEN
```

barlage: I like your logic, but is there anything special with the OPTIONAL status. I don't think these are actually optional but are declared so.

davegill: Mike,
Likely OPTIONAL for some NMM reason.

I would like to get this fixed soon so that we can include NoahMP in our testing. What is your preference?

barlage: To not deal with the optional issue, I prefer this:
```
IF ( config_flags%opt_run.eq.5) THEN
  IF (mod(grid%itimestep,grid%STEPWTD).eq.0 )  THEN
```

phys/module_surface_driver.F
```diff
@@ -3100,7 +3100,8 @@ SUBROUTINE surface_driver(                                         &
 	 
 	 ENDIF 
 
-  if(iopt_run.eq.5.and.mod(itimestep,STEPWTD).eq.0)then
+  if(STEPWTD.ne.0)then
+     if(iopt_run.eq.5.and.mod(itimestep,STEPWTD).eq.0)then
```
barlage: I prefer this:
```
    if(iopt_run.eq.5) then
      if(mod(itimestep,STEPWTD).eq.0) then

```

phys/module_surface_driver.F
```diff
@@ -3100,7 +3100,8 @@ SUBROUTINE surface_driver(                                         &
 	 
 	 ENDIF 
 
-  if(iopt_run.eq.5.and.mod(itimestep,STEPWTD).eq.0)then
+  if(STEPWTD.ne.0)then
```
barlage: if below is used, then remove this line.

dyn_em/module_first_rk_step_part1.F
barlage: if this is used, then remove `.and. ( grid%STEPWTD .ne. 0 ) ` above

barlage:
Looks good, @davegill

PR #1104, opened 2020-02-25
Updates to MYNN-EDMF and related modules

joeolson42:
TYPE: bug fix, enhancement

KEYWORDS: subgrid cloud ice, shallow-cumulus clouds

SOURCE: Joseph Olson (NOAA) and Jaymes Kenyon (NOAA/CIRES)

DESCRIPTION OF CHANGES: 
Added new array qi_bl as opposed to using qc_bl for both subgrid-scale (SGS) qc and qi. This
gives us more control of the magnitudes of SGS qc & qi, which can be confounded by using
a single array. As a result of this change, many subsequent changes were required,
especially in subroutine mym_condensation, which was cleaned up in the process.

Added the blending of the stratus component of the SGS clouds to the mass-flux
clouds to account for situations where stratus and cumulus may exist in the grid cell.
This has a small impact, but can slightly increase the tenuous shallow-cu cloud decks.

Miscellaneous tweaks and small-impact bug fixes:
1) dz was incorrectly indexed in mym_condensation
2) configurations with icloud_bl = 0 were using uninitialized arrays
3) cloud water detrainment was reduced - was a bit excessive

LIST OF MODIFIED FILES: 
M	Registry/Registry.EM_COMMON
M	dyn_em/module_first_rk_step_part1.F
M	phys/module_bl_mynn.F
M	phys/module_pbl_driver.F
M	phys/module_radiation_driver.F

TESTS CONDUCTED: 
- Case studies performed. 
- Retrospective period tests in progress. 
- NMM/HWRF compiles.

### DIFF
--- a/Registry/Registry.EM_COMMON
+++ b/Registry/Registry.EM_COMMON
@@ -1536,8 +1536,9 @@ state    real  CCN4_GS          ikj      misc        1         -      rh      "C
 state    real  CCN5_GS          ikj      misc        1         -      rh      "CCN5_GS"             "Grid Scale Cloud Condensation Nuclei at S=0.3%"       "#/cm-3"
 state    real  CCN6_GS          ikj      misc        1         -      rh      "CCN6_GS"             "Grid Scale Cloud Condensation Nuclei at S=0.5%"       "#/cm-3"
 state    real  CCN7_GS          ikj      misc        1         -      rh      "CCN7_GS"             "Grid Scale Cloud Condensation Nuclei at S=1.0%"       "#/cm-3"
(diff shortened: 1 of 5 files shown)


davegill:
@weiwangncar @dudhia @joeolson42 
Folks,
Let's verify that this is OK for NMM. There is a mod in Registry.EM_COMMON, and a mod to drivers (either of which is a bit of a red flag). Those variables _are_ declared optional, but that "not really there" variable still gets passed down to subsequent routines.

weiwangncar:
@davegill @joeolson42 Just compile NMM on my desktop, it is ok.

PR #1118, opened 2020-03-04
Fix nudging of temperature obs at edge of tile

### LINKED ISSUE #1115: Obs nudging tile issue
When observations are on the edge of a tile there is an error in the determination of the potential temperature difference between the model and the observation for observation nudging.

In subroutine errob in phys/module_fddaobs_rtfdda.F when it calculates the potential temperature difference between the model and the observation (stored in ERRF(3,N)), it interpolates the model value of tb.  The interpolation relies on adjacent points and thus when the ob is on the edge of a tile the interpolation will need points outside of the tile.  However, the points outside of the tile are not set correctly (in the case tested they were set to 0.0).  This results in the difference between the model and the observation being incorrect and thus the model nudges towards the wrong value.

In errob, the model potential temperature is accessed via variable tb, which is listed as:
REAL,   INTENT(IN) :: tb( ims:ime, kms:kme, jms:jme )
In the case tested where the problem occurred the observation was at j=82 and the various j end values are:
jme=93
jde=205
jte=82
Thus, it would appear that the j values needed for the interpolation should be present in tb since j=83 < jme=93.  However all j=83 values of tb used by the interpolation are set to 0.0.

The variable tb in subroutine errob is passed into errob from fddaobs_driver (phys/module_fddaobs_driver.F) from tb.  It is passed into fddaobs_driver by subroutine first_rk_step_part2 (dyn_em/module_first_rk_step_part2.F) from grid%th_phy_m_t0. [In earlier versions (e.g., V3.9.1.1) instead of grid%th_phy_m_t0, grid%t_2 was passed in (I think the change is related to the hybrid vertical coordinate), but I have not tested to see if the problem occurs when grid%t_2 is instead passed.]

The problem is resolved for the case tested by adding th_phy_m_t0 to the following line in Registry/Registry.EM_COMMON:
halo      HALO_OBS_NUDGE dyn_em 24:ph_2,p,uratx,vratx,tratx,kpbl
so that it is instead:
halo      HALO_OBS_NUDGE dyn_em 24:ph_2,p,uratx,vratx,tratx,kpbl,th_phy_m_t0

However, I lack sufficient knowledge of WRF halos to know if this is the proper fix and so am not currently submitting this as a pull request.

The other variables used by subroutine errob to calculate the model vs. obs differences do not have the issue of being set to zero outside the tile, even though they are not included in the halo statement and are dimensioned the same, e.g.:
REAL,   INTENT(IN) :: qvb( ims:ime, kms:kme, jms:jme )

Should all variables dimensioned similarly be included in the halo statement if one is going to access values outside of the tile?  If this is the case perhaps the line in the Registry should be:
halo      HALO_OBS_NUDGE dyn_em 24:ph_2,p,uratx,vratx,tratx,kpbl,u_2,v_2,th_phy_m_t0,moist,pb,p

brianreen:
TYPE: bug fix

KEYWORDS: observation nudging, temperature, tile, halo

SOURCE: Brian Reen (Army Research Lab)

DESCRIPTION OF CHANGES: 
Added th_phy_m_t0 to the HALO statement for observation nudging in the Registry. 

Problem:
Previously, observation nudging using temperature observations on the boundary edge of a 
tile used an incorrect nudging term.  During the interpolation of model potential temperature 
to the observation location, model values outside of the current tile are needed when the 
observation is on the boundary edge of the tile.  However, the model potential temperature 
for these points just outside the current tile was not available in the variable used for this calculation (tb in phys/module_fddaobs_rtfdda.F).  During testing, a value of zero was instead 
seen in these off-tile gridpoints (TB holds dry potential temperature minus 300 K, thus a value 
of tb = zero means that the model potential temperature = 300 K).  

Solution:
Now, the observation nudging HALO statement in the Registry has had th_phy_m_t0 added to 
it so that the values of tb one grid point beyond the tile boundary edge are correct.  This 
allows the first-order linear interpolation of model potential temperature to produce the 
correct value for comparison against the observation.  

The variable th_phy_m_t0 is the added to the HALO statement because subroutine 
first_rk_step_part2 (dyn_em/module_first_rk_step_part2.F) passes grid%th_phy_m_t0 to 
subroutine fddaobs_driver (phys/module_fddaobs_driver.F) which in turn passes it to 
subroutine errob (phys/module_fddaobs_rtfdda.F) where it is tb.  

The variable passed by first_rk_step_part2 for eventual use as tb was originally grid%t_2 in 
v3.9.1.1 and earlier versions of WRF. This bug was introduced during the change from 
grid%t_2 to grid%th_phy_m_t0 in WRFV4.0 (the dry potential temperature to moist potential 
temperature modification).

ISSUE: 
```
Fixes #1115 
```

LIST OF MODIFIED FILES: 

TESTS CONDUCTED: 

1. Inserted in subroutine errob in phys/module_fddaobs_rtfdda.F prints of the values used 
to interpolate the model potential temperature (tb) to the observed value.  For a temperature 
observation on the edge of a tile, previously the values of tb outside of the tile were zero 
whereas now the values of tb outside of the tile are very similar to those inside the tile.
2. Jenkins all PASS

### DIFF
--- a/Registry/Registry.EM_COMMON
+++ b/Registry/Registry.EM_COMMON
@@ -3248,7 +3248,7 @@ halo      HALO_EM_F_1 dyn_em  24:muus,muvs
 #halo      em_shift_halo_x  dyn_em 48:imask_nostag,imask_xstag,imask_ystag,imask_xystag,u_2,v_2,t_2
 
 # For observational nudging
-halo      HALO_OBS_NUDGE dyn_em 24:ph_2,p,uratx,vratx,tratx,kpbl
+halo      HALO_OBS_NUDGE dyn_em 24:ph_2,p,uratx,vratx,tratx,kpbl,th_phy_m_t0
 
 # cyl: For 3DPWP
 halo      HALO_PWP dyn_em 8: OM_TMP,OM_S,OM_U,OM_V,OM_DEPTH,OM_TINI,OM_SINI


davegill:
@brianreen 
Brian,
You should receive an email from "jenkins". Our emails originally went to SPAM, so you might need to look there. Would you please post the text from that email  in one of these comments.

It should look something like this:
```
    Number of Tests        : 10           10
    Number of Builds       : 23           23
    Number of Simulations  : 140          140         12
    Number of Comparisons  : 81          81         0

    Failed Simulations are: 
    output_1:9 = STATUS test_001s em em_real 32 50
output_1:9 = STATUS test_001s em em_real 32 51
output_1:9 = STATUS test_001s em em_real 32 60
output_1:9 = STATUS test_001s em em_real 32 60NE
output_1:9 = STATUS test_001o em em_real 33 50
output_1:9 = STATUS test_001o em em_real 33 51
output_1:9 = STATUS test_001o em em_real 33 60
output_1:9 = STATUS test_001o em em_real 33 60NE
output_1:9 = STATUS test_001m em em_real 34 50
output_1:9 = STATUS test_001m em em_real 34 51
output_1:9 = STATUS test_001m em em_real 34 60
output_1:9 = STATUS test_001m em em_real 34 60NE
    None
```

brianreen:
Here is the text emailed from Jenkins:
    Number of Tests        : 10           10
    Number of Builds       : 23           23
    Number of Simulations  : 140           140        0
    Number of Comparisons  : 89           89        0

    Failed Simulations are: 
    None
    None

davegill:
@weiwangncar @dudhia @brianreen 
Wei and Jimy,
I am OK with this PR. 

1. Generally, you really can't do any harm with an extra field that is to be communicated. 
2. Brian is able to show before vs after that this field is mandatory.
3. This is a very small mod and a legitimate bug-fix
4. The PR is well put together and references an outstanding issue

PR #1128, opened 2020-03-09
Change default value of fractional_seaice from 0 to 1

weiwangncar:
### openmp b4b broken with 07NE, 20, 20NE, 60, 71, 78, tropical

TYPE: enhancement

KEYWORDS: namelist fractional_seaice, default

SOURCE: Internal

DESCRIPTION OF CHANGES: 
Since GFS, the most commonly used input data source, has been providing fractional seaice 
data for a while, it seems reasonable to make the default option for fractional_seaice to be 1. 
This should also help our general users. Both ERA5 and ERA-Interim provide fractional seaice 
data, too.

LIST OF MODIFIED FILES: 

TESTS CONDUCTED: 
Reg test:

RELEASE NOTE: The default option for using fractional_seaice has been changed from 0 (off) to 1 (on) starting with release v4.2.


davegill:
@weiwangncar 
Wei,
Changing this default causes some troubles. 
```
FATAL CALLED FROM FILE:  <stdin>  LINE:    2317
SLAB scheme cannot be used with fractional seaice
```
Please add a check_a_mundo test that turns off the fractional seaice if the LSM is the slab model.

davegill:
@weiwangncar 
Wei,
This mod breaks the regression test, as there are a number of sf_surface_physics=1 options for the em_real cases. I need to put in the fractional_seaice = 0 for all of those. Until then, with this mod, there will be about 30 FAILs.

davegill:
@weiwangncar 
Wei,
It also seems that PX does not always work well with fractional_seaice. Maybe due to the moisture nudging?

davegill:
Replaced by: Change default fractional_seaice option from 0 to 1 (#1401)

PR #1129, opened 2020-03-09
bug fix: NoahMP STEPWTD==0, divide by zero

smileMchen:
TYPE: bug fix

KEYWORDS: NoahMP, opt_run option

SOURCE: internal 

DESCRIPTION OF CHANGES: 
Problem:
When running the NoahMP scheme with the ground water option (opt_run=5), the STEPWTD 
variable is initialized to a non-zero value. However when opt_run!=5, then the STEPWTD 
variable remains an uninitialized zero value. In the WRF source code, there are a couple of
places where mod(x,STEPWTD) occurs. This causes an immediate "divide by zero" error in
the first call to the surface driver.

Solution:
The IF test logic is re-arranged so that only when opt_run=5 (which means that STEPWTD
is non-zero) is there a second IF test (using mod(x,STEPWTD)) to compute whether 
additional processing takes place.

This modification was made separately to the develop branch after the branch release-v4.1.4 
was released. See https://github.com/wrf-model/WRF/pull/1101. The purpose of this modification 
is to allow users access to the latest v4.1.* branch that works with NoahMP.

As this will be a single issue release, the README and inc/version_decl are also modified 
in preparation for the release of WRF release-v4.1.5 with this PR.

LIST OF MODIFIED FILES: 
M     dyn_em/module_first_rk_step_part1.F
M     phys/module_surface_driver.F
M     README
M     inc/version_decl

TESTS CONDUCTED: 
1. Jenkins test PASS (positively tests NoahMP)
2. Regression test in NCAR-MMM classroom machine PASS.

RELEASE NOTES: Newer versions of the GNU compiler more tightly conform to the Fortran standard and do not immediately short circuit IF test processing once a statement is known to be FALSE. This newer and legal behavior recently caused troubles with NoahMP regarding the frequency to process ground water. This release fixes that trouble with modified logic, but the still existing problem could extend (depending on the compiler) back to version 3.6.

### DIFF
--- a/README
+++ b/README
@@ -1,4 +1,4 @@
-WRF Model Version 4.1.4
+WRF Model Version 4.1.5
 
 http://www2.mmm.ucar.edu/wrf/users/
 
--- a/dyn_em/module_first_rk_step_part1.F
+++ b/dyn_em/module_first_rk_step_part1.F
@@ -437,8 +437,8 @@ BENCH_START(surf_driver_tim)
 
 !gmm halo of wtd and riverflow for leafhydro
 #ifdef DM_PARALLEL
-  IF ( ( config_flags%sf_surface_physics.eq.NOAHMPSCHEME ) .and. ( config_flags%opt_run.eq.5 ) ) THEN
-       IF ( mod(grid%itimestep,grid%STEPWTD).eq.0 )  THEN
+  IF ( ( config_flags%sf_surface_physics.EQ.NOAHMPSCHEME ) .and. ( config_flags%opt_run.EQ.5 ) ) THEN
+       IF ( mod(grid%itimestep,grid%STEPWTD).EQ.0 )  THEN
 #     include "HALO_EM_HYDRO_NOAHMP.inc"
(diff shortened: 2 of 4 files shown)


weiwangncar:
The couple of lines in module_first_rt_step_part1.F first appeared in the code in v3.6. So potentially all of the versions since then could fail if a new compiler is used. If this is true, we should add a note for ourselves.

davegill:
@smileMchen 
MIng,
Here is the jenkins test:
```
Please find result of the WRF regression test cases in the attachment. This build is for Commit ID: 014ff12ee33d756e8e611fce50bf2bf7677385a0, requested by: smileMchen for PR: https://github.com/wrf-model/WRF/pull/1129. For any query please send e-mail to David Gill.

    Number of Tests        : 10           10
    Number of Builds       : 23           23
    Number of Simulations  : 140           140        0
    Number of Comparisons  : 89           89        0

    Failed Simulations are: 
    None
    None
```

PR #1131, opened 2020-03-12
Allow DA and FSBM to peacefully coexist in same repo

davegill:
TYPE: bug fix

KEYWORDS: DA FSBM real*8

SOURCE: internal

DESCRIPTION OF CHANGES:
The Fast Spectral Bin Microphysics (FSBM) code only works with default real
size = 4 bytes due to a mixture of declarations using real, real(kind=8),
and double precision. To solve this problem, the FSBM code is ifdef'ed out
for any real(kind=8) build. This is handled via the configure step.

Problem:
The DA code is built with real(kind=8) by default without resorting to any special
setup in the configure file. The choice was to either disable FSBM, or not
build DA. Temporarily before the v4.2 release, we have disabled the FSBM code.

Solution:
This PR identifies the DA builds, and sets internal flags to let the build
know to ifdef out the FSBM code. If no DA code is requested AND if this is
not a real(kind=8) build, then the FSBM code is part of the built executable.

LIST OF MODIFIED FILES:
M       arch/postamble
M       configure

TESTS CONDUCTED:
1. Jenkins tests real(kind=4) and real(kind=8) builds - PASS
2. WRFPlus, 3DVAR, 4DVAR compiled successfully with this branch.

### DIFF
--- a/arch/postamble
+++ b/arch/postamble
@@ -28,7 +28,7 @@ ARCHFLAGS       =    $(COREDEFS) -DIWORDSIZE=$(IWORDSIZE) -DDWORDSIZE=$(DWORDSIZ
                       -DLIMIT_ARGS \
                       -DBUILD_RRTMG_FAST=0 \
                       -DBUILD_RRTMK=0 \
-                      -DBUILD_SBM_FAST=0 \
+                      -DBUILD_SBM_FAST=1 \
                       -DSHOW_ALL_VARS_USED=0 \
                       -DCONFIG_BUF_LEN=$(CONFIG_BUF_LEN) \
                       -DMAX_DOMAINS_F=$(MAX_DOMAINS) \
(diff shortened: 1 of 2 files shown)


davegill:
@liujake @jjguerrette @weiwangncar 
Jake and JJ,
I have a simple PR put together for this DA vs FSBM battle. If this PR passes the regression test (later tonight), I'll send you a note. After receiving my thumbs up, let's see if this modification to the configure script is sufficient to allow the DA options to build as usual.

davegill:
@liujake @jjguerrette 
Jake and JJ,
Would you see if this works for the various DA builds? I got enough PASSes to say this PR is OK for the forward nonlinear model.

liujake:
WRFPlus, 3DVAR, 4DVAR are compiled successfully with this branch.

davegill:
@liujake 
> WRFPlus, 3DVAR, 4DVAR are compiled successfully with this branch.

Jake,
Thanks for the testing.

PR #1146, opened 2020-03-24
Syntax fixes from HPE

huangwei0123:
TYPE: bug fix

KEYWORDS: double minus sign

SOURCE: Huang Wei (HPE)

DESCRIPTION OF CHANGES: 

1. There were two minus signs in a row. Not intended.
2. Vertical indexes should be defined as integer, not real (BBOT and TTOP).

LIST OF MODIFIED FILES: 
modify    phys/module_sf_mynn.F
modify    phys/module_cu_bmj.F

TESTS CONDUCTED: 
1. Regression test status is OK.
2. Modified routine with double negative is not used.
3. Routine with integer indexing gives bit-wise identical results (before vs after).


dudhia:
For BMJ, I wonder if this is part of the new changes Wei Wang just added. Need to check if they intended these to be integer or real.
For MYNN, does a double - - give a positive. This would be serious if so.

davegill:
@weiwangncar @dudhia @huangwei0123 @joeolson42 
Folks,
The BMJ CU mod is to fix indexing that should be an integer.
The MYNN SF mod is due to double minus signs.

These were found by the Cray compiler.

weiwangncar:
@huangwei0123 @dudhia @davegill Thanks for the correction. Yes, BBOT and TTOP are new variables added in PR#921 and they are used for loop indices.. I just made two runs with and without this change, the BMJ with bmj_rad_feedback set to true produce identical results. Hence I'm good with this change.

joeolson42:
Technically, the SUBROUTINE PSI_CB2005 isn't being called any more now that the tables are being made by different functions (psim_stable_full & psih_stable_full), so this won't change any results.

davegill:
@weiwangncar @joeolson42 
Wei and Joe,
Thanks for chiming in so quickly. You comments are now part of the PR.

dudhia:
Is a double negative treated as a positive?

davegill:
@joeolson42 @dudhia @weiwangncar 
Folks,
Here's some source code:
```
mmm-docker:/root>cat dub_neg.f90 
      program dub_neg
      real :: a, b, c
      integer :: i, j, k

      a = -1.
      b = 2.

      i = -1
      j = 2

      print *,'a = ',a
      print *,'b = ',b
      c = -a - (-b)
      print *,'c = -a - -b = ',c

      print *,'i = ',i
      print *,'j = ',j
      k = -i - (-j)
      print *,'k = -i - -j',k

      end program dub_neg
```

I needed to wrap the unary operators in parens, otherwise it would not compile. 

Here's the output with GNU
```
mmm-docker:/root>a.out
 a =   -1.00000000    
 b =    2.00000000    
 c = -a - -b =    3.00000000    
 i =           -1
 j =            2
 k = -i - -j           3
```

davegill:
Jenkins is OK:

```
Please find result of the WRF regression test cases in the attachment. This build is for Commit ID: 84062e10680a18c0d175ddd8e3ee3e293701bffd, requested by: huangwei0123 for PR: https://github.com/wrf-model/WRF/pull/1146. For any query please send e-mail to David Gill.

    Number of Tests        : 16           10
    Number of Builds       : 41           23
    Number of Simulations  : 155           65        0
    Number of Comparisons  : 99           39        0

    Failed Simulations are: 
    None
    None
```

PR #1155, opened 2020-04-02
FSBM: don't communicate global PARAMETER values to other MPI processes

davegill:
TYPE: bug fix

KEYWORDS: bcast, parameter, fsbm

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
With GNU compilers 5, 8, and 9: FSBM runs OK serially. FSBM fails with
during intialization with MPI. It is traced to bcast of some variables that are
defined in module_mp_SBM_polar_radar.F.

INTEGER PARAMETER variables:
   * temps_water, temps_fd, temps_crystals, temps_snow, temps_graupel, temps_hail

REAL PARAMETER variables:
   * fws_fd, fws_crystals, fws_snow, fws_graupel, fws_hail

These variables are above the CONTAINS statement in 
module_mp_SBM_polar_radar.F and are PARAMETER values.

Solution:
We are not supposed to communicate PARAMETER values. These are already global
and cannot be written to. Just remove these unnecessary and illegal bcast statements.

LIST OF MODIFIED FILES:
m   module_mp_fast_sbm.F

TESTS CONDUCTED:
1. jenkins all pass
2. Without mods: seg fault 
3. With mods: code runs to completion. Test domain is:
Screen Shot 2020-04-06 at 10 52 27 AM

### DIFF
--- a/phys/module_mp_fast_sbm.F
+++ b/phys/module_mp_fast_sbm.F
@@ -6071,19 +6071,6 @@ enddo
    	DM_BCAST_MACRO_R16 ( FBF5 )
    	DM_BCAST_MACRO_R16 ( FAB5 )
    	DM_BCAST_MACRO_R16 ( FBB5 )
- ! ### (KS) - Broadcating Temperature intervals
-   	CALL wrf_dm_bcast_integer ( temps_water , size ( temps_water ) )
-   	CALL wrf_dm_bcast_integer ( temps_fd , size ( temps_fd ) )
-   	CALL wrf_dm_bcast_integer ( temps_crystals , size ( temps_crystals ) )
-   	CALL wrf_dm_bcast_integer ( temps_snow , size ( temps_snow ) )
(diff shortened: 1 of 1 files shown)


davegill:
@FS-WRF-SBM @weiwangncar @dudhia 
Kobby,
Take a look at this PR. It would be best if you clone this fork/branch and run a test. We understand that it takes a while for your high-resolution case, but it should be ready by early next week.

davegill:
Reggie is OK
```
    Number of Tests        : 18           10
    Number of Builds       : 47           23
    Number of Simulations  : 185           65        0
    Number of Comparisons  : 119           39        0

    Failed Simulations are: 
    None
    None
```

davegill:
@FS-WRF-SBM @weiwangncar @dudhia 
Kobby,
Any status on this? All of our GNU MPI jobs fail with an immediate seg fault due to an attempt to communicate parameter statements inside your initialization routine. We need to know if you get the same correct results in your tests WITHOUT those communications. We need to test your code, and we do not have a version of your code to test.

PR #1156, opened 2020-04-03
force b4b errors in jenkins

davegill:
## TEST to see if Jenkins finds bit-for-bit errors

modified:   start_em.F


davegill:
Just for testing a failed comparison

PR #1158, opened 2020-04-07
Improved logic to initialize aerosols in mp = 28

pedro-jm:
TYPE: Enhancement

KEYWORDS: aerosol initialization, real, aerosol aware microphysics (mp = 28)

SOURCE: Pedro A. Jimenez, Tim Juliano and G. Thompson (NCAR/RAL)

DESCRIPTION OF CHANGES: Improve the logic of the aerosol aware initialization. The code uses flags from the met_em files to initialize the aerosols with either a first guess (flag_qn?fa) or a climatology (flag_qn?fa_cl). This is controlled via the use_aero_icbc option that has been changed from logical to integer: 1) no aerosol initialization (aerosol number = 0.0), 2) aerosol climatology, and 3) aerosol first guess. Legacy code in the aerosol initialization has been removed.

ISSUE: The logic of the aerosol aware initialization did not properly handle the use of climatologies or the first guess

LIST OF MODIFIED FILES: 
M       Registry/Registry.NMM
M       Registry/registry.new3d_wif
M       dyn_em/solve_em.F
M       main/real_em.F
M       run/README.namelist
M       share/module_optional_input.F

TESTS CONDUCTED: It has been tested that the initialization of aerosol is correct setting use_aero_icbc to 0, 1, and 2. It has also been tested that if the aerosols are not present in the met_em files real gratefully stops and indicates the reason for stopping.

RELEASE NOTE: Once approved this PR, it will follow a WPS one to set the flags in the metgrid table.  The default values of the use_aero_icbc, num_wif_levels, and wif_input_opt namelist options have been changed so by default the user only needs to set up mp = 28 to run the aerosol aware code.


dudhia:
This appears to remove monthly Thompson aerosols and related code which is not what is wanted.

pedro-jm:
@dudhia 
It may appear but it does not. 
The couple of large blocks of code removed are the legacy code that I referred to. My impression is that that was the way to initialize the aerosols when the GOCART climatology had to be in the same vertical levels of metgrid.

dudhia:
It seems to include removing monthly Registry variables

pedro-jm:
Those are the legacy ones.
The new ones are here in Registry/registry.new3d_wif which is the registry for the aerosol climatology on the GOGART model levels which also includes pressure at the levels.

davegill:
@pedro-jm @dudhia 
Pedro,
With mp_physics=28, I get this error:
```
FATAL CALLED FROM FILE:  <stdin>  LINE:    2081
mp = 28 but there is not aerosol clim
```

dudhia:
OK, I did not realize those were outdated.

pedro-jm:
@davegill  @dudhia 
Yes, this is expected since we did not modify the metgrid table yet. In the Metgrid table the flags flag_qnwfa and flag_qnifa should be replaced by flag_qnwfa_cl and flag_qnifa_cl, respectively. The _cl refers to climatology. They are located in the December aerosol variables.

The other change in the metgrid table is to add a couple more flags. The variables QNWFA and QNIFA should have the flag_qnwfa and flag_qnifa.

This is to have different flags to tell real if we have the clim and the first guess in the met_em files

dudhia:
We need to also make sure of backward compatibility.

weiwangncar:
@pedro-jm It may be good to make a PR for WPS, so we can see what's there. We cannot approve this PR with testing with modified WPS. In fact this may break Dave Jenkins test, if this MP is one of the test there.

davegill:
@mgduda 
Michael,
Take a look at the comments between Jimy, Wei, Pedro, and myself in this PR. 
We might need to put in a PR to the metgrid TBL file to accommodate this.

weiwangncar:
@pedro-jm What is aerosol first-guess? Where does this data come from?

dudhia:
The text still mentions changing a logical to a real with the same name. I think this is now out of date? What was decided for this?

davegill:
@pedro-jm @dudhia @weiwangncar @gthompsnWRF @mgduda 
> The text still mentions changing a logical to a real with the same name. I think this is now out of date? What was decided for this?

Jimy,
This PR is not intended to be part of v4.2. A variant of this PR might be ready for v4.2.1, if we can change the proposed user interface so that it is backwards compatible (the real vs logical namelist entry you mentioned). The specific user-level concern of a constant aerosol distribution (and the user not being made aware of that distribution) was addressed in a separate PR (already in v4.2): #1162 "MP=28: avoid unintended constant aerosol distribution".

I do not completely understand what the requirements are for proper handling of the specialized Thompson aerosols in the WRF model. We have several different vertical interpolations of aerosols inside of real, with some as a single date (these tend to be on the same vertical coordinate as the atm fields) and some as monthly (these tend to not be on the atm vertical coordinate). I believe that both types of aerosol input are required for some subsequent treatment of surface fluxes in the WRF model (I am fuzzy on that). The WPS package also supports multiple ways to get these aerosols into the data stream, via either geogrid (with fixed vertical levels) or metgrid (with the matching vertical coordinate as with the other atm variables).

The static data from geogrid on fixed levels is monthly, and the second set of _today's_ aerosols goes  into metgrid from some intermediate file.

We do not have the logic to cleanly handle aerosols on two different coordinates in the real code right now (the names of the aerosols are different in WPS, so WPS does seems to work OK). Additionally, to positively test this capability requires a revamp of the testing harness input data. These types of logical restructuring and new input data pose difficulties when trying to go back and do historical testing (for example, the reason that this code is currently failing a regression test is due to insufficient input data). 

After the release, we can have a discussion with all of the stakeholders and see what is required.

Registry/Registry.EM_COMMON
```diff
@@ -125,31 +125,7 @@ state    real   qnh_gc         igj      dyn_em      1        Z     i1  "QNH"   "
 state    real   qnwfa_gc       igj      dyn_em      1        Z     i1  "QNWFA"   "water-friendly aerosol  num concentration"   "# kg-1"
 state    real   qnifa_gc       igj      dyn_em      1        Z     i1  "QNIFA"   "water-friendly aerosol  num concentration"   "# kg-1"
 state    real   qnwfa_now      igj      dyn_em      1        Z     -   "QNWFA_NOW"   "num water-friendly aerosol Now"   "# kg-1"
-state    real   qnwfa_jan      igj      dyn_em      1        Z     i1  "QNWFA_JAN"   "num water-friendly aerosol Jan"   "# kg-1"
```
dudhia, line 128: several other Registry files includes these arrays in their package lists and those should be removed too.

davegill:

PR #1161, opened 2020-04-07
Adding a couple of users to the CODEOWNERS file

kkeene44:
TYPE: text only

KEYWORDS: CODEOWNERS, SBM, Thompson, phys

SOURCE: Internal

DESCRIPTION OF CHANGES: Added 2 new users to the physics section of the CODEOWNERS file. The users are specific to the SBM and Thompson files.

LIST OF MODIFIED FILES: 
M   .github/CODEOWNERS

TESTS CONDUCTED: None necessary - text only


davegill:
@kkeene44 
Kelly,
This does not change any code that is compilable or part  of the build. You can merge without reggie results.

PR #1166, opened 2020-04-09
Add a few urban initial fields to restart file

weiwangncar:
TYPE: bug fix

KEYWORDS: urban options, restart

SOURCE: internal

DESCRIPTION OF CHANGES: 
Tests revealed that all three urban options failed restart tests, meaning the model output after restart does not produce bit-for-bit results with none-restart output. This is a problem if a user include URB_PARAM from geogrid in the input. The few fields derived from this multi-dimensional field need to be put in restart file. 

LIST OF MODIFIED FILES: 
M      Registry/Registry.EM_COMMON

TESTS CONDUCTED: 
- Tested on previously failed restart case with single and nested runs. For nested run, both with and without feedback are ok too.
- Jenkins tests have all passed.

RELEASE NOTE: 
Fixed a restart problem with all three urban options by adding a few initial urban fields (building height, building fraction, etc.) to restart file. This is a problem if those urban initial fields are non zero. A user could get these fields from either the NUDAPT44 data directly or have URB_PARAM in the GEOGRID.TBL which uses a pre-processed dataset. Since NUDAPT data is US only, so this fix only affect US domain applications.


dudhia:
Approved.
Not related to this fix but it was i0 before. I am wondering if using external datasets like NUDAPT would provide these as inputs to real in which case they would need to be i01r.

PR #1169, opened 2020-04-15

davegill:
TYPE:  enhancement

SOURCE: Ted Mansell

DESCRIPTION OF CHANGES: 
Replaces "Efficiency fix for nwp_diag reflectivity calculation" #1060

Problem:
With commit 86f7675359234d239, PR #749 "make refl_10cm available if nwp_diagnostics is on", 
solve_em.F was changed so that `diag_flag` was always set to true when `nwp_diagnostics==1`. 
This caused the full 3D radar reflectivity to be calculated on each time step, whereas the diagnostic 
for the max dBZ only needs to compute lowest level (k=1). The full 3D nature of the new computation
makes the cost of this diagnostic between one and two orders of magnitude more expensive than 
necessary. 

Solution:
A variable `ke_diag` is added in solve_em to tell the reflectivity calculation whether to do the full depth 
(for history output) or only the lowest level (for `nwp_diagnostics==1`). This avoids unnecessary 3D
computation of dBZ for `nwp_diagnostics==1`, which is expensive for a number of schemes. The variable
`ke_diag` is passed to the microphysics driver and from there to individual microphysics schemes. 
The NSSL and Thompson schemes have been updated to use this new capability, and other MP
schemes could be done in a very similar manner.

LIST OF MODIFIED FILES: 
 dyn_em/solve_em.F
 phys/module_mp_nssl_2mom.F
 phys/module_mp_thompson.F


davegill:
@weiwangncar @dudhia @gthompsnWRF @MicroTed 

From Greg:

> So does that new PR mean that a decision has been made just to accept Ted's changes that will affect *only* our 2 schemes but not the other 8 that users may wish to run?

> If that statement is true, then I will immediately edit all the MP schemes to fix this situation, because I just don't believe these are good ideas to solve for one or two schemes, then leave the problem for future fixing later.  As a community service, I'll fix it myself if I have to.

> List of other MP schemes affected: 
  WSM5, WSM6, WDM5, WDM6 (4)
  Morrison 2-moment (1)
  Milbrandt-Yau (1)
 Purdue-Lin (1)
 Goddard, GSFC (1)

>Only a couple lines per needed and while I didn't *break* the code, I feel obligated that all persons benefit, not just Ted and I.

davegill:
merged release-v4.2 into v4.1.4, not the best idea

PR #1170, opened 2020-04-16

davegill:
TYPE:  enhancement

SOURCE: Ted Mansell (NSSL)

DESCRIPTION OF CHANGES: 
Replaces "Efficiency fix for nwp_diag reflectivity calculation" #1060

Problem:
With commit 86f7675359234d239, PR #749 "make refl_10cm available if nwp_diagnostics is on", 
solve_em.F was changed so that `diag_flag` was always set to true when `nwp_diagnostics==1`. 
This caused the full 3D radar reflectivity to be calculated on each time step, whereas the diagnostic 
for the max dBZ only needs to compute lowest level (k=1). The full 3D nature of the new computation
makes the cost of this diagnostic between one and two orders of magnitude more expensive than 
necessary. 

Solution:
A variable `ke_diag` is added in solve_em to tell the reflectivity calculation whether to do the full depth 
(for history output) or only the lowest level (for `nwp_diagnostics==1`). This avoids unnecessary 3D
computation of dBZ for `nwp_diagnostics==1`, which is expensive for a number of schemes. The variable
`ke_diag` is passed to the microphysics driver and from there to individual microphysics schemes. 
The NSSL and Thompson schemes have been updated to use this new capability, and other MP
schemes could be done in a very similar manner.

LIST OF MODIFIED FILES: 
 dyn_em/solve_em.F
 phys/module_mp_nssl_2mom.F
 phys/module_mp_thompson.F

### DIFF
--- a/phys/module_mp_nssl_2mom.F
+++ b/phys/module_mp_nssl_2mom.F
@@ -1625,7 +1625,7 @@ SUBROUTINE nssl_2mom_driver(qv, qc, qr, qi, qs, qh, qhl, ccw, crw, cci, csw, chw
 !                              re_liquid, re_graupel, re_hail, re_icesnow,               &
 !                              vtcloud, vtrain, vtsnow, vtgraupel, vthail,               &
                               ipelectmp,                                                &
-                              diagflag,                                                 &
+                              diagflag,ke_diag,                                         &
                               nssl_progn,                                              & ! wrf-chem 
 ! 20130903 acd_mb_washout start
                               rainprod, evapprod,                                       & ! wrf-chem 
@@ -1694,7 +1694,7 @@ SUBROUTINE nssl_2mom_driver(qv, qc, qr, qi, qs, qh, qhl, ccw, crw, cci, csw, chw
       real, intent(in)::    dtp
       integer, intent(in):: itimestep !, ccntype
       logical, optional, intent(in) :: diagflag, f_cna, f_cn
-      integer, optional, intent(in) :: ipelectmp
+      integer, optional, intent(in) :: ipelectmp, ke_diag
 
   LOGICAL, INTENT(IN), OPTIONAL ::    nssl_progn   ! flags for wrf-chem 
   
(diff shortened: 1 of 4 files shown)


davegill:
jenkins
```
Please find result of the WRF regression test cases in the attachment. This build is for Commit ID: fdd2458934fa84d66bf777c5fbbf36e6f690a3e9, requested by: davegill for PR: https://github.com/wrf-model/WRF/pull/1170. For any query please send e-mail to David Gill.

    Number of Simulations  : 167           167        0
    Number of Comparisons  : 106           106        0

    Failed Simulations are: 
    None
    None
```

davegill:
@weiwangncar 
Wei,
I closed the other PR. This is the one to review. It is identical, but I swapped the order of the merging (I put these mods into v4.2, not all of v4.2 in to this repo).

phys/module_mp_thompson.F
```diff
@@ -1285,8 +1287,15 @@ SUBROUTINE mp_gt_driver(qv, qc, qr, qi, qs, qg, ni, nr, nc,       &
 
          IF ( PRESENT (diagflag) ) THEN
          if (diagflag .and. do_radar_ref == 1) then
+          
+          IF ( present(ke_diag) ) THEN
```
weiwangncar, line 1291: @davegill @MicroTed Here is another one...

phys/module_mp_nssl_2mom.F
```diff
@@ -2241,12 +2243,17 @@ SUBROUTINE nssl_2mom_driver(qv, qc, qr, qi, qs, qh, qhl, ccw, crw, cci, csw, chw
    ! calc dbz
       
       IF ( .true. ) THEN
+      IF ( present(ke_diag) ) THEN
```
weiwangncar, line 2246: @davegill @MicroTed I doubt this is the right check: ke_diag is always there, always defined. Perhaps ke_diag can be passed into the radar routine directly.

MicroTed: @weiwangncar Yes, going forward for WRF it is always there, but I also back-port current code to older versions we're still using around here (and 3 other models) where it doesn't exist.  I appreciate the thought!

That "IF ( .true.) " is old debugging cruft, though... hahaha.

weiwangncar:
@davegill I just ran a test with nwp_diagnostics on, and this code produced what it is supposed to do. I'm ok with this.

PR #1172, opened 2020-04-16
If (FARMS+Goddard MP) then -> must use RRTMG radiation

davegill:
TYPE: bug fix

KEYWORDS: FARMS, Goddard, RRTMG

SOURCE: internal

DESCRIPTION OF CHANGES:
A new fatal check is added to check_a_mundo.
```
IF ( FARMS and Goddard MP) THEN
   IF ( using any radiation scheme except for RRTMG ) THEN
      FAIL
```

LIST OF MODIFIED FILES:
modified:   module_check_a_mundo.F

TESTS CONDUCTED:
1. jenkins is all PASS
2. Code proceeds as expected with OK nml options. Basically, no harm done.
```
 &physics
 physics_suite = 'CONUS'
 swint_opt     = 2
 mp_physics    = 7
 ra_lw_physics = 4
 ra_sw_physics = 4
```
3. Code correctly stops with trapped bad combination
```
 &physics
 physics_suite = 'CONUS'
 swint_opt     = 2
 mp_physics    = 7
 ra_lw_physics = 1
 ra_sw_physics = 1
```
With this output:
```
--- ERROR: Option FARMS + Goddard MP requires RRTMG radiation
FATAL CALLED FROM FILE:  <stdin>  LINE:    2340
```
4. Code proceeds as expected with non-trapped schemes. Basically, still no harm done.
```
 &physics
 physics_suite = 'CONUS'
 swint_opt     = 1
 mp_physics    = 7
 ra_lw_physics = 4
 ra_sw_physics = 4
```

### DIFF
--- a/share/module_check_a_mundo.F
+++ b/share/module_check_a_mundo.F
@@ -576,6 +576,27 @@
          END IF
       ENDDO
 
+!-----------------------------------------------------------------------
+! The combination of using the FARMS option (swint_opt==2) along with the
+! Goddard MP scheme requires that the RRTMG radiation be used.
+!-----------------------------------------------------------------------
+      oops = 0
+      DO i = 1, model_config_rec % max_dom
+         IF ( .NOT. model_config_rec % grid_allowed(i) ) CYCLE
+         IF ( ( model_config_rec%swint_opt .EQ. SWINTOPT2 ) .AND. &
+              ( model_config_rec%mp_physics(i) .EQ. nuwrf4icescheme ) ) THEN
+            IF ( ( model_config_rec%ra_sw_physics(i) .NE. rrtmg_lwscheme ) .OR. &
+                 ( model_config_rec%ra_lw_physics(i) .NE. rrtmg_swscheme ) ) THEN
+              oops = oops + 1
+            END IF
+         END IF
+      ENDDO
+
+      IF ( oops .GT. 0 ) THEN
+         wrf_err_message = '--- ERROR: Option FARMS + Goddard MP requires RRTMG radiation'
+         CALL wrf_message ( wrf_err_message )
+         count_fatal_error = count_fatal_error + 1
+      END IF
 #endif
 
 !-----------------------------------------------------------------------


davegill:
@weiwangncar @dudhia 
Jimy and Wei,
Just to be clear ... is _any_ flavor of RRTMG acceptable? For example KIAPS or FAST?

davegill:
@weiwangncar @dudhia 
Jimy and Wei,
A second question. Do we need to use RRTMG for both SW and LW? Is really only one of these important for the consistency test?

davegill:
jenkins
```
Please find result of the WRF regression test cases in the attachment. This build is for Commit ID: d2da025e227b36f1a9bc55da2f82edda6730fb08, requested by: davegill for PR: https://github.com/wrf-model/WRF/pull/1172. For any query please send e-mail to David Gill.

    Number of Simulations  : 164           164        0
    Number of Comparisons  : 104           104        0

    Failed Simulations are: 
    None
    None
```

weiwangncar:
@davegill I thought the idea is if you're using the new Goddard microphysics, you cannot use FARM.

davegill:
@dudhia @weiwangncar 
Jimy,
Is the new logic supposed to be
```
IF ( FARMS and Goddard MP) THEN
   IF ( using any radiation scheme except for RRTMG ) THEN
      FAIL
```
or
```
IF ( FARMS and Goddard MP) THEN
      FAIL
```

dudhia:
The key is the conversion code in the radiation driver for re_cloud_gsfc to
re_cloud.
So far this is only in the regular RRTMG sections. This means those others
won't work with Goddard mp either.
We are still looking into whether having one of RRTMG SW or LW works.

weiwangncar:
Apart RRTMG and the new Goddard, I don't think others are taking effective radii.

dudhia:
Yes true for the KIAPS version, but the fast version does use re
Jimy

weiwangncar:
We probably should put in the conversion for RRTMG_FAST.

dudhia:
It looks like it should be easy enough. Do you want to do it?
Jimy

davegill:
@dudhia @weiwangncar 
Folks,
We add 
```
             IF ( mp_physics .EQ. nuwrf4icescheme ) THEN
                DO j=jts,jte
                DO k=kts,kte
                DO i=its,ite
                   re_cloud(i,k,j) = re_cloud_gsfc(i,k,j) * 1.E-6
                   re_ice(i,k,j) = re_ice_gsfc(i,k,j) * 1.E-6
                   re_snow(i,k,j) = re_snow_gsfc(i,k,j) * 1.E-6
                ENDDO
                ENDDO
                ENDDO
```

This looks like it needs to be added only to RRTMG FAST and FARMS.
Also, with this new change, this current PR is not required, right?

dudhia:
That should work.
FARMS needs the re fields, so if there is a check-a-mundo it would be
to not use it with mp schemes that don't provide them (not sure we have
many).

davegill:
Upon further review, there is a better way to do this.

PR #1177, opened 2020-04-21
MYNN-EDMF enhancements/fixes for stable layers and solar forecasts

joeolson42:
TYPE: enhancement

KEYWORDS: stable layers, clouds, radiation 

SOURCE: Joseph Olson (NOAA)

DESCRIPTION OF CHANGES:
Two issues were revealed in recent testing:
1) While revisiting important case studies, we found that the maintenance of stable layers were degraded due to a combination of the addition of the top-down mixing from cloud-top radiative cooling and a small increase in the buoyancy mixing length scale. The following tweaks allowed us to regain the original HRRR performance (see figure 1):

- Reduction of top-down mixing from cloud-top radiative cooling over land. Although we don’t like to have land-water dependencies, this will have to do until a more general solution is found.
- Revert constant alp2 back to 0.3 from 0.35 as it was prior to v.4.2.

2) Solar forecast verification were showing a slight degradation relative to the v4.1 configuration. This is partly due to a reduction of the analytically prescribed subgrid-scale qi (qi_bl), which we think is justifiable, but we think we reduced it too much. Also, the mass-flux plumes were not penetrating the LCL as often as before due to the drying associated with the parameterized subsidence. The first two changes listed below were made to alleviate this problem and the third change is a necessary subsequent modification to go with the second change and also helps for coarse vertical grid spacing. Figure 2 shows evidence that these changes get the SWDOWN bias back to where we were before (or very slightly better in the afternoon-evening) but the MAE is still about the same.

- We increased the analytical specification for SGS ice mixing ratios (qi_bl), although I suspect it may still be too small. This at least helps, but in pure ice clouds (i.e. cirrus), we still have more SWDOWN than in previous versions.
- We also decreased the entrainment rates to allow plumes to penetrate a little higher – more likely to penetrate the LCL, condense, and form shallow cumulus.
- As plumes penetrate into higher model levels, where dz can become very large (>300 m), the detrainment rates that work well for moderate and small dz (< 300 m) result in excessive dynamical detrainment further aloft. This would have been notice much earlier if we tested with a variety of vertical grid spacing but we currently do not. We added some limits in the detrainment section to address this issue and also slightly reduced come coefficients to minimize the chance of the dynamical detrainment to contribute to high RH biases at or above 700 mb.

Screen Shot 2020-04-21 at 2 14 26 PM
Figure 1. Stratus-topped stable layer mix-out over the front range. The HRRRv4 actually holds onto the clouds (as shown by the ceilings) and with the mods described above, now the v4.2 version of the MYNN does as well.

Screen Shot 2020-04-21 at 2 17 16 PM
Figure 2. SWDOWN verification against SURFAD for a 6-day HRRR retrospective run. Red is HRRRv4, blue is the proposed v4.2 MYNN. With the mods described above, the HRRR can at least match it's former performance. 

LIST OF MODIFIED FILES: M phys/module_bl_mynn.F

TESTS CONDUCTED:
Jenkins tests all PASS.


davegill:
@dudhia @weiwangncar @joeolson42 
Jimy and Wei,
Other than a new IF test block, these are just replacing numerical constants. I just want to say that I am good with this PR being merged into v4.2.

Please review

dudhia:
Looks fine to me. Thanks for the explanations, Joe.

weiwangncar:
I am ok with the change.

PR #1179, opened 2020-04-22
Bug fixes for MYNN-EDMF: two uninitialized variables 

joeolson42:
TYPE: bug fix

KEYWORDS: remember to compile to check for fpe every now and then

SOURCE: Joseph Olson

DESCRIPTION OF CHANGES: 
Found 2 uninitialized variables that were found in the FV3 version of the MYNN-EDMF when compiling with the fpe debug flags. The impacts are small, should be hidden in the noise (see example below).

LIST OF MODIFIED FILES: 
phys/module_bl_mynn.F

TESTS CONDUCTED: 

Screen Shot 2020-04-22 at 1 22 55 PM
This figure shows the 2-m temperature differences (bugfix - control) from a 6-hour forecast over the entire HRRR domain.


joeolson42:
oops.... 13 commits?

davegill:
@joeolson42 
> oops.... 13 commits?

Joe,
No biggy. This is correct. Only a couple of lines changed.

dudhia:
seems OK to me

davegill:
@joeolson42 
Joe,
Would you post the text from the jenkins email in these comments

weiwangncar:
I'm ok with it too.

joeolson42:
Please find result of the WRF regression test cases in the attachment. This build is for Commit ID: 975696bddf6563ca0cb0515643832019155060ee, requested by: joeolson42 for PR: https://github.com/wrf-model/WRF/pull/1179. For any query please send e-mail to David Gill.

    Number of Simulations  : 164           164        0
    Number of Comparisons  : 104           104        0

    Failed Simulations are: 
    None
    None

dudhia:
Seems important and it is surprising the bug makes it through
regtests using an
uninitialized variable.

joeolson42:
I agree. I typically run HRRR case studies with experimental version of the MYNN compiled with:

FCDEBUG         =       -g $(FCNOOPT) -traceback # -fpe0 -check... etc ...

after initially compiling the whole model with this entire line commented out. I usually don't use the -fpe flag but that probably would have caught it. The compile script in FV3/CCPP has a DEBUG mode that uses the -fpe0 flag by default. That caught it for me.

dudhia:
If this occurs at a processor boundary, our bit-for-bit checks can also
catch them
depending if the value comes from a previous loop index.

PR #1182, opened 2020-04-23
Bug fix for WRFPLUS with SBM

davegill:
TYPE: bug fix

KEYWORDS: WRFPLUS, AD, SBM

SOURCE: internal

DESCRIPTION OF CHANGES:
A new argument was added to the nonlinear model for the microphysics
driver (an integer flag for the spectral bin microphysics diagnostics
`sbm_diagnostics`). This argument has been added to the call to the
microphysics driver in `solve_em_ad.F`.

While tracking down this compile-time bug, an old debugging comment for 
WRFPLUS in the `compile` script was found and removed.

LIST OF MODIFIED FILES:
modified:   compile
modified:   wrftladj/solve_em_ad.F

TESTS CONDUCTED:
1. Without mods, the WRFPLUS code does not compile. With mod, the WRFPLUS
code compiles.
2. Jenkins will continue to pass because this code was not activated.
3. WRFPLUS regtest: ???

### DIFF
--- a/compile
+++ b/compile
@@ -73,7 +73,6 @@ foreach a ( $argv )
     set ZAP = ( main/wrfplus.exe )
     setenv WRF_EM_CORE   1
     setenv WRF_PLUS_CORE 1
-    echo not sure what to do, but this seems to be legit so far
   else if   ( "$a" ==  "wrf" ) then
     set arglist = ( $arglist $a )
     set ZAP = ( main/wrf.exe )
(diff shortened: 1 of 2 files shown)


davegill:
@liujake 
Jake,
Please review. With this mod, the WRFPLUS code executable builds.

liujake:
so newly-added 'ke_diag' argument does not matter? I compiled it by adding also ke_diag, but not remove it with only adding sbm_diagnostics.

davegill:
@liujake 
Jake,
The `ke_diag` value is optional.
```
   INTEGER, OPTIONAL, INTENT(IN) :: ke_diag ! tells reflectivity calculation whether to do full depth or only k=1
```
So, no need for that to be in the solve_em_ad.F call to the microphysics driver.

PR #1190, opened 2020-05-04
Filling aerssa2d/aerasy2d if sw = 5 and prescribed values from namelist

pedro-jm:
TYPE: bug fix

KEYWORDS: aer_opt = 2, Goddard SW, single scattering albedo, asymmetry factor, prescribed aerosol optical properties from namelist.input

SOURCE: Pedro A. Jimenez and Ju Hye Kim (NCAR/RAL), and J. Dudhia (NCAR/MMM) 

DESCRIPTION OF CHANGES: Fill in aerssa2d and aerasy2d arrays if aer_opt = 2 and sw = 5 (Goddard shortwave parameterization) with the values imposed in the namelist via aer_ssa_opt = 1 and/or aer_asy_opt = 1 options. The variables aerssa2d and aerasy2d are always present for ARW so there is no concern with memory packaging. The filling of the variables is properly handled by RRTMG but not by the Goddard SW parameterization. Similarly to the RRTMG implementation, no special treatment for HWRF (e.g. ifdef) is necessary to fill in the variables. If the variables are not filled in, FARMS will not use the values specified in the namelist. Instead, FARMS will use zero values for the single scattering albedo (SSA) and the asymmetry factor (g) leading to an incorrect irradiance parameterization. 

LIST OF MODIFIED FILES: 
M       phys/module_ra_aerosol.F

TESTS CONDUCTED: 
1. Before the fix, aerssa2d and aerasy2d were zero. 
2. After the fix these arrays have the values specified in the namelist.
3. Jenkins test is all PASS

RELEASE NOTE: Bug fix for aer_opt =2 and Goddard SW radiation to fill in the 2D arrays with the aerosol optical properties (SSA and g) from namelist.input. The bug affects the FARMS output which will not properly account for either SSA or g (both values will be zero).

### DIFF
--- a/phys/module_ra_aerosol.F
+++ b/phys/module_ra_aerosol.F
@@ -256,6 +256,12 @@ subroutine calc_aerosol_goddard_sw(ht,dz8w,p,t3d,qv3d,aer_type,                 
              end do
           end do
 
+          do j = jts, jte
+             do i = its, ite
+                aerssa2d(i, j) = aer_ssa_val
+             end do
+          end do
+
        case(2)
           if (.not.(present(aerssa2d))) then
              write(wrf_err_message,*) &
@@ -327,6 +333,12 @@ subroutine calc_aerosol_goddard_sw(ht,dz8w,p,t3d,qv3d,aer_type,                 
                       asyaer(i,k,j,nb)=aer_asy_val
                    end do
                 end do
+             end do
+          end do
+
+          do j = jts, jte
+             do i = its, ite
+                aerasy2d(i, j) = aer_asy_val
              end do
           end do
 


davegill:
@pedro-jm 
Pedro,
Please include a statement such as "Jenkins test is all PASS" in the commit message under the TESTING section (it indeed you did get that message).

davegill:
@pedro-jm 
Pedro,
Please include that the variables aerssa2d and aerasy2d are always present for ARW in your commit message, so there is no concern with memory or packaging. How is HWRF handled? Is it bypassed with an `ifdef`? Include that information in your commit message as well.

pedro-jm:
@davegill 
Dave, I enlarged the commit message to include the above information. I do not think there is a special treatment for HWRF. In any case I replicated what is done for RRTMG in the Goddard subroutine. So even if there is a special treatment for HWRF it was there before this fix.
Pedro.

weiwangncar:
@pedro-jm Do you know if Goddard radiation can use aer_opt = 2 option? Sometime ago when Jose worked on this option, Goddard radiation worked. But since it was updated recently, we haven't tested it.

pedro-jm:
@weiwangncar 
To the extent of my knowledge aer_opt = 2 works with Goddard. We are extensively using the option that imposes a value from the namelist, like this PR, using WRF version 4.0.1. 
Ju Hye is going to do a final cross check, but I think is working fine.
Pedro.

pedro-jm:
@weiwangncar 
Ju Hye indicates that she does not see any issues with Goddard and aer_opt =2 (imposing the values from the namelist like this PR) in WRFv4.2+ (This PR code). 
Pedro.

weiwangncar:
@pedro-jm Thanks for the update. Good to know.

PR #1191, opened 2020-05-05
FARMS can use climatological effective radius

pedro-jm:
TYPE:  Enhancement

KEYWORDS: FARMS, climatological hydrometeor radius

SOURCE: Ju Hye Kim (NCAR/RAL), Jimy Dudhia (NCAR/MMM) and Pedro Jimenez (NCAR/RAL)

DESCRIPTION OF CHANGES: FARMS required the microphysics parameterization to provide the effective radius of the hydrometeors. If a given micrphysics parameterization does not provide the effective radius FARMS could not be used. To avoid this limitation we have enhanced FARMS to use climatological effective radius if the microphysics parameterization does not provide them. 

LIST OF MODIFIED FILES: 
M       Registry/Registry.EM_COMMON
M       phys/module_physics_init.F
M       phys/module_ra_farms.F
M       phys/module_radiation_driver.F

TESTS CONDUCTED: We have tested that FARMS works when one uses the microphysics option 2. This microphysics does not provide effective radius so FARMS is using the climatological effective radius. Jenkins test is all PASS.

RELEASE NOTE: FARMS radiation parameterization uses climatological effective radius of the hydrometeors if the radius is not provided by the microphysics parameterization.

PR #1197, opened 2020-05-18
FARMS accounts for the effective radius of unresolved hydrometeors

pedro-jm:
TYPE: bug fix

KEYWORDS: FARMS, effective radius unresolved hydrometeors

SOURCE: Pedro A. Jimenez and Ju Hye Kim (NCAR/RAL), and Jimy Dudhia (NCAR/MMM)

DESCRIPTION OF CHANGES: 
The parameterizations accounting for radiative effects of unresolved clouds do not provide 
effective radius. The radius is estimated based on climatologies inside the radiation 
parameterizations. The radius climatologies were missing in FARMS. We have added the 
climatologies for the effective radius for liquid, ice and snow hydrometeors. The fix is valid 
for mp = 6, 8, 28 and for all the microphysics options that do not provide effective radius. We 
have added the logic to stop and raise and exception if an incompatible microphysics is 
selected.

The default radius is changed to zero instead of a small value.

This PR is going to create a conflict with PR #1191. It is better to merge PR #1191 first, and afterwards, merge the present PR.

LIST OF MODIFIED FILES: 
M       phys/module_mp_wsm6.F

TESTS CONDUCTED: 
1. We have run FARMS with mp = 6, 8, and 28 and the code behaves as expected. 
2. We have also run the code of this PR with mp = 2 that does not provide the effective radius 
of the hydrometeors and the code works fine. 
3. If a non-supported microphysics is selected WRF stops and prints the correct error.
4. Jenkins test is all PASS

RELEASE NOTE: Bug fix to account for the effective radius of the unresolved hydrometeors by FARMS.

### DIFF
--- a/phys/module_mp_thompson.F
+++ b/phys/module_mp_thompson.F
@@ -1304,16 +1304,16 @@
 
          IF (has_reqc.ne.0 .and. has_reqi.ne.0 .and. has_reqs.ne.0) THEN
           do k = kts, kte
-             re_qc1d(k) = 2.49E-6
-             re_qi1d(k) = 4.99E-6
-             re_qs1d(k) = 9.99E-6
+             re_qc1d(k) = 0.0
+             re_qi1d(k) = 0.0
+             re_qs1d(k) = 0.0
           enddo
           call calc_effectRad (t1d, p1d, qv1d, qc1d, nc1d, qi1d, ni1d, qs1d,  &
                       re_qc1d, re_qi1d, re_qs1d, kts, kte)
           do k = kts, kte
-             re_cloud(i,k,j) = MAX(2.49E-6, MIN(re_qc1d(k), 50.E-6))
-             re_ice(i,k,j)   = MAX(4.99E-6, MIN(re_qi1d(k), 125.E-6))
-             re_snow(i,k,j)  = MAX(9.99E-6, MIN(re_qs1d(k), 999.E-6))
+             if (re_qc1d(k) /= 0.0) re_cloud(i,k,j) = MAX(2.49E-6, MIN(re_qc1d(k), 50.E-6))
+             if (re_qi1d(k) /= 0.0) re_ice(i,k,j)   = MAX(4.99E-6, MIN(re_qi1d(k), 125.E-6))
+             if (re_qs1d(k) /= 0.0) re_snow(i,k,j)  = MAX(9.99E-6, MIN(re_qs1d(k), 999.E-6))
           enddo
          ENDIF
 
@@ -5065,7 +5065,7 @@
 
       if (has_qc) then
       do k = kts, kte
-         re_qc1d(k) = 2.49E-6
+         re_qc1d(k) = 0.0
          if (rc(k).le.R1 .or. nc(k).le.R2) CYCLE
          if (nc(k).lt.100) then
             inu_c = 15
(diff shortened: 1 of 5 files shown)


davegill:
jenkins is OK with non-KPP (previous KPP was also OK)
```
Please find result of the WRF regression test cases in the attachment. This build is for Commit ID: e108e28844a8a6c6a9343d1de1dea036bd398b9e, requested by: pedro-jm for PR: https://github.com/wrf-model/WRF/pull/1197. For any query please send e-mail to David Gill.

    Failed Simulations are: 
    None
    None
```

davegill:
@pedro-jm @dudhia @weiwangncar 
> This PR is going to create a conflict with PR #1191. It is better to merge PR #1191 first, and afterwards, merge the present PR.

Pedro,
The PR #1191 is to the develop branch, and this PR is a bug fix to v4.2.1.  If this PR should come after #1191, then this PR should be to the develop branch, too. Right?

pedro-jm:
Dave @davegill 
You are right with the logic. We took a look to the codes of both PRs and we realized that is even simpler. This PR updates #1191. So there is no need to merge #1191. We can ignore #1191 and move forward merging this PR only.
This PR is to the bug fix branch and this is the correct branch since this fixes the bug described in the commit message.
Pedro.

davegill:
@weiwangncar @dudhia 
Folks,
Can you take a look at this bug-fix PR for MP effective radii from Pedro.

dudhia:
I think this changes the default radius to zero instead of a small value. That is not mentioned in the PR. That should be fine as long as a radius is provided when there are only non-microphysics clouds in the radiation, which I think is true but should be verified.

weiwangncar:
@pedro-jm Why is it needed to change the effective radii to 0. rather than keeping them as small? If you do this, it seems it will not just impact FARMS, but a WRF run that makes use of the effective radii from the microphysics. Is Greg ok this change for his schemes?

pedro-jm:
@dudhia 
Ju-Hye will verify your comment. I updated the PR description to mention that the effective radius is changed to 0 if there are no hydrometeors.

@weiwangncar 
We could have kept the small value to indicate that there are not resolved hydrometeors. However, we changed the value to better document the code. It is more logical that if there are no hydrometeors the effective radius is 0. 
I think it is fine to have default values for the radius that are equal or smaller than the previous default value. RRTMG checks the effective radius and if it is smaller than a threshold assigns a clim effective radius. So changing the value to zero will not affect the logic on RRTMG. We did not look at other radiation schemes.
Greg has been informed and he is OK with this change.

weiwangncar:
@pedro-jm Thanks.

pedro-jm:
Just a quick update. We have tested Goddard radiation with these changes and we have found the expected behavior.

weiwangncar:
@pedro-jm Can you confirm that changing the default from a small value to zero DOES NOT change the results?

weiwangncar:
@pedro-jm Two more questions: What do you mean by (in your PR message) 'The fix is valid for mp = 6, 8, 28 ' which do provide effective radii? Do you mean for these schemes, the default values are changed? If so, this should be explicitly stated. The other question is about 'unresolved clouds'. Again what do you mean by that?

dudhia:
I think this PR is on hold because we tried to change the default radius
to zero and need to track down some differences. Pedro is out at the moment,

weiwangncar:
We should move this PR as well as #1191 to 4.2.2 then.

davegill:
@weiwangncar @dudhia 
> We should move this PR as well as #1191 to 4.2.2 then.

1. I'll remove this from the v4.2.1 label
2. The #1191 PR is o longer under consideration, according to Pedro:

> Dave
You are right with the logic. We took a look to the codes of both PRs and we realized that is even simpler. This PR updates #1191. So there is no need to merge #1191. We can ignore #1191 and move forward merging this PR only.

weiwangncar:
@pedro-jm @dudhia Is PR #1295 replacing this PR? If so, please close this one.

dudhia:
replaced by #1295 - closing

PR #1201, opened 2020-05-29
Commit

janmandel:
Print a line with the git version hash and if any files have changed 

TYPE: no impact

KEYWORDS: commit hash, check changes, user support, development, version control

SOURCE: Jan Mandel (University of Colorado Denver)

DESCRIPTION OF CHANGES:  User support is more difficult without knowing if the user may have changed any files in the distribution, and, for developers, it is not possible to check if the wrf.exe binary is the same version as the source files, leading to unneeded rebuilds. Including a line with the git commit hash and a summary of changes in addition to the WRF release version makes it possible to see quickly if any files were changed, and the command line
strings wrf.exe | grep 'git commit' 
shows what exact version is the binary from. The functionality is implemented by generating a new file inc/commit_decl and using it similarly as inc/version_decl. The new file is deleted by cleaning the build.

LIST OF MODIFIED FILES:
M       clean
M       compile
M       main/module_wrf_top.F

TESTS CONDUCTED: Built and run with Intel and GNU compilers, with and without git available, in a cloned repository, and in a directory not under git control.

RELEASE NOTE: A line is printed with the git commit version and a summary if any files were changed.


janmandel:
This PR got confused apparently because was based on a wrong branch. I do not see how to change the base branch, instead resubmitted cleanly as #1202.

PR #1214, opened 2020-06-19
bug fix for m_opt: fixed units of surface momentum flux

Matze77:
TYPE: bug fix

KEYWORDS: m_opt, momentum flux 

DESCRIPTION OF CHANGES:
When outputting SGS momentum fluxes using the option m_opt, the units of the fluxes should be m2 s-2 according 
to registry.les. The fluxes above the surface are divided by the density and thus have the correct units. The surface 
fluxes, however, still contain the density.

This bug fix divides the surface fluxes by the density to obtain the correct units.

This change only affects the output of the SGS fluxes (nba_mij). The tendencies and thus the results are not changed.

LIST OF MODIFIED FILES:
dyn_em/module_diffusion_em.F

TESTS CONDUCTED:
1. Jenkins OK

RELEASE NOTE: Fixed units of surface momentum flux in dyn_em/module_diffusion_em.F when using m_opt. This bug fix divides the surface fluxes by the density to obtain the correct units. This change only affects the output of the SGS fluxes (nba_mij). The tendencies and thus the results are not changed.

### DIFF
--- a/dyn_em/module_diffusion_em.F
+++ b/dyn_em/module_diffusion_em.F
@@ -4165,8 +4165,8 @@ SUBROUTINE vertical_diffusion_2   ( ru_tendf, rv_tendf, rw_tendf, rt_tendf,   &
                            v_2(i-1,kts,j  )+          &
                            v_2(i-1,kts,j+1))/4)**2))+epsilon
 
-       tao_xz=cd0*V0_u*u_2(i,kts,j)*(rho(i,kts,j)+rho(i-1,kts,j))/2.
-       ru_tendf(i,kts,j)=ru_tendf(i,kts,j) +   g*tao_xz/dnw(kts)
+       tao_xz=cd0*V0_u*u_2(i,kts,j)
+       ru_tendf(i,kts,j)=ru_tendf(i,kts,j) +   g*tao_xz*0.5*(rho(i,kts,j)+rho(i-1,kts,j))/dnw(kts)
        IF ( (config_flags%m_opt .EQ. 1) .OR. (config_flags%sfs_opt .GT. 0) ) THEN
           nba_mij(i,kts,j,P_m13) = -tao_xz
        ENDIF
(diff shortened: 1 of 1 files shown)


davegill:
@Matze77 @weiwangncar @dudhia 
When there are the introduction of new horizontal differences in algorithms, we must be concerned about boundary differences that would be caused by parallel communications. However, the d/dx and d/dy of rho already exists in those loops, so this is OK.

davegill:
@dudhia 
Jimy,
So these are entirely diagnostic?

dudhia:
That was my question.
With some digging I could find out but it is easier to ask the proposer.

dudhia:
This code change needs to be run to check the impact anyway. It is an
sfs_opt, m_opt combo.

Matze77:
@dudhia @davegill 
yes, this change only affects the output of the SGS fluxes (nba_mij). The tendencies and thus the results are not changed.

davegill:
@dudhia @Matze77 
Jimy,
Are you OK with this PR now? Is there a test that you are requesting of Matthias?

dudhia:
Either Matthias or our people could run a test to demonstrate the
non-effect on
other results when the change is made.

Matze77:
In my opinion, a test is not strictly necessary. I thoroughly checked that nba_mij, which is modified by this PR, is not used later on in any tendency calculations. The tendency calculations are happening right before the location of the changes. Thus, I don't want to invest time in a test run. If you want to have it tested anyway, you would have to do it yourself, I'm afraid. Or you ask the developers of that sfs_opt option. They can probably judge this PR much better than I can.

dudhia:
Yes, we could do it internally. Thanks for this fix,
Jimy

weiwangncar:
Should we contact the original contributor to check on this?

dudhia:
He already replied in the thread that he won't test it but is sure of it
not affecting results.
Up to us to test it if we want.

dudhia:
I will look at the code and get back to you.

weiwangncar:
@dudhia I meant the original developer who contributed this option..

dudhia:
It looks like a valid bug. I don't think we have any questions. If you
want, that would be Jeff Mirocha.

weiwangncar:
@dudhia I have sent an email to Jeff, and let's see if we can get any response.

PR #1221, opened 2020-06-23
Remove nonstandard spacing in macro as workaround for LLVM bug 

### LINKED ISSUE #1206: Usage of macros is incompatible with clang cpp -traditional-cpp
**Problem Description**
When compiling on MacOS using both the gcc and clang compiler configurations, building fails when the system version of `cpp` is used. The reason has to do a difference in the way that the MacOS system version of cpp (for me, Apple clang version 11.0.3 clang-1103.0.32.62) treats spacing in macros when using the `-traditional-cpp` flag. The difference could be considered a bug in llvm, but the [bug report documenting the difference](https://bugs.llvm.org/show_bug.cgi?id=18011) has been untouched since 2013 and the developers have explicitly expressed their intentions not to provide more than bare-bones support for `-traditional-cpp`.

The specific problem is that the clang distribution of cpp does not recognize and replace usages of macros with parameters if the macro is used with a space between the macro name and the opening parenthesis. Technically, there cannot be a space there in the macro definition, but 

As an example, in `phys/module_mp_fast_sbm.F`, a macro is defined like this:
```Fortran
#define DM_BCAST_MACRO_R16(A) CALL wrf_dm_bcast_bytes(A, size(A)*R16SIZE)
```
and used like this:
```Fortran
DM_BCAST_MACRO_R16 ( FAF1 )
```

In most instances, there is no space between the macro name and the opening parenthesis, but in this particular file there are a few instances that cause compilation to fail, because the macros are never replaced with valid Fortran code and therefore cannot be compiled to object files.

**Current Workaround**:
Changing the $PATH or creating a symbolic link to use the actual gcc preprocessor rather than the clang version called by default when using `cpp`. The real gcc version exhibits correct behavior with the flag.

**Proposed Solution**:
Remove "non-standard" spacing from macros. I searched through all the Fortran files, and the only one that uses this spacing that trips up the clang distribution of cpp is `phys/module_mp_fast_sbm.F`. Removing the space between the macro name and parenthesis is sufficient to fix the issue.

I have made local commits to fix this problem, but I haven't contributed to this project before and don't know what the correct process is. If everyone concurs on this solution, I can work on putting in a PR for this. It should only be one commit with one file diff.

irowebbn:
TYPE: bug fix

KEYWORDS: macros, compiler, llvm, clang

SOURCE: Isaac Rowe (University of Kentucky)

DESCRIPTION OF CHANGES: 
Problem:
Some versions of LLVM clang cannot parse spaces near parentheses in cpp macros. For example:
```
DM_BCAST_MACRO_R16 ( FAF1 )
```
This bug has been discussed on the LLVM site: https://bugs.llvm.org/show_bug.cgi?id=18011. 

Solution:
The surrounding spaces around the  parentheses from the macros with parameters in 
phys/module_mp_fast_sbm.F were removed. 
```
DM_BCAST_MACRO_R16(FAF1)
```
Only a single file uses this "nonstandard" spacing for the cpp macros.

ISSUE: 
Fixes #1206 (Usage of macros is incompatible with clang cpp -traditional-cpp).

LIST OF MODIFIED FILES: 
phys/module_mp_fast_sbm.F

TESTS CONDUCTED: 
1. Now builds with LLVM cpp (Apple clang version 11.0.3 (clang-1103.0.32.62)). Previously received `Unclassifiable statement error` at `module_mp_fast_sbm.f90:6065:4` during next step in compilation.
2. Jenkins test all PASS.

RELEASE NOTES: A fix was provided to support building the WRF system with LLVM clang. Some versions of LLVM clang cannot correctly parse spaces near parentheses in cpp macros, and those spaces have been removed.

### DIFF
--- a/phys/module_mp_fast_sbm.F
+++ b/phys/module_mp_fast_sbm.F
@@ -6052,25 +6052,25 @@ enddo
 
  ! ... (KS) - Broadcating Liquid drops
 #if (defined(DM_PARALLEL))
-   	DM_BCAST_MACRO_R16 ( FAF1 )
-   	DM_BCAST_MACRO_R16 ( FBF1 )
-   	DM_BCAST_MACRO_R16 ( FAB1 )
-   	DM_BCAST_MACRO_R16 ( FBB1 )
+   	DM_BCAST_MACRO_R16(FAF1)
+   	DM_BCAST_MACRO_R16(FBF1)
+   	DM_BCAST_MACRO_R16(FAB1)
+   	DM_BCAST_MACRO_R16(FBB1)
    ! ... (KS) - Broadcating Snow
-   	DM_BCAST_MACRO_R16 ( FAF3 )
-   	DM_BCAST_MACRO_R16 ( FBF3 )
-   	DM_BCAST_MACRO_R16 ( FAB3 )
-   	DM_BCAST_MACRO_R16 ( FBB3 )
+   	DM_BCAST_MACRO_R16(FAF3)
+   	DM_BCAST_MACRO_R16(FBF3)
+   	DM_BCAST_MACRO_R16(FAB3)
+   	DM_BCAST_MACRO_R16(FBB3)
    ! ... (KS) - Broadcating Graupel
-   	DM_BCAST_MACRO_R16 ( FAF4 )
-   	DM_BCAST_MACRO_R16 ( FBF4 )
-   	DM_BCAST_MACRO_R16 ( FAB4 )
-   	DM_BCAST_MACRO_R16 ( FBB4 )
+   	DM_BCAST_MACRO_R16(FAF4)
+   	DM_BCAST_MACRO_R16(FBF4)
+   	DM_BCAST_MACRO_R16(FAB4)
+   	DM_BCAST_MACRO_R16(FBB4)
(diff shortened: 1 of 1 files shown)


davegill:
@irowebbn 
Isaac,
Did you receive an email from jenkins about the testing status? For me, it originally showed up in my spam folder.

irowebbn:
Yes, I received the email: 
```

   Failed Simulations are:
   None
   None
```

davegill:
@irowebbn 
Isaac,
Please review the above PR message that we have produced. Use this as an example in your other PRs to WRF that you are working on.

In the PR message, it is always nice to have a "before vs after" in the testing area. For this example, if you have the text output from the failed parsing, that would be nice to show. Then for the "after" part, you can say that now the code builds and runs. Additionally, can you put a version of clang that has troubles?

irowebbn:
@davegill Thanks for the helpful feedback. I made some edits noting the version of LLVM clang I am using and the error I was previously getting.

davegill:
@irowebbn 
Isaac,
This is now a well put together PR. For your other PRs, this is a good template to emulate. 

Thanks for participating in the "community" aspect of WRF.

PR #1247, opened 2020-07-09
Fix 1d advection testing program, broken with hybrid vert coord

davegill:
TYPE: bug fix

KEYWORDS: advection, program, 1d

SOURCE: internal

DESCRIPTION OF CHANGES: 
There were two types of problems. First the original cpp commands were broken. The cpp command was commented out, 
but it is used to build the source code, with cut and paste). The second problem is caused by the modification of the 
advection interface with the release of v4.0.

First Problem and Solution:
1. Original suggested using `cpp -traditional`. This is an incorrect cpp option. The flag is supposed to be 
`cpp -traditional-cpp`.

2. Original suggested using `cpp -C -P`. The `-C` flag puts these GNU comments in the code (which is OK if you 
are writing in the C language).
```
/* Copyright (C) 1991-2012 Free Software Foundation, Inc.
   This file is part of the GNU C Library.

   The GNU C Library is free software; you can redistribute it and/or
   modify it under the terms of the GNU Lesser General Public
   License as published by the Free Software Foundation; either
   version 2.1 of the License, or (at your option) any later version.

   The GNU C Library is distributed in the hope that it will be useful,
   but WITHOUT ANY WARRANTY; without even the implied warranty of
   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
   Lesser General Public License for more details.

   You should have received a copy of the GNU Lesser General Public
   License along with the GNU C Library; if not, see
   <http://www.gnu.org/licenses/>.  */

```

Second Problem:
The existing advection testing program had not been updated to assume the existence of the hybrid vertical coordinate.
That meant a few variables in the calls to the scalar advection routines (c1 and c2) were missing. The advection testing
tool would not compile.

Second Solution:
Initialize the 1d arrays c1 and c2 to terrain-following coordinate settings (c1=1, c2=0), and pass those args into the 
WRF scalar advection drivers. The simplified settings are not important since this is just for a 1d advection.

LIST OF MODIFIED FILES: 
modified: dyn_em/module_advection_em.F

TESTS CONDUCTED: 
1. All of the changes are inside of ifdefs, so there is no impact to the compilable WRF code. The advection main program
had the interfaces updated to match the current advection drivers.
2. Previously, the advection testing code did not compile. Now the tester program builds and runs. Here are the results 
after 0, 1200, 1400, 1600, 1800, and 2000 full time steps (each full time step is constructed of three small time steps). 
Results from the advection schemes scalar_adv_opt = 0 (standard scalar advection), 1 (positive definite), 2 (monotonic), 
3 (WENO), and 4 (WENO PD) are shown. Every 200 full steps, the square wave should be back in the original location.

Screen Shot 2020-07-09 at 12 45 43 PM

Screen Shot 2020-07-09 at 12 47 18 PM

Screen Shot 2020-07-09 at 12 42 29 PM

Screen Shot 2020-07-09 at 12 49 10 PM

Screen Shot 2020-07-09 at 12 52 12 PM

RELEASE NOTE: I don't want to support this, so no release information.

### DIFF
--- a/dyn_em/module_advect_em.F
+++ b/dyn_em/module_advect_em.F
@@ -2,7 +2,7 @@
 !WRF:MODEL_LAYER:DYNAMICS
 !
 #if ( defined(ADVECT_KERNEL) )
-! cpp -traditional -C -P -DADVECT_KERNEL module_advect_em.F > advection_kernel.f90
+! cpp -traditional-cpp -P -DADVECT_KERNEL module_advect_em.F > advection_kernel.f90
 ! gfortran -ffree-form -ffree-line-length-none advection_kernel.f90
 ! ./a.out
 MODULE advection_kernel
--- a/dyn_em/module_advect_em.F
+++ b/dyn_em/module_advect_em.F
@@ -10558,7 +10558,8 @@ PROGRAM feeder
                                             msfty
    REAL , DIMENSION( : ), ALLOCATABLE :: fzm, &
                                   fzp, &
-                                  rdzw, znw,dnw, rdnw, dn, rdn
+                                  rdzw, znw,dnw, rdnw, dn, rdn, &
+                                  c1, c2
    REAL :: rdx, &
            rdy, &
            dt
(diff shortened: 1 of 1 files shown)


davegill:
@weiwangncar @dudhia @kkeene44 @smileMchen 
Folks,
No one has ever used this capability, except for me. I did need it to generate some figs for a paper review, so I am updating it in the repo. There are no impacts to WRF, as this code is all buried inside of ifdefs.

I'd like someone to give it a cursory review.

weiwangncar:
Looks reasonable to me.

davegill:
@kkeene44 
Kelly,
Wei said this 1d advection thing was OK. Would you go ahead and review?
Thanks

PR #1259, opened 2020-07-22
additional factor when converting surface heat fluxes from dry to moist theta

matzegoebel:
To close the dry theta budget at the lowest grid point when use_theta_m is enabled, the surface heat fluxes, and in the case of km_opt=5 also the non-local heat flux, are multiplied with 1+Rv/Rd*qv.

TYPE: bug fix

KEYWORDS: surface heat flux, use_theta_m, moist theta, non-local flux, 3D TKE

DESCRIPTION OF CHANGES: The conversion of surface heat fluxes from dry to moist theta in WRF is inconsistent. This leads to an unclosed dry theta budget at the lowest grid point. The surface sensible heat flux in the routines `vertical_diffusion_2` and `vertical_diffusion_implicit` should be multiplied with  as is derived below.
Using Reynold's decomposition and averaging we can show for the perturbation moist theta:
 
with water vapor mixing ratio q and
 

This yields for the converted surface heat flux:

The triple correlation in the equation above cannot be easily estimated in WRF, but judging from my tests it probably has a negligible impact.
Instead of the given equation, WRF uses:

We therefore multiplied the surface heat flux with , when `use_theta_m=1`.
For `km_opt=5`, also the contribution from the non-local heat flux is multiplied with a correction factor.

LIST OF MODIFIED FILES: dyn_em/module_diffusion_em.F 

TESTS CONDUCTED: checked the dry theta budget of a simulation with use_theta_m=1: simple idealized LES once with model-computed surface heat flux and once with prescribed heat fluxes, RRTMG radiation, no microphysics, zero eddy diffusivities. Changed the code to output advection, SGS diffusion and radiation tendencies for dry theta (regardless of use_theta_m and without changing the tendencies used by the model). These forcing tendencies are averaged online over 30 minutes. The sum of the forcing terms is plotted against the actual tendency of dry theta  (change of instantaneous dry theta between the half-hourly time stamps divided by 30 minutes) for all gridpoints and time stamps. All tendency terms are divided by the dry air mass to obtain more familiar units. The modified code fixes the issue of an unclosed dry theta budget at the lowest gridpoint (see attached figures). The simulations were repeated for km_opt=5.

RELEASE NOTE: To close the dry theta budget at the lowest grid point when use_theta_m is enabled, the surface heat fluxes, and in the case of km_opt=5 also the non-local heat flux, are multiplied with 1+Rv/Rd*qv.

test_fluxes_flat_thd_3rd+5th
test_fluxes_flat_thdm_3rd+5th
test_fluxes_flat_fixed-sfc-fluxes_thdm_3rd+5th

### DIFF
--- a/dyn_em/module_diffusion_em.F
+++ b/dyn_em/module_diffusion_em.F
@@ -8050,6 +8050,9 @@ END SUBROUTINE phy_bc
      DO i = i_start, i_end
         nlflux_rho(i,k,j) = nlflux(i,k,j) * &
                            (fnm(k)*rho (i,k,j)+fnp(k)*rho (i,k-1,j))
+        IF(config_flags%use_theta_m == 1)THEN
+             nlflux_rho(i,k,j) = nlflux_rho(i,k,j)*(1.+rvovrd*moist(i,k,j,P_QV))
+        ENDIF
      ENDDO
      ENDDO
      ENDDO
--- a/dyn_em/module_diffusion_em.F
+++ b/dyn_em/module_diffusion_em.F
@@ -4278,7 +4278,6 @@ SUBROUTINE vertical_diffusion_2   ( ru_tendf, rv_tendf, rw_tendf, rt_tendf,   &
   CASE (0,2) ! with fixed surface heat flux given in the namelist
     heat_flux = config_flags%tke_heat_flux  ! constant heat flux value
                                             ! set in namelist.input
-
     DO j = j_start, j_end
     DO i = i_start, i_end
        cpm = cp * (1. + 0.8 * moist(i,kts,j,P_QV))
@@ -4300,7 +4299,6 @@ SUBROUTINE vertical_diffusion_2   ( ru_tendf, rv_tendf, rw_tendf, rt_tendf,   &
 
        cpm = cp * (1. + 0.8 * moist(i,kts,j,P_QV))
        heat_flux = hfx(i,j)/cpm
-
        if(config_flags%use_theta_m == 1)THEN
          rt_tendf(i,kts,j)=rt_tendf(i,kts,j)  &
              -g*heat_flux*(1.+rvovrd*moist(i,kts,j,P_QV))/dnw(kts) &
@@ -7723,7 +7721,7 @@ END SUBROUTINE phy_bc
    INTEGER :: i_start, i_end, j_start, j_end
  
    REAL :: V0_u,V0_v,ustar,beta
-   REAL :: heat_flux, moist_flux
+   REAL :: heat_flux, moist_flux, qfac
    REAL :: cpm,rdz_w,rhoavg_u,rhoavg_v,rdz_z
 ! End declarations.
 !-----------------------------------------------------------------------
@@ -8050,9 +8048,6 @@ END SUBROUTINE phy_bc
      DO i = i_start, i_end
         nlflux_rho(i,k,j) = nlflux(i,k,j) * &
                            (fnm(k)*rho (i,k,j)+fnp(k)*rho (i,k-1,j))
-        IF(config_flags%use_theta_m == 1)THEN
-             nlflux_rho(i,k,j) = nlflux_rho(i,k,j)*(1.+rvovrd*moist(i,k,j,P_QV))
-        ENDIF
      ENDDO
      ENDDO
      ENDDO
@@ -8103,17 +8098,23 @@ END SUBROUTINE phy_bc
                   - dt*rdz_w*nlflux_rho(i,k+1,j)        &
                   + dt*rdz_w*hfx(i,j)/cpm
           ELSE
+             qfac = 1.+rvovrd*moist(i,kts,j,P_QV)
              d(k) = var_mix(i,kts,j)                    &
-                  - dt*rdz_w*nlflux_rho(i,k+1,j)        &
-                  + dt*rdz_w*((1.+rvovrd*moist(i,kts,j,P_QV))*hfx(i,j)/cpm+1.61*theta(i,kts,j)*qfx(i,j))
+                  - dt*rdz_w*nlflux_rho(i,k+1,j)*qfac   &
+                  + dt*rdz_w*(qfac*hfx(i,j)/cpm+1.61*theta(i,kts,j)*qfx(i,j))
           ENDIF
  
           DO  k = kts+1, ktf-1
             rdz_w = -g/dnw(k)/(c1h(k)*mu(i,j) + c2h(k))
+            IF(config_flags%use_theta_m == 1)THEN
+                qfac = 1.+rvovrd*moist(i,k,j,P_QV)
+            ELSE
+                qfac = 1.
+            ENDIF
             a(k)  = -rdz_w*rdz(i,k,j)*xkxavg(i,k,j)*dt
             b(k)  = 1.+rdz_w*(rdz(i,k+1,j)*xkxavg(i,k+1,j)+rdz(i,k,j)*xkxavg(i,k,j))*dt
             c(k)  = -rdz_w*rdz(i,k+1,j)*xkxavg(i,k+1,j)*dt
-            d(k)  = -rdz_w*(nlflux_rho(i,k+1,j)-nlflux_rho(i,k,j))*dt + var_mix(i,k,j)
+            d(k)  = -rdz_w*(nlflux_rho(i,k+1,j)-nlflux_rho(i,k,j))*qfac*dt + var_mix(i,k,j)
           ENDDO
  
             a(ktf) = 0.
--- a/dyn_em/module_diffusion_em.F
+++ b/dyn_em/module_diffusion_em.F
@@ -4278,6 +4278,10 @@ SUBROUTINE vertical_diffusion_2   ( ru_tendf, rv_tendf, rw_tendf, rt_tendf,   &
   CASE (0,2) ! with fixed surface heat flux given in the namelist
     heat_flux = config_flags%tke_heat_flux  ! constant heat flux value
                                             ! set in namelist.input
+    IF (config_flags%use_theta_m == 1)THEN
+      heat_flux = heat_flux * (1. + rvovrd * moist(i,kts,j,P_QV))
+    ENDIF
+
     DO j = j_start, j_end
     DO i = i_start, i_end
        cpm = cp * (1. + 0.8 * moist(i,kts,j,P_QV))
@@ -4297,6 +4301,10 @@ SUBROUTINE vertical_diffusion_2   ( ru_tendf, rv_tendf, rw_tendf, rt_tendf,   &
 
        cpm = cp * (1. + 0.8 * moist(i,kts,j,P_QV))
        heat_flux = hfx(i,j)/cpm
+       IF (config_flags%use_theta_m == 1)THEN
+         heat_flux = heat_flux * (1. + rvovrd * moist(i,kts,j,P_QV))
+       ENDIF
+
        rt_tendf(i,kts,j)=rt_tendf(i,kts,j)  &
             -g*heat_flux/dnw(kts)
        if(config_flags%use_theta_m == 1)THEN
@@ -8096,7 +8104,7 @@ END SUBROUTINE phy_bc
           ELSE
              d(k) = var_mix(i,kts,j)                    &
                   - dt*rdz_w*nlflux_rho(i,k+1,j)        &
-                  + dt*rdz_w*(hfx(i,j)/cpm+1.61*theta(i,kts,j)*qfx(i,j))
+                  + dt*rdz_w*((1.+rvovrd*moist(i,kts,j,P_QV))*hfx(i,j)/cpm+1.61*theta(i,kts,j)*qfx(i,j))
           ENDIF
  
           DO  k = kts+1, ktf-1
--- a/dyn_em/module_diffusion_em.F
+++ b/dyn_em/module_diffusion_em.F
@@ -4278,19 +4278,18 @@ SUBROUTINE vertical_diffusion_2   ( ru_tendf, rv_tendf, rw_tendf, rt_tendf,   &
   CASE (0,2) ! with fixed surface heat flux given in the namelist
     heat_flux = config_flags%tke_heat_flux  ! constant heat flux value
                                             ! set in namelist.input
-    IF (config_flags%use_theta_m == 1)THEN
-      heat_flux = heat_flux * (1. + rvovrd * moist(i,kts,j,P_QV))
-    ENDIF
 
     DO j = j_start, j_end
     DO i = i_start, i_end
        cpm = cp * (1. + 0.8 * moist(i,kts,j,P_QV))
        hfx(i,j)=heat_flux*cpm*rho(i,kts,j)         ! provided for output only
-       rt_tendf(i,kts,j)=rt_tendf(i,kts,j)  &
-             -g*heat_flux*rho(i,kts,j)/dnw(kts)
        if(config_flags%use_theta_m == 1)THEN
          rt_tendf(i,kts,j)=rt_tendf(i,kts,j)  &
+             -g*heat_flux*(1.+rvovrd*moist(i,kts,j,P_QV))*rho(i,kts,j)/dnw(kts) &
              -g*1.61*theta(i,kts,j)*qfx(i,j)/dnw(kts)
+       ELSE
+         rt_tendf(i,kts,j)=rt_tendf(i,kts,j)  &
+             -g*heat_flux*rho(i,kts,j)/dnw(kts)
        ENDIF
     ENDDO
     ENDDO
@@ -4301,15 +4300,14 @@ SUBROUTINE vertical_diffusion_2   ( ru_tendf, rv_tendf, rw_tendf, rt_tendf,   &
 
        cpm = cp * (1. + 0.8 * moist(i,kts,j,P_QV))
        heat_flux = hfx(i,j)/cpm
-       IF (config_flags%use_theta_m == 1)THEN
-         heat_flux = heat_flux * (1. + rvovrd * moist(i,kts,j,P_QV))
-       ENDIF
 
-       rt_tendf(i,kts,j)=rt_tendf(i,kts,j)  &
-            -g*heat_flux/dnw(kts)
        if(config_flags%use_theta_m == 1)THEN
          rt_tendf(i,kts,j)=rt_tendf(i,kts,j)  &
+             -g*heat_flux*(1.+rvovrd*moist(i,kts,j,P_QV))/dnw(kts) &
              -g*1.61*theta(i,kts,j)*qfx(i,j)/dnw(kts)
+       ELSE
+         rt_tendf(i,kts,j)=rt_tendf(i,kts,j)  &
+             -g*heat_flux/dnw(kts)
        ENDIF
 
     ENDDO


davegill:
@skamaroc @dudhia @weiwangncar 
Folks,
What is the best way for us to include Joe in this conversation? Does he do github?

weiwangncar:
@davegill Send him an email, and he can view the PR.

dudhia:
I checked and it is the same as Bill suggested.
This is a correct equation fix in my view. Will check code fix now.

dyn_em/module_diffusion_em.F
```diff
@@ -4278,6 +4278,10 @@ SUBROUTINE vertical_diffusion_2   ( ru_tendf, rv_tendf, rw_tendf, rt_tendf,   &
   CASE (0,2) ! with fixed surface heat flux given in the namelist
     heat_flux = config_flags%tke_heat_flux  ! constant heat flux value
                                             ! set in namelist.input
+    IF (config_flags%use_theta_m == 1)THEN
+      heat_flux = heat_flux * (1. + rvovrd * moist(i,kts,j,P_QV))
+    ENDIF
+
```
dudhia, line 4284: This is outside the i and j loop - needs fixing for case (0,2)

Matze77: > This is outside the i and j loop - needs fixing for case (0,2)

you're totally right. I fixed this issue and also rearranged the code a bit to make it more readable.

davegill:
@Matze77 @dudhia @weiwangncar 
Matthias,
You have added two IF tests inside of DO loops. Could you re-write the code to put the IF tests outside of the DO loops?

Matze77:
@davegill 
there is no additional if test, just two additional ELSE clauses compared with the original code, so I don't quite get what you mean. Can you elaborate?

dudhia:
This looks good to me. Especially appreciate hearing about the conservation aspect of this fix.
Someone should run isfflx=0 with a fixed heat flux and moisture case to fully check the change.
It also applies to the km_opt=5 3d tke scheme.

dyn_em/module_diffusion_em.F
```diff
@@ -4278,15 +4278,18 @@ SUBROUTINE vertical_diffusion_2   ( ru_tendf, rv_tendf, rw_tendf, rt_tendf,   &
...
     DO j = j_start, j_end
     DO i = i_start, i_end
        cpm = cp * (1. + 0.8 * moist(i,kts,j,P_QV))
        hfx(i,j)=heat_flux*cpm*rho(i,kts,j)         ! provided for output only
-       rt_tendf(i,kts,j)=rt_tendf(i,kts,j)  &
-             -g*heat_flux*rho(i,kts,j)/dnw(kts)
        if(config_flags%use_theta_m == 1)THEN
          rt_tendf(i,kts,j)=rt_tendf(i,kts,j)  &
+             -g*heat_flux*(1.+rvovrd*moist(i,kts,j,P_QV))*rho(i,kts,j)/dnw(kts) &
```
davegill, line 4288: @Matze77 
Matthias,
This was originally not written with performance in mind (an IF test inside of the DO loop). The IF and ELSE tests could be outside, and each could have a DO loop inside.

dudhia: I am not concerned with this because the bug-fix does not affect the existing IF structure. So I prefer it to be just left as a bug-fix with other issues separated.

davegill:
@dudhia @weiwangncar @mgduda 
Folks,
Take a look at the text of this PR commit message. I did not know that you can insert this text to manufacture equations on the fly?
```
https://latex.codecogs.com/gif.latex?1-\frac{1}{2}+\frac{1}{3}...
```
becomes

weiwangncar:
Pretty cool!

Matze77:
@dudhia
What do you mean with running a test case to fully check the change? What exactly do you want to have tested/checked?

dudhia:
It's only a formality, but since the change affects all the isfflx options
and you
only tested one, we would like to see that the results from the other
options are
also affected in the expected way which I fully expect to be true given the
code.
Then you can add in your description that all isfflx options were tested.
Thanks,
Jimy

Matze77:
So now I repeated the runs described in the initial PR message with isfflx in {1,2} (radiation scheme vs. specified heat flux} and km_opt in {2,5} once with the original code and once with the modified code.
org_hfx_2_None_3rd+5th
org_hfx_2_0 15_3rd+5th
org_hfx_5_None_3rd+5th
org_hfx_5_0 15_3rd+5th
fixed_hfx_2_None_3rd+5th
fixed_hfx_2_0 15_3rd+5th
fixed_hfx_5_None_3rd+5th
fixed_hfx_5_0 15_3rd+5th

Matze77:
For km_opt=5 and isfflx=2, there is still some inconsistency visible close to the surface. In my opinion, the non-local flux also needs to be converted to theta_m. I added the conversion factor in the latest commit. That fixes the issue as can be seen here:
fixed_hfx_nlflux_5_None_3rd+5th
fixed_hfx_nlflux_5_0 15_3rd+5th

Matze77:
For km_opt=5 and isfflx=2, there is still some inconsistency visible close to the surface. In my opinion, the non-local flux also needs to be converted to theta_m. I added the conversion factor in the latest commit. That fixes the issue as can be seen here:
fixed_hfx_nlflux_5_None_3rd+5th
fixed_hfx_nlflux_5_0 15_3rd+5th

dyn_em/module_diffusion_em.F
```diff
@@ -8044,6 +8050,9 @@ SUBROUTINE vertical_diffusion_implicit(ru_tendf, rv_tendf, rw_tendf, rt_tendf,&
      DO i = i_start, i_end
         nlflux_rho(i,k,j) = nlflux(i,k,j) * &
                            (fnm(k)*rho (i,k,j)+fnp(k)*rho (i,k-1,j))
+        IF(config_flags%use_theta_m == 1)THEN
+             nlflux_rho(i,k,j) = nlflux_rho(i,k,j)*(1.+rvovrd*moist(i,k,j,P_QV))
+        ENDIF
```
dudhia, line 8055: I am not sure about this. Check if this can be done instead with the line that sets sflux(i,j) based on hfx.

dudhia:
Thanks, very thorough. I left a comment on your code change for km_opt=5.

dyn_em/module_diffusion_em.F
Matze77: This should be equivalent. When looking at how the non-local flux is computed, you see that all terms that contribute are proportional to sflux (via the variables amf1, amf2, bmf2, hfxpbl):
https://github.com/wrf-model/WRF/blob/6d0090fe41bc4fc5a1746a420e825039c1d544b7/dyn_em/module_diffusion_em.F#L5206-L5226
I thought it would be more intuitive to view the non-local flux as being computed for dry theta and then convert it via the same equation derived in the initial PR message:

The water vapor flux term is dropped because there is no non-local water vapor flux in this scheme.

But if you still think it would be better/more readable to convert sflux instead, I can also do that.

dudhia:
What I noticed is that the nlflux is proportional to hfx so that is the term
to modify. Note that this term goes to zero when hfx is zero. This makes
it more like the other changes you already have made.

dyn_em/module_diffusion_em.F
dudhia: I added a general comment, but I will put one here too.
The nlflux terms are proportional to the surface heat flux, so it should be possible to just modify the surface heat flux the same way you did for other options. The difference from this code is that the correcting q term should be the surface one.

Matze77:
ok, I get your point that the change would then be more similar to my other changes. That would however change the grid variable nlflux (in contrast to my change, which is only local). I think it might be better if that potential output variable would be the same regardless of whether theta_m is used or not. In addition, grid%nlflux is also used in the subroutine tke_bouyancy. I think that in the bouyancy tke production term, the non-local heat flux should not include the moisture factor. If we already multiplied the term before, we would need to divide it here again, which is somehow inelegant. What do you think?

dudhia:
You have looked at this in more detail than I have. In that case
multiplying by the q term as you had it is OK, but make sure it is the
surface q not the k level value. Since this is nlflux_rho that avoids the
issue with the use of nlflux in the tke buoyancy.

Matze77:
ah, now I understand that there is a difference between the two approaches. You suggest using the surface q for the correction factor at all levels, while I implemented the k-level q.
In my opinion, taking the k-level q is correct. This modification is different from my other modifications because it applies to the non-local flux at all levels not only at the surface. So we cannot use an analogy argument.
I took again a look at the Zhang 2018 paper. They write the equations for the non-local flux of dry potential temperature. To convert this flux to moist potential temperature we need to multiply by a correction factor using the k-level q as in the equation that I wrote above:

 How the dry non-local flux is calculated, i.e. that is proportional to the surface heat flux, should be irrelevant for the conversion. 
We could also involve the developers of this package to come to a decision.

dudhia:
This term is tied to the surface flux. If that is zero this term goes away.
That is why it should
only be modified by the surface q because it represents a surface theta_m
flux. Internally
at other levels theta_m is the variable already, so no modification is
needed.

Matze77:
Why do you say, that internally at other levels theta_m is the variable already? In the non-local flux subroutine, I can't see any connection to theta_m, neither at the surface nor higher up. The routine uses the input variable th_phy, which is dry theta. 
The method uses a lot of constant parameters that are derived from a coarse-grained LES. These parameters are tied to the dry theta flux. If we just convert the surface flux, these parameters would not fit anymore. Instead, I suggest converting the whole non-local flux after computation, which requires the k-level q. But maybe we need a third opinion on this?

dudhia:
It's tricky and I think I have it right looking again at the code, but you
should check. The variable being mixed (var_mix) is thp which is theta_m,
not th_phy, so internal
fluxes are already coupled with the moisture term. The purpose of nlflux is
just to
modify the surface flux of theta_m.

matzegoebel:
When you say the variable being mixed is theta_m, you're talking about the local component of the scheme (down-gradient diffusion proportional to `xkxavg`) in the routine `vertical_diffusion_implicit`, right? That one is using theta_m, that's right. I was referring only to the routine `nonlocal_flux`, which does not include any connections to theta_m and thus needs to be converted.
The non-local flux is the second, independent component of the scheme and an important one at all vertical levels. It's proportional to the surface flux, but that's irrelevant for the conversion. It could be calculated in any conceivable way, the conversion should always be the same. The developers devised the non-local flux in a way that assumes dry theta to be the variable that is mixed. Just converting the surface flux is not the proper way of converting the non-local flux as a whole, in my opinion.
 It's a bit hard to see in the code, because it's an implicit algorithm, but the local and the non-local component are independent components that are treated separately. The former is already in the right form by using theta_m as the variable to be mixed, but the latter is not. At the surface both components are zero and the surface flux hfx is used which I already converted to theta_m in my initial commit.

dudhia:
The idea of the nonlocal flux is to represent surface-based thermals that
have a flux
in addition to the local diffusion-type flux. As it represents thermals
from the surface,
this flux should only depend on surface properties. It is like the gamma
term in the
YSU PBL scheme for example, or the Troen-Mahrt countergradient flux.

matzegoebel:
yes, that's right. But the parameters that are used to compute the vertical profile of the non-local flux based on the surface flux are tied to the dry theta flux (in the formulation of the authors) and would not be the same for theta_m, right?

dudhia:
The way I think of it, the use of theta_m as a variable only modifies what
we
consider the surface flux to be because now it has to include the moisture
source.
This is consistent with how we modify the surface flux for LES.

matzegoebel:
I'm afraid I still can't follow your argument. I suggest involving the developer(s) of the scheme. @zhangxsms, what do you think?

dudhia:
OK, I will email him. What we need is the formulation of the nonlinear
flux term as an equation. They may have a paper or presentation somewhere
too.

matzegoebel:
Thanks! Yes, this is the paper: https://journals.ametsoc.org/doi/10.1175/MWR-D-17-0356.1
equations 18, 23 and 24 refer to the non-local flux

dudhia:
More complex than I thought. I will study this.

dudhia:
The way I read it, the proportionalities are still to the surface flux.

dudhia:
@matzegoebel  - I did not hear back from the author. Do you have any more thoughts on this? I still see that all the terms are proportional to the surface flux so the q term should also be there.

matzegoebel:
no, no new thoughts, but I can try to restate my argument. I agree that the non-local flux is proportional to the surface heat flux, but I do not see why this matters for the conversion.
In my view, the computed non-local flux represents a part of the vertical **dry** theta flux at every level. For the conversion of this flux to **moist** theta, the details of the flux computation are irrelevant. The only thing that matters is that it corresponds to a dry theta flux. At every vertical level, we need to use the conversion factor that is appropriate at that level using q at that level.

If we only converted the surface heat flux by using the surface q for the conversion, the vertical profile of the flux would not be adequate for neither moist nor dry theta, because the profile calculation assumes a dry surface heat flux not a heat flux converted to moist theta.

dudhia:
This where the variable being mixed is important to know and it is the
moist theta already
if you trace var_mix back (I mentioned this in my August 26th message).
This make this diffusion
routine similar to the previous one you have dealt with by just modifying
the surface flux.

matzegoebel:
yes, var_mix corresponds to theta_m. But that variable is only used in the local part (Deardorff type) of the scheme and has nothing to do with the non-local part. The non-local flux is computed separately and its divergence is added as a source term to the implicit solution in line 8107:
https://github.com/wrf-model/WRF/blob/f311cd5e136631ebf3ebaa02b4b7be3816ed171f/dyn_em/module_diffusion_em.F#L8102-L8108
In my opinion, the var_mix in line 8107 belongs to the local part of the scheme. So maybe that's confusing, because it stands in the same line as the non-local source term. The non-local flux in line 8107 nevertheless corresponds to a dry theta flux at all vertical levels.

dudhia:
I think the question is about line 8054. If that can be kts in the q term
modifying hfx with the kts q term before everything allows removing the
changes in 8054 and 8107.

matzegoebel:
I don't think so, because the change in my line 8054 only affects the non-local flux which is only taken into account for k > kts, while the change in line 8107 converts the surface heat flux, so for k=kts. These are two separate issues.

dudhia:
This is where I was suggesting that the nonlocal flux is proportional to
the surface flux as represented by Fig. 2 in the paper
so it should only be a kts term there.

serafis3:
Dear all, I'm the PI of @matzegoebel's project in Innsbruck and I'm jumping in because we had a few discussions about this issue recently. I don't have the time to dig deep into the WRF code, but I'm hoping to help with some considerations on general principles. So...

(1) At any vertical level k, the following flux decomposition holds true (neglecting third-order moments):


(2) If now we want the turbulent flux to be parameterized with a local plus a non-local component, we simply add the non-local fluxes where appropriate on the RHS:


(3) The above can be restated as:


I guess this is what @matzegoebel suggested back on 12 August. I interpret the equation as follows: Regardless of how the non-local contributions to turbulent flux are modelled, consistency with the local tendency of theta_m requires them to be weighted with the local values of theta and q. Does this sound convincing?

I do see the uncomfortable aspect of the thing: Even if the non-local term remains proportional to the surface fluxes, now its impact is not constant throughout the BL. This is a side-effect of reconstructing the fluxes of theta_m from those of theta and q.

dudhia:
Thanks for entering this discussion. The changes by Matthias are important
and we appreciate
his adding them. The other diffusion options are certainly helped by what
he is doing and our
only sticking point is km_opt=5.
I will try to bring the developer into this discussion @zhangxsms
<https://github.com/zhangxsms> . He is at the Shanghai Met Service in China.
As I mentioned my view is that the surface proportionality is intended
because this nonlocal
terms represents a flux that originates at the surface.
Jimy

serafis3:
Thanks for getting back! Sure, there is no doubt about the purpose and origin of the nonlocal terms. To clarify my thought: In Eq (3) above, the nonlocal fluxes of (dry) theta and q (if existing) remain just as they are in the Shanghai Met Service scheme (kmopt=5). But there is an open issue with theta_m conservation. A closure such as:

https://latex.codecogs.com/gif.latex?\overline{w'\theta'}=-K\frac{\partial\overline{\theta}}{\partial{z}}+\overline{w'\theta'}_{NL}

is fine with theta in a dry atmosphere. But if one wishes to conserve theta_m in a moist atmosphere, it seems the closure should become:

https://latex.codecogs.com/gif.latex?\overline{w'\theta'_m}=-K\frac{\partial\overline{\theta}_m}{\partial{z}}+\overline{w'\theta'_m}_{NL}

which expands to:

https://latex.codecogs.com/gif.latex?\overline{w'\theta'_m}=-K\frac{\partial\overline{\theta}_m}{\partial{z}}+\overline{w'\theta'}_{NL}(1+\alpha\overline{q})+\alpha\overline{\theta}\overline{w'q'}_{NL}

This doesn't change the parameterization of the nonlocal fluxes at all, it only adds two terms to account for the influence of moisture on theta_m. If there is no need to conserve theta_m, those two terms can be left out... it can be a legitimate modelling choice, if the consequences are clear.

I'm not sure if the authors of the SMS scheme considered the case of moist air. It doesn't seem so. Yes, getting their feedback would be precious!

Stefano

PS: For unknown reasons, it seems equations won't be displayed in this comment. Sorry about that. Just click on the link to see the gif...

matzegoebel:
@serafis3, thanks for your support. I totally agree. 
In order to make equations be displayed, the URL has to be embedded in an image command like this:
``

dudhia:
Can we take this conversation off github? I am cc'ing my email address.
Future messages can exclude the github accounts. Please respond only
to my email address, so that I can have yours.
Thanks,
Jimy

matzegoebel:
I just pushed commit 654069f3455ffbc62d7301de312ab56d04a46db2 which leads to an even better conservation of dry theta. Instead of multiplying the non-local flux in km_opt=5 directly with q(k), the vertical derivative of nlflux is first calculated and then multiplied with q(k). This is more consistent, because nlflux is on w-levels, while q is on mass levels, as is the vertical derivative of nlflux. I did a few simple CBL simulations to show the difference between not modifying the non-local flux at all, using q(k) in the conversion factor, and using q(kts) in the conversion factor, as Jimy suggests. The following figure shows the RMSE of comparing the sum of all forcings with the actual dry theta tendency. Each lines represents an average over 10 minutes, x, y and over 4 ensemble simulations with perturbed initial conditions.

rmse_nlflux_conversion_tmean

Using q(k) in the conversion, seems to lead to a slightly lower error at most levels, especially at the beginning of the simulation,
The results at later times need be taken with care, since the different simulations are not exactly equivalent due to the different conversion factors. Therefore, I argue that the q(k) version should be the correct one (in addition to the theoretical arguments that we presented earlier).
In any way, both versions are much better than not modifying the non-local flux (the green line in the left plot actually extends to 5e-5, thus an order of magnitude worse).

dudhia:
OK, I think mine only differs from yours in using q(kts) in that line, just
to summarize the difference
for Xu,
Jimy

matzegoebel:
yes that's right, in line 8110

davegill:
@dudhia @weiwangncar 
What is the status of this PR? Are we waiting for external comments from anyone?

dudhia:
I am still trying to make my case.
At least Xu Zhang is now involved in the discussion.
Jimy

dudhia:
@matzegoebel I want to finalize for 4.2.2 bug-fix which is about ready.
Can we do it my way for km_opt=5? Which is to use kts for qv. Thanks.

dudhia:
In fact, it seems like it is already this way, so I will go ahead and approve soon after you confirm it is ready now.

dudhia (approved):

matzegoebel:
ok

dudhia:
It is now approved and can be merged and committed. When doing
it make sure the commit message is not overwritten by something
automatically changed.
Or we can do the merge if you prefer.

davegill:
@matzegoebel @dudhia 
Jimy and Matthias,
We will do the merge.

matzegoebel:
I just see that I have a bunch of equations and plots in the commit message. Should remove those?

davegill:
@matzegoebel 
If those equations and figures are still valid, they can certainly remain.

dudhia:
It worked, but now I realize the change I wanted, k-> kts, is not in there
yet.
We also had Xu Zhang suggest he preferred it to be in.
I will set up a separate PR,
Jimy

matzegoebel:
> It worked, but now I realize the change I wanted, k-> kts, is not in there yet. 

Oh yes you're right. Sorry, I overlooked this.

> We also had Xu Zhang suggest he preferred it to be in. 

He changed his mind several times during the email discussion, but in his latest mail on October 26th he wrote "Hi Matthias, I totally agree with your opinion in the view of theory." This refers to the discussion we had about whether to use q(k) or q(kts). So I guess he actually supports my view. But we can also ask him again for a final statement, if you're not convinced.

dudhia:
I am still convinced that this change being to the surface flux has to
apply to kts.
Since this added fix is under my name, I can be blamed if it is wrong.

PR #1291, opened 2020-09-23
Prevent stack smashing caused by incompatible namelist options

lpilz:
TYPE: bug fix

DESCRIPTION OF CHANGES:
Problem:
A namelist option incompatibility between chem_opt (16) and gas_drydep_opt (1) leads to stack smashing.
When choosing a tracer-only chem_opt, numgas is 0.
If gas_drydep_opt is then 1 (as is default), in the Wesely scheme initialization (dep_init), the field dvj is initialized with size numgas (here 0).
This leads to stack smashing upon further accesses.

Solution:

LIST OF MODIFIED FILES: chem/chemics_init.F

TESTS CONDUCTED:

RELEASE NOTE: Added fatal error for incompatible chem_opt and gas_drydep_opt namelist options.

### DIFF
--- a/chem/chemics_init.F
+++ b/chem/chemics_init.F
@@ -1944,6 +1944,10 @@ ENDIF ghg_block
 !
    drydep_select: SELECT CASE(config_flags%gas_drydep_opt)
      CASE (WESELY)
+        IF (numgas .eq. 0) THEN
+            CALL wrf_error_fatal("ERROR: numgas = 0, SELECTED CHEM OPT IS &
+            &NOT COMPATIBLE WITH WESELY DRY DEPOSITION")
+        ENDIF
        CALL wrf_debug(15,'initializing dry dep (wesely)')
 
         call dep_init( id, config_flags, numgas, mminlu_loc, &


jordanschnell (approved):
Approved by chem

davegill:
@lpilz @jordanschnell 
There are a couple of items:
1. This PR is really a bug fix, and could therefore use the base branch "release-v4.2.2". This PR currently uses the base branch "develop". The impact is that this bug fix is not sen by users until April 2021. I tried a simple "change the base branch", but that caused mods to 28 files. I have reverted that change back to the original PR. So, we can keep the PR as-is, which means no bug-fix for users for 6 months or so. Alternatively, this PR is closed, and a new PR is submitted. That would use the base branch as "release-4.2.2". The identical PR message could be just copied into here. I am assuming that nearly identical mods to chem_init would also be OK.
2. Since these are namelist considerations, perhaps the better place to test the inconsistent use of these options is in the share/module_check_a_mundo.F file. The entire purpose for that file is to process namelist settings prior to any other input. There would likely need to be inside of an ifdef for WRF_CHEM, but those exist in that file. Follow the templates. For example, we do not want you to issue the wrf_error_fatal call, just augment an error counter. Here is an existing example.
```
   IF ( model_config_rec % perturb_chem_bdy .EQ. 1 ) then

#if (WRF_CHEM != 1)
      wrf_err_message = '--- ERROR: This option is only for WRF_CHEM.'
         CALL wrf_debug ( 0, TRIM( wrf_err_message ) )
         count_fatal_error = count_fatal_error + 1
#endif
```
Your entire IF test would go inside of the ifdef. You may also need to consider logic concerning use of the chem options on multiple domains. Also in that file are examples with this structure (that max_dom DO loop is checking each "column" of the namelist):
```
  do i=1,model_config_rec%max_dom
    if ( model_config_rec%afwa_diag_opt(i)   .EQ. 1 ) then
      model_config_rec%afwa_diag_opt(:)   = 1
      exit
    endif
  enddo
```

jordanschnell:
@lpilz I agree with @davegill on this - Dave, thanks for the suggestion regarding check_a_mundo. 

@lpilz Can you implement this change? And including the loop over the domains. We can also get this pushed through with release-4.2.2

lpilz:
@davegill Cheers for the thorough review :)
1. I will open a new PR based on the release branch. I think we'd want this pushed out asap. Sorry for that I didn't know your release schedule.

2. I'm not quite sure I agree. If I were to implement this in `share/module_check_a_mundo.F`, then I'd either: 1. have to hardcode the chem options for which `get_last_gas` returns 0 (which I find ugly and prone to errors upon updates) or 2. include `module_input_chem_data` and `get_last_gas` and check whether in this run with `gas_drydep_opt=1`, `get_last_gas` returns 0. However, I don't know if you want dependencies from `module_input_chem_data` into this file.

I agree, that it breaks a little late but the beauty of this solution is that you catch the problem itself (numgas=0 in dep_init) and not a correlate to it (some namelist options being set). However, if you really want me to, I will implement your proposed fix.

davegill:
@lpilz 
Thanks for the more complete explanation.
1. I would suggest that you close this PR and open a new one with the release-v4.2.2 branch.
2. I agree with your reasoning for NOT putting the code into check_a_mundo. Please include that reasoning (which is in the comment above) in your follow-on PR description for completeness.

PR #1294, opened 2020-09-24
Chem: Prevent seg fault due to numgas==0

lpilz:
TYPE: bug fix

DESCRIPTION OF CHANGES:
Problem:
A namelist option incompatibility between chem_opt (16) and gas_drydep_opt (1) leads to trying to access a variable
that is not available in memory. This leads to unpredictable behaviors. When choosing a tracer-only chem_opt, numgas 
is 0. If gas_drydep_opt is 1 (the default), in the Wesely scheme initialization (dep_init), the field `dvj` is initialized with 
size numgas (here 0). Accessing the variable `dvj` is an error.

Solution:

This fix is in `chem/chemics_init.F` and not `share/module_check_a_mundo.F` because checking the namelist options 
would require some tricky-to-maintain hardcoding.

LIST OF MODIFIED FILES: 
chem/chemics_init.F

TESTS CONDUCTED:
3. The jenkins testing is OK.

RELEASE NOTE: For WRF-Chem, a namelist option incompatibility between chem_opt (16) and gas_drydep_opt (1) leads to trying to access a variable that is not available in memory. This leads to unpredictable behaviors. A fatal error was added for when the Wesely scheme initialization is called with numgas = 0.

### DIFF
--- a/chem/chemics_init.F
+++ b/chem/chemics_init.F
@@ -1948,7 +1948,7 @@ ENDIF ghg_block
             CALL wrf_error_fatal("ERROR: numgas = 0, SELECTED CHEM OPT IS &
             &NOT COMPATIBLE WITH WESELY DRY DEPOSITION")
         ENDIF
-        CALL wrf_debug(15, "initializing dry dep (wesely)")
+        CALL wrf_debug(15,'initializing dry dep (wesely)')
 
         call dep_init( id, config_flags, numgas, mminlu_loc, &
                       its, ite, jts, jte, ide, jde )
--- a/chem/chemics_init.F
+++ b/chem/chemics_init.F
@@ -1948,7 +1948,7 @@ ENDIF ghg_block
             CALL wrf_error_fatal("ERROR: numgas = 0, SELECTED CHEM OPT IS &
             &NOT COMPATIBLE WITH WESELY DRY DEPOSITION")
         ENDIF
-        CALL wrf_debug(15,'initializing dry dep (wesely)') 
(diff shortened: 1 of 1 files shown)


davegill:
@lpilz 
Lukas,
1. Can we add a better title? It needs to refer to WRF Chem. Also, is only Wesely dry dep a problem with numgas=0? If so, let's include that focused information in the PR title. Users will want to know if they need this update. The PR title should let them know if this important for their research.

2. Previously, when the wrong namelist options were set, what happened? Did the model fail? Did the model run to completion, but with bad results?

jordanschnell:
@davegill I can review/approve this, but what is happening with the jenkins test?

lpilz:
Hi @davegill 
1. How about "WRF-Chem: Prevent stack smashing due to numgas==0"
2. I checked around a bit and found instanciations with numgas in `wesely_driver`, `rc`, and `HL_init`. `wesely_driver` is called in the case we prevent, `dep_init` is, `rc` is only called inside of `wesely_driver` and as far as I see, `HL_init` is never called. I'll check some more but for now, I think this covers it.

lpilz:
@davegill @jordanschnell 

What's the latest update on this PR? Is it ready to be merged?

jordanschnell:
@lpilz We were waiting on the automated tests to be back up but it seems the check has passed. @davegill, OK to approve and merge?

davegill:
@lpilz @jordanschnell 
Folks,
Yes, we are all good with this PR.

jordanschnell (approved):
Approved by chem

PR #1295, opened 2020-09-28
Background effective radius consistency fix

juhyejuhye:
FARMS correctly detects unresolved clouds

TYPE: bug fix

KEYWORDS: FARMS, effective radius of unresolved hydrometeors, cloud fraction

SOURCE: Ju-Hye Kim and Pedro A. Jimenez (NCAR/RAL), and Jimy Dudhia (NCAR/MMM)

DESCRIPTION OF CHANGES:
Problem:
The previous version of FARMS did not correctly account for the radiative effects of unresolved clouds, because the FARMS calculates the surface irradiances using both cloud water contents and cloud effective radii. In previous version, cloud water contents came from both resolved and unresolved clouds, on the other hand, effective radii came only from resolved clouds. 

Solution:
The climatological effective radii are assigned to unresolved clouds of liquid, ice and snow species, and are specified as the effective radii of clouds in microphysics schemes (e.g. mp=2) that do not provide any effective radii. The fix is also valid for Thompson MPs, WSMMPs, and WDMMPs by adjusting the background effective radii of these schemes.  In addition to this, we improved calculations of surface irradiance by adding cloud fraction to the FARMS scheme.

ISSUE: N/A

LIST OF MODIFIED FILES: 
M       phys/module_mp_wdm5.F
M       phys/module_mp_wdm6.F
M       phys/module_mp_wdm7.F
M       phys/module_mp_wsm3.F
M       phys/module_mp_wsm5.F
M       phys/module_mp_wsm6.F
M       phys/module_mp_wsm7.F

TESTS CONDUCTED: 

Yes. The downward short wave flux (GHI) will increase in the cloud region with unresolved clouds by reduced scatterings due to increased cloud size. This was tested in the WRF-Solar model.

We did't do the Jenkins tests.

RELEASE NOTE: Bug fix to account for the effective radius of the unresolved hydrometeors by FARMS.


davegill:
@juhyejuhye @dudhia @weiwangncar 
Folks,
There are IDENTICAL settings in the modified MP schemes to set the effective radii. These constants should probably be set once in share/module_model_constants.F, and then assigned in each of the modified MP schemes.

weiwangncar:
@juhyejuhye Does change the default constants change the results WITHOUT FARMS?

juhyejuhye:
Wei,
I've checked SWDOWN before (This is from RRTMG) and it showed some changes
in the WRF-Solar simulation results. I'll test this WRF PR version and let
you know. Then is there a variable you want to check?

Thanks,
Ju-Hye

weiwangncar:
@juhyejuhye Do you expect results to be different without FARMS? If not, we should check for bit-for-bit results using diffwrf utility in external/io_netcdf/. Run WRF with previous constants and new and use diffwrf to compare the results: diffwrf file1 file2. And what's reason to change the constants, some of them only slightly (e.g. 2.49e-6 to 2.51e-6)?

juhyejuhye:
Wei,

We expect the results will be different.

The changes in minimum values are very small for Thompson, so it brings a
very small difference in the output as well. However, for the WSM and WDM
MPs family, the minimum re_qi changed from 10.01e-6 to 4.99e-6 and re_qs
changed from 25.e-6 to 9.99e-6, so the effect of the changed minimum on the
results is bigger. We were with Jimy on this test and he agreed that this
is a positive direction for the results of the WSM WDM MPs as well.


Thanks,

Ju-Hye

weiwangncar:
@juhyejuhye OK. Can you then show some of the differences from the runs?

juhyejuhye:
Wei,

I’ll compare results using this PR version and send you.


Thanks,

Ju-Hye

weiwangncar:
@juhyejuhye Thanks. Do we need to get the developers for WSM/WDM to agree on the change?

juhyejuhye:
Wei,

These are plots comparing GHIs from RRTMG. The first figure is a comparison
of the domain averaged GHIs of previous Thompson and WSM6 codes with the
Thompson and WSM6 codes with the minimum cloud effective raii changed, and
the other is the differences. GHI has not changed much for Thompson scheme,
but as I explained, it has changed a lot more in WSM6.


[image: mean_swdown_cycle.png]
[image: diff_mean_swdown_cycle.png]


And, I'll talk with Jimy and Pedro if we should talk to the WSM/WDM
developers on this issue.

Thanks,
Ju-Hye

weiwangncar:
@juhyejuhye Thanks. But your figures were not posted. Can you try again?

juhyejuhye:
Wei, I'll try one more time. I'm attaching the files.

Ju-Hye

juhyejuhye:
1_mean_swdown
2_diff_mean_swdown

Can you see?
Ju-Hye

weiwangncar:
@juhyejuhye Yes, I can see now. Thanks.

juhyejuhye:
Dave,
We updated the code to share the background effective radii of clouds in
'module_model_constats.F' for WSM/WDM and Thompson microphysics schemes.
Please let me know if there is anything strange.

Wei,
We talked about the changes related to the effective radii of clouds with
the WSM/WDM MPs group and they agreed to this change. One thing to clarify
is that what we have done is that we assigned the background effective
radii values differently from the minimum radii values of clouds in WSM/WDM
MPs. The background effective radius and the minimum effective radius were
not exactly distinguished in WSM/WDM MPs before. (The background effective
radius indicates the size of effective radius when there is no cloud.) We
only changed the background effective radius values. And inside of
subroutine effecRad_wsm6 remains unchanged, so the minimum size of clouds
has not changed when there is a cloud.

Thanks,
Ju-Hye

weiwangncar:
@juhyejuhye Thanks for addressing my questions. I'm ok with this PR. Could you add a comment in the PR (if it is appropriate) to state that it will have up to 30 (or another number) Wm-2 impact on the clear sky GHI for the  WSM/WDM schemes. I have to say I'm still not clear about this: the background effective radii have an impact on radiation flux. So simply by turning on this option we get less radiation flux even without clouds? What does the background value represent?

dudhia:
The background value determines whether RRTMG uses the radius from the
microphysics or its own climatology. This usually should happen with clouds
from other physics, but not when the microphysics is active. If the
microphysics
provides a radius less than the background, RRTMG does its own climatology
which this fix addresses by making sure it is larger than the background
value
when there is a cloud.

weiwangncar:
@dudhia So you're saying that by reducing the background radius values for these species, the effective radii from a microphysics would be used more than the RRTMG climatology, and hence the differences in radiation fluxes.

dudhia:
Yes, for example 2.51 would not be detected by RRTMG as a no microphysics
cloud case, but the new 2.49 will
because RRTMG uses 2.5. I think this only matters when other physics
produces the clouds and not microphysics.
Jimy

weiwangncar:
@dudhia Thanks for the explanation. I'm less confused now...

juhyejuhye:
Thanks, Jimy and Wei!

It's good to summarize this together.


What I understand is when there is no cloud from MPs and yes cloud from
other physics, Re_QC/RE_QI/RE_QS were 2.51,10.01, 25 micron (in this case
RRTMG does not use their own values, just use 2.51, 10.01, and 25 micron)
in previous code. However, as a result of our modifications, those values
have been changed to 2.49, 4.99, and 9.99 micron from microphysics, so
RRTMG WILL use their own climatology values in this case (no clouds in MPs
and yes clouds in physics). And it brings a difference in SWDOWN.


Thanks,

Ju-Hye

dudhia:
Correct.

weiwangncar:
@juhyejuhye Since these are not just fixes for FARMS, the title of this PR should be updated to reflect the actual change.

dudhia:
A suggested title may be
Background effective radius consistency fix

juhyejuhye:
Wei,

I agree to change by 'Background effective radius consistency fix'.
Then how can I change the title?

Thanks,
Ju-Hye

dudhia:
OK, I did.

dudhia (approved):

PR #1296, opened 2020-09-29
Prevent stack smashing caused by incompatible namelist options - 2

davegill:

DESCRIPTION OF CHANGES:
Problem:
A namelist option incompatibility between chem_opt (16) and gas_drydep_opt (1) leads to stack smashing.
When choosing a tracer-only chem_opt, numgas is 0.
If gas_drydep_opt is then 1 (as is default), in the Wesely scheme initialization (dep_init), the
field dvj is initialized with size numgas (here 0).  This leads to stack smashing upon further accesses.

Solution:

This fix is in `chem/chemics_init.F` and not `share/module_check_a_mundo.F` because checking the namelist options
would:
1. either require hardcoding, or
2. create a new dependency of `module_input_chem_data` into `check_a_mundo`.

This  solution on hand (mods to chemics_init) adresses the issue itself (numgas being 0 in dep_init).

LIST OF MODIFIED FILES:
chem/chemics_init.F

TESTS CONDUCTED:

### DIFF
--- a/chem/chemics_init.F
+++ b/chem/chemics_init.F
@@ -1944,7 +1944,11 @@ ENDIF ghg_block
 !
    drydep_select: SELECT CASE(config_flags%gas_drydep_opt)
      CASE (WESELY)
-       CALL wrf_debug(15,'initializing dry dep (wesely)')
+        IF (numgas .eq. 0) THEN
+            CALL wrf_error_fatal("ERROR: numgas = 0, SELECTED CHEM OPT IS &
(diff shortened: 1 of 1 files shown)


davegill:
@jordanschnell @lpilz
Folks,
I think that something is wrong with the regression testing. I am basically re-doing this exact PR to see what happens.

davegill:
@jordanschnell 
Please hold off for a few days. I will give the all clear.

davegill:
This is just a test PR.
The real PR is "Chem: Prevent seg fault due to numgas==0" #1294

PR #1268, opened 2020-08-12
Fix MPI synchronization in real.exe

honnorat:
Fixes MPI synchronization bug in real.exe  #1267

TYPE: [bug fix]

KEYWORDS: mpi, real, bug

SOURCE: "Marc Honnorat (EXWEXs)"

DESCRIPTION OF CHANGES:
Problem:
When running real.exe on multiple processes with MPI, one or more process occasionnaly crashes in `setup_physics_suite` (in `share/module_check_a_mundo.F`). This may be due to the fact that `wrf_dm_initialize` is apparently non-blocking from an MPI point of view. See #1267 for an example.

Solution:
An MPI barrier is added just after the call to `wrf_dm_initialize`  to force all processes be synced before checking namelist consistancy.

ISSUE:
Fixes #1267

LIST OF MODIFIED FILES: 
- M       main/real_em.F

TESTS CONDUCTED: 
On the only machine were I have seen the bug occur, this change fixes the problem. No other test was conducted since I couldn't reproduce the bug on another setup.

PR #1299, opened 2020-10-12
Fix MPI synchronization in real

### LINKED ISSUE #1267: Irregular MPI synchronization bug in real.exe
**Describe the bug**
When running real.exe on multiple processes with MPI, one or more process occasionnaly crashes in `setup_physics_suite` (in `share/module_check_a_mundo.F`).

**To Reproduce**
The bug is difficult to reproduce since it occurs randomly in any of the following situations. I've tried two versions of WRF (v4.1.5 and v4.2.1) and several versions of `ifort` compiler (Intel Parallel Studio 2017 and 2018) with Intel MPI library, either on a single node (shared-memory `shm` network fabric) or on two nodes with InfinyBand, with or without optimizations.

**Expected behavior**
The output of `rsl.error.0000` should be like:
```
taskid: 0 hostname: node377.exwexs.fr
 module_io_quilt_old.F        2931 T
 Ntasks in X            6 , ntasks in Y            8
  Domain # 1: dx =  8100.000 m
  Domain # 2: dx =  2700.000 m
REAL_EM V4.2.1 PREPROCESSOR
[...]
d01 2016-12-19_00:00:00 real_em: SUCCESS COMPLETE REAL_EM INIT
```

**Screenshots**
However, one random process sometimes crashes with:
```
$ cat -v rsl.error.0020
 taskid: 20 hostname: node172.exwexs.fr
 module_io_quilt_old.F        2931 T
 Ntasks in X            6 , ntasks in Y            8
Configuring physics suite '^@^@^@^@^@^@^@^[CUT]@^@^@^@^@^@^@^@^@^@^@'

FATAL CALLED FROM FILE:  <stdin>  LINE:    2529
Unrecognized physics suite
application called MPI_Abort(MPI_COMM_WORLD, 1) - process 20
```

**Additional context**
The `physics_suite` parameter seems wrong for the faulty processes. However, the file `namelist.output` is OK.

honnorat:
TYPE: bug fix

KEYWORDS: mpi, real, bug

SOURCE: Marc Honnorat (EXWEXs)

DESCRIPTION OF CHANGES:
Problem:
When running real.exe on multiple processes with MPI, one or more process occasionally crashes in setup_physics_suite 
(in share/module_check_a_mundo.F). This has been linked to wrf_dm_initialize non-blocking MPI. 

The real.exe occasionally crashes in setup_physics_suite (in share/module_check_a_mundo.F#L2640) because the latter 
uses model_config_rec % physics_suite, which on some machines is not initialized. The behavior is as if the broadcast of 
model_config_rec performed just before in main/real_em.F#L124 had not been received by all processes.

I had never seen this bug before, it has only happened on one machine (an Intel-based cluster using Intel-MPI and ifort).

The current fix makes sure that all processes are well synced before proceeding with setup_physics_suite. It solves the 
issue on my machine. Since this is immediately after reading in the namelist, no performance issues are expected as this 
read and broadcast of the namelist occurs only once per real / WRF / ndown run.

Solution:
An MPI barrier is added at the end of wrf_dm_initialize to force all of the processes to be synchronized before checking 
the namelist consistency.

This is a simplification of PR #1268, which had extra white space.

ISSUE:
Fixes #1267

LIST OF MODIFIED FILES:

TESTS CONDUCTED:
1. On the only machine were I have seen the bug occur, this change fixes the problem. No other test was conducted since I couldn't reproduce the bug on another setup.
2. Jenkins testing is all PASS.

RELEASE NOTES: When running real.exe on multiple processes with MPI, one or more processes occasionally crash in setup_physics_suite (in share/module_check_a_mundo.F). This has been traced to the fact that wrf_dm_initialize is non-blocking from an MPI point of view.  The problem is intermittent and has only happened on one machine (an Intel-based cluster using Intel-MPI and ifort). An MPI barrier has been added at the end of wrf_dm_initialize to force all processes to be synchronized before checking namelist consistency.

### DIFF
--- a/external/RSL_LITE/module_dm.F
+++ b/external/RSL_LITE/module_dm.F
@@ -228,6 +228,9 @@ CONTAINS
       CALL nl_set_nproc_y ( 1, ntasks_y )
       WRITE( wrf_err_message , * )'Ntasks in X ',ntasks_x,', ntasks in Y ',ntasks_y
       CALL wrf_message( wrf_err_message )
+#ifndef STUBMPI
+      CALL MPI_BARRIER( local_communicator, ierr )
+#endif
       RETURN
    END SUBROUTINE wrf_dm_initialize
 
--- a/external/RSL_LITE/module_dm.F
+++ b/external/RSL_LITE/module_dm.F
@@ -228,7 +228,7 @@ CONTAINS
       CALL nl_set_nproc_y ( 1, ntasks_y )
       WRITE( wrf_err_message , * )'Ntasks in X ',ntasks_x,', ntasks in Y ',ntasks_y
       CALL wrf_message( wrf_err_message )
-#ifndef STUBMPI
+#if ( defined( DM_PARALLEL ) && ( ! defined( STUBMPI ) ) )
       CALL MPI_BARRIER( local_communicator, ierr )
 #endif
       RETURN


davegill:
@honnorat 
Marc,
I don't have permission to try a mod:
Use this instead:

#if ( defined( DM_PARALLEL ) && ( ! defined( STUBMPI ) ) )

honnorat:
Done. But this is not necessary since every call to `wrf_dm_initialize` is already fenced this way. If we had to take this path, we would have to do the same for each of the 55 other occurences of `#ifndef STUBMPI` in `module_dm.F`.

honnorat:
> Did you receive an email from jenkins?

After the first commit, yes. Everything seemed ok.

davegill:
@honnorat 
Marc, 
Could you put that email text in one of these comments? Thanks

davegill:
```
Please find result of the WRF regression test cases in the attachment. This build is for Commit ID: aef012df1aab738aeb9805dd1c6afaf305e307f9, requested by: honnorat for PR: https://github.com/wrf-model/WRF/pull/1299. For any query please send e-mail to David Gill.

    Failed Simulations are: 
    None
    None
```

honnorat:
From the last commit:  
[wrf_output.zip](https://github.com/wrf-model/WRF/files/5372114/wrf_output.zip)

```
Please find result of the WRF regression test cases in the attachment.
This build is for Commit ID: aef012df1aab738aeb9805dd1c6afaf305e307f9,
requested by: honnorat for PR: https://github.com/wrf-model/WRF/pull/1299.
For any query please send e-mail to David Gill.

    Failed Simulations are: 
    None
    None
```

davegill (approved):

davegill:
@honnorat 
Marc,
Thanks for this contribution! Welcome aboard the WRF developer's train.

PR #1196, opened 2020-05-16
Add multilayer UCMs to YSU PBL parameterization

davegill:
TYPE: enhancement

KEYWORDS: BEP, BEP+BEM, YSU PBL scheme, vertical diffusion solvers

SOURCE: Eric A. Hendricks (NCAR/RAL/NSAP)

DESCRIPTION OF CHANGES: 
(Replaces #1195 Add multilayer UCMs to YSU PBL parameterization)   
Physics enhancement where multilayer BEP and BEP+BEM UCMs are added to the YSU PBL scheme. 1. The implicit (A) and explicit (B) BEP and BEP+BEM source terms are added as forcing to the YSU implicit vertical diffusion equation solvers. Forcing is applied for zonal momentum, meridional momentum, potential temperature, and water vapor mixing ratio. The A sources are added to the tridiagonal matrix and the B sources are added to the explicit term vectors. 
2. The finite differencing is modified using the BEP volume fraction and surface fraction not occupied by buildings. 
3. The BEP and BEP+BEM TKE tendency terms are added to the diagnostic TKE in YSU when topographic drag is turned on.

References:
Martilli et al. (2009), description of the modifications made in WRF.3.1 and short user's manual of BEP, NCAR technical note.   
Hendricks, E. A., J. C. Knievel, and Y. Wang, 2020: Addition of multilayer urban canopy models to a nonlocal planetary boundary layer parameterization and evaluation using ideal and real cases, J. Appl. Met. Clim., submitted.

LIST OF MODIFIED FILES: 
M       phys/Makefile
A        phys/module_bl_ysuurb.F
M       phys/module_pbl_driver.F
M       phys/module_physics_init.F

TESTS CONDUCTED: 
Real-case 48-h simulations of Houston urban area were conducted (10/05/2017-10/07/2017) with an innermost nest grid spacing of 1 km. The following test suite was set up based upon discussions with Mike Barlage:

Baseline tests (version before physics enhancement):
1)	MYJ BEP
2)	MYJ BEP+BEM 
3)	YSU BULK 
4)	YSU SLUCM
New code tests (new version with enhancement):
5)	YSU BULK
6)	YSU SLUCM
7)	YSU BEP
8)	YSU BEP+BEM   
9)	MYJ BEP
10)	MYJ BEP+BEM 

[...]

PR #1305, opened 2020-10-26
Add multilayer UCMs to YSU PBL parameterization 

eahendricks:
TYPE: Enhancement

KEYWORDS: BEP, BEP+BEM, YSU PBL scheme, vertical diffusion equations

SOURCE: Eric A. Hendricks (NCAR/RAL/NSAP)

DESCRIPTION OF CHANGES:
The multilayer BEP (Building Effects Parameterization) and BEP+BEM (BEP with the integration of a Building Energy Model) urban canopy models (UCMs) are added to the YSU planetary boundary layer (PBL) scheme. The implicit (A) and explicit (B) BEP and BEP+BEM source terms are added as forcing to the YSU implicit/explicit vertical diffusion equation solvers. Forcing is applied for zonal momentum, meridional momentum, potential temperature, and water vapor mixing ratio. The A sources are added to the tridiagonal matrix and the B sources are added to the explicit term vectors. Additionally, the finite differencing is modified using the BEP volume fraction and surface fraction not occupied by buildings. The BEP and BEP+BEM TKE tendency terms are added to the diagnostic TKE in YSU when topographic drag is turned on.

References:
   * Hendricks, E. A., J. C. Knievel, and Y. Wang, 2020: Addition of multilayer urban canopy models to a nonlocal planetary boundary layer parameterization and evaluation using ideal and real cases, J. Appl. Met. Clim., 59, 1369-1392.
   * Martilli et al. (2009), Description of the modifications made in WRF.3.1 and short user's manual of BEP, NCAR technical note. 

Hendricks et al. (2020) demonstrated that the scheme works properly and produces expected behaviors. Results are qualitatively similar to MYJ and BOULAC with BEP and BEP+BEM. Minor differences are due to the different treatment of vertical mixing among the schemes.

Note: This is a new pull request to replace pull request #1196. Pull request #1196 had the changes implemented into a new urban YSU routine (module_bl_ysuurb.F). This pull request adds the changes to the original YSU PBL routine (module_bl_ysu.F) instead of a new urban routine. After discussions with the WRF model and YSU scheme experts, it was determined that it would be best to implement the changes into the original YSU routine.

LIST OF MODIFIED FILES:
phys/module_bl_ysu.F
phys/module_pbl_driver.F
phys/module_physics_init.F
run/README.namelist

TESTS CONDUCTED:
1. Real-case 48-h simulations of Houston urban area were conducted (10/05/2017-10/07/2017) with an innermost nest grid spacing of 1 km. The following test suite was set up:

Baseline tests (version before physics enhancement):
   a. MYJ BEP
   b. MYJ BEP+BEM
   c. SU BULK
   d. SU SLUCM

New code tests (new version with enhancement):
   e. YSU BULK
   f. YSU SLUCM
   g. YSU BEP
   h. YSU BEP+BEM   
   i. MYJ BEP
   j. MYJ BEP+BEM 

   * Tests (e) and (f) were identical to tests (c) and (d), demonstrating that the new code did not affect the BULK and SLUCM existing code with the YSU PBL parameterization. 
   * Tests (g) and (h) were shown to produce the same behaviors as Hendricks et al. (2020).
   * Tests (i) and (j) were identical to tests (a) and (b), demonstrating that the new code did not affect any existing BEP and BEP+BEM code with the MYJ PBL parameterization. 

2. The real-case example test of the NMM dynamical core was also run for the YSU scheme without BEP and BEP+BEM prior to code modifications, and after the code modifications. The output was examined and shown to be the same. Therefore, the new YSU scheme with BEP and BEP+BEM functions properly in the NMM dynamical core when BEP and BEP+BEM are not active (note that the NMM dynamical core does not support BEP and BEP+BEM).

3. Since YSU is now functional with BEP and BEP+BEM, this line was added to replace the old lines in module_physics_init.F for all other PBL routines:
```
IF ((SF_URBAN_PHYSICS.eq.2).OR.(SF_URBAN_PHYSICS.EQ.3)) CALL wrf_error_fatal &
            ( 'module_physics_init: use ysu (option 1), myj (option 2), or boulac (option 8) with BEP/BEM urban scheme' )
```
This printout was shown to occur when attempting to run with PBL schemes that do not support the BEP and BEP+BEM functionality.

4. The restart capability was demonstrated to function properly with the new YSU BEP and YSU BEP+BEM code. 
5. Jenkins testing is all PASS. 

RELEASE NOTE: The multilayer BEP (Building Effects Parameterization) and BEP+BEM (BEP with the Building Energy Model) urban canopy models (UCMs) are added to the Yonsei University (YSU) planetary boundary layer parameterization. Reference: Hendricks, E. A., J. C. Knievel, and Y. Wang, 2020: Addition of multilayer urban canopy models to a nonlocal planetary boundary layer parameterization and evaluation using ideal and real cases, J. Appl. Met. Clim., 59, 1369-1392.

### DIFF
--- a/phys/module_bl_ysu.F
+++ b/phys/module_bl_ysu.F
@@ -29,6 +29,13 @@ contains
                   uoce,voce,                                                   &
                   rthraten,ysu_topdown_pblmix,                                 &
                   ctopo,ctopo2,                                                &
+                  IDIFF,FLAG_BEP,FRC_URB2D,                                    &
+                  A_U_BEP,A_V_BEP,A_T_BEP,                                     &
+                  A_Q_BEP,                                                     &
+                  A_E_BEP,B_U_BEP,B_V_BEP,                                     &
+                  B_T_BEP,B_Q_BEP,                                             &
+                  B_E_BEP,DLG_BEP,                                             &
+                  DL_U_BEP,SF_BEP,VL_BEP,                                      &
                   ids,ide, jds,jde, kds,kde,                                   &
                   ims,ime, jms,jme, kms,kme,                                   &
                   its,ite, jts,jte, kts,kte,                                   &
@@ -108,6 +115,23 @@ contains
 !-- jte         end index for j in tile
 !-- kts         start index for k in tile
 !-- kte         end index for k in tile
+!-- IDIFF       BEP/BEM+BEM diffusion flag
+!-- FLAG_BEP    FLAG TO USE BEP AND BEP+BEM
+!-- FRC_URB2D   urban fraction (not currently used)
+!-- A_U_BEP     BEP/BEP+BEM implicit component u-mom
+!-- A_V_BEP     BEP/BEP+BEM implicit component v-mom
+!-- A_T_BEP     BEP/BEP+BEM implicit component pot. temp.
+!-- A_Q_BEP     BEP/BEP+BEM implicit component vapor mixing ratio
+!-- A_E_BEP     BEP/BEP+BEM implicit component TKE
+!-- B_U_BEP     BEP/BEP+BEM explicit component u-mom
+!-- B_V_BEP     BEP/BEP+BEM explicit component v-mom
+!-- B_T_BEP     BEP/BEP+BEM explicit component pot.temp.
+!-- B_Q_BEP     BEP/BEP+BEM explicit component vapor mixing ratio
+!-- B_E_BEP     BEP/BEP+BEM explicit component TKE
+!-- DLG_BEP     Height above ground Martilli et al. (2002) Eq. 24
+!-- DL_U_BEP    modified length scale Martilli et al. (2002) Eq. 22
+!-- SF_BEP      fraction of vertical surface not occupied by buildings
+!-- VL_BEP      volume fraction of grid cell not occupied by buildings
 !-------------------------------------------------------------------------------
 !
    integer,parameter ::  ndiff = 3
@@ -180,6 +204,17 @@ contains
    integer,  dimension( ims:ime, jms:jme )                                   , &
              intent(out  )   ::                                        kpbl2d
    logical,  intent(in)      ::                                       flag_qi
+   integer, intent(in) :: IDIFF
+   logical, intent(in) ::     FLAG_BEP
+   real,dimension(ims:ime,kms:kme,jms:jme),intent(in) :: A_U_BEP    &
+                                              ,A_V_BEP,A_T_BEP         &
+                                             ,A_E_BEP,B_U_BEP         &
+                                             ,A_Q_BEP,B_Q_BEP         &
+                                             ,B_V_BEP,B_T_BEP         &
+                                             ,B_E_BEP,DLG_BEP         &
+                                             ,DL_U_BEP                &
+                                             ,VL_BEP,SF_BEP
+   real, dimension(ims:ime,jms:jme),intent(in) :: FRC_URB2D
 !
 !optional
 !
@@ -206,10 +241,29 @@ contains
                                                                         dvsfc, &
                                                                         dtsfc, &
                                                                         dqsfc
-
+   real,dimension(ims:ime,kms:kme,jms:jme) :: A_U,A_V,A_T,A_E,B_U,B_V,B_T,B_E
+   real,dimension(ims:ime,kms:kme,jms:jme) :: A_Q,B_Q,DLG,DL_U,B_E_FACE,DZ_KFLIP
+   real,dimension(ims:ime,kms:kme,jms:jme) :: SFK, VLK
+   real :: bepswitch ! 0 if not using bep or bep+bem, 1 if using
 !
    qv2d(its:ite,:) = 0.0
 !
+   bepswitch = 0.0
+   A_U(:,:,:)=0.0
+   A_V(:,:,:)=0.0
+   A_T(:,:,:)=0.0
+   A_Q(:,:,:)=0.0
+   A_E(:,:,:)=0.0
+   B_U(:,:,:)=0.0
+   B_V(:,:,:)=0.0
+   B_T(:,:,:)=0.0
+   B_Q(:,:,:)=0.0
+   B_E(:,:,:)=0.0
+   SFK(:,:,:)=1.0
+   VLK(:,:,:)=1.0
+   DL_U(:,:,:)=0.0
+   DLG(:,:,:)=0.0
+
    do j = jts,jte
       do k = kts,kte+1
         do i = its,ite
@@ -225,6 +279,28 @@ contains
           if(flag_qi) qv2d(i,k+kte+kte) = qi3d(i,k,j)
         enddo
       enddo
+
+      if(flag_bep) then
+        bepswitch=1.0
+        do k=kts,kte
+          do i=its,ite
+            A_U(i,k,j)=A_U_BEP(i,k,j)
+            A_V(i,k,j)=A_V_BEP(i,k,j)
+            A_T(i,k,j)=A_T_BEP(i,k,j)
+            A_Q(i,k,j)=A_Q_BEP(i,k,j)
+            A_E(i,k,j)=A_E_BEP(i,k,j)
+            B_U(i,k,j)=B_U_BEP(i,k,j)
+            B_V(i,k,j)=B_V_BEP(i,k,j)
+            B_T(i,k,j)=B_T_BEP(i,k,j)
+            B_Q(i,k,j)=B_Q_BEP(i,k,j)
+            B_E(i,k,j)=B_E_BEP(i,k,j)
+            SFK(i,k,j)=SF_BEP(i,k,j)
+            VLK(i,k,j)=VL_BEP(i,k,j)
+            DL_U(i,k,j)=DL_U_BEP(i,k,j)
+            DLG(i,k,j)=DLG_BEP(i,k,j)
+          enddo
+        enddo
+      endif
 !
       call ysu2d(J=j,ux=u3d(ims,kms,j),vx=v3d(ims,kms,j)                       &
               ,tx=t3d(ims,kms,j)                                               &
@@ -254,6 +330,14 @@ contains
               ,rthraten=rthraten(ims,kms,j),p2diORG=p3di(ims,kms,j)            &
               ,ysu_topdown_pblmix=ysu_topdown_pblmix                           &
               ,ctopo=ctopo(ims,j),ctopo2=ctopo2(ims,j)                         &
+              ,a_u2d=A_U(ims,kms,j), a_v2d=A_V(ims,kms,j)                      &
+              ,a_t2d=A_T(ims,kms,j), a_q2d=A_Q(ims,kms,j)                      &
+              ,b_u2d=B_U(ims,kms,j), b_v2d=B_V(ims,kms,j)                      &
+              ,b_t2d=B_T(ims,kms,j), b_q2d=B_Q(ims,kms,j)                      &
+              ,b_e2d=B_E(ims,kms,j), a_e2d=A_E(ims,kms,j)                      &
+              ,sfk2d=SFK(ims,kms,j), vlk2d=VLK(ims,kms,j)                      &
+              ,dlu2d=DL_U(ims,kms,j), dlg2d=DLG(ims,kms,j)                     &
+              ,frc_urb1d=FRC_URB2D(ims,j), bepswitch=bepswitch                 &
               ,ids=ids,ide=ide, jds=jds,jde=jde, kds=kds,kde=kde               &
               ,ims=ims,ime=ime, jms=jms,jme=jme, kms=kms,kme=kme               &
               ,its=its,ite=ite, jts=jts,jte=jte, kts=kts,kte=kte   )
@@ -288,6 +372,11 @@ contains
                   rthraten,p2diORG,                                            &
                   ysu_topdown_pblmix,                                          &
                   ctopo,ctopo2,                                                &
+                  a_u2d, a_v2d, a_t2d, a_q2d,                                  &
+                  b_u2d, b_v2d, b_t2d, b_q2d,                                  &
+                  b_e2d, a_e2d, sfk2d, vlk2d,                                  &
+                  dlu2d, dlg2d,                                                &
+                  frc_urb1d, bepswitch,                                        &
                   ids,ide, jds,jde, kds,kde,                                   &
                   ims,ime, jms,jme, kms,kme,                                   &
                   its,ite, jts,jte, kts,kte,                                   &
@@ -344,9 +433,11 @@ contains
 !              ==> kzo = 0.1 for momentum and = 0.01 for mass to account for
 !                  internal wave mixing of large et al. (1994), songyou hong, feb 2016
 !              ==> alleviate superious excessive mixing when delz is large
+!              add multilayer urban canopy models of BEP and BEP+BEM, jan 2021
 !
 !     references:
 !
+!        hendricks, knievel, and wang (2020), j. appl. meteor. clim.
 !        hong (2010) quart. j. roy. met. soc
 !        hong, noh, and dudhia (2006), mon. wea. rev.
 !        hong and pan (1996), mon. wea. rev.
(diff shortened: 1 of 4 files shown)


davegill:
jenkins is OK
```
Please find result of the WRF regression test cases in the attachment. This build is for Commit ID: fb1dcdad5f9544e598a12e3d31f2fdc9eedfd11b, requested by: eahendricks for PR: https://github.com/wrf-model/WRF/pull/1305. For any query please send e-mail to David Gill.

    Failed Simulations are: 
    None
    None
```

dudhia:
Check if mods are needed in README.namelist.

eahendricks:
Thanks for letting me know about that. Yes, the README.namelist should be changed to the bolded type. 

sf_urban_physics(max_dom)           = 0, ! activate urban canopy model (in Noah and Noah-MP LSMs only)
                                                           = 0: no
                                                           = 1: Single-layer, UCM 
                                                           = 2: Multi-layer, Building Environment Parameterization (BEP) scheme 
                                                                 **(works only with YSU, MYJ, and BouLac PBL)**
                                                           = 3: Multi-layer, Building Environment Model (BEM) scheme 
                                                                **(works only with YSU, MYJ, and BouLac PBL)**

weiwangncar:
I'm ok with this PR.

phys/module_bl_ysu.F
```diff
@@ -206,10 +241,30 @@ subroutine ysu(u3d,v3d,th3d,t3d,qv3d,qc3d,qi3d,p3d,p3di,pi3d,               &
                                                                         dvsfc, &
                                                                         dtsfc, &
                                                                         dqsfc
-
+   real,dimension(ims:ime,kms:kme,jms:jme) :: a_u,a_v,a_t,a_e,b_u,b_v,b_t,b_e, &
```
davegill, line 244: Should these be tile-sized, if they are not sent out of the scheme?

phys/module_bl_ysu.F
```diff
@@ -206,10 +241,30 @@ subroutine ysu(u3d,v3d,th3d,t3d,qv3d,qc3d,qi3d,p3d,p3di,pi3d,               &
...
+   b_v(:,:,:)=0.0
+   b_t(:,:,:)=0.0
+   b_q(:,:,:)=0.0
+   b_e(:,:,:)=0.0
+   sfk(:,:,:)=1.0
+   vlk(:,:,:)=1.0
+   dl_u(:,:,:)=0.0
+   dlg(:,:,:)=0.0
+   frcurb(:,:)=0.0   
+
```
davegill, line 267: Are these initializations necessary? 20 lines down, if the "flag_bep" is true, then they are all initialized.

dudhia:
It looks like these are packaged with the urban options, so they should not
be accessed outside anything that checks whether those options are used.

phys/module_bl_ysu.F
eahendricks: Yes, they are necessary for the initialization when "flag_bep" is false. I wanted to have a default initialization for the way most users will run (without BEP).

dudhia:
Note that being in a package, these arrays will not be allocated space when
the urban option is 0,
so this code should not be needed. Given this, any code using these arrays
has to be protected
to avoid out-of-bounds usage of unallocated arrays.

weiwangncar:
@eahendricks Can you address the package issue raised by Jimy?

phys/module_bl_ysu.F
eahendricks: Sorry for the late reply. I did not understand what "tile-sized" meant. Correct, they are not sent out of the scheme. Let me know any suggested changes. Thank you.

eahendricks:
These are not packaged only with the urban options. They need to be allocated and have the default values for urban option 0, at least in the current way I wrote the code (see lines 1260-1265, for example, or lines 1230-1231). I'd be happy to modify this code as you all see fit, but right now I am unsure what you want me to change.

phys/module_bl_ysu.F
weiwangncar: @eahendricks Tile size means the dimensions are its:ite, jts:jte and kts:kte. If you look at lines just above yours, you'll see local arrays declared for YSU: rqvbl2dt, qv2d, etc.,

phys/module_bl_ysu.F
```diff
@@ -254,6 +332,14 @@ subroutine ysu(u3d,v3d,th3d,t3d,qv3d,qc3d,qi3d,p3d,p3di,pi3d,               &
               ,rthraten=rthraten(ims,kms,j),p2diORG=p3di(ims,kms,j)            &
               ,ysu_topdown_pblmix=ysu_topdown_pblmix                           &
               ,ctopo=ctopo(ims,j),ctopo2=ctopo2(ims,j)                         &
+              ,a_u2d=a_u(ims,kms,j), a_v2d=a_v(ims,kms,j)                      &
+              ,a_t2d=a_t(ims,kms,j), a_q2d=a_q(ims,kms,j)                      &
+              ,b_u2d=b_u(ims,kms,j), b_v2d=b_v(ims,kms,j)                      &
+              ,b_t2d=b_t(ims,kms,j), b_q2d=b_q(ims,kms,j)                      &
+              ,b_e2d=b_e(ims,kms,j), a_e2d=a_e(ims,kms,j)                      &
+              ,sfk2d=sfk(ims,kms,j), vlk2d=vlk(ims,kms,j)                      &
+              ,dlu2d=dl_u(ims,kms,j), dlg2d=dlg(ims,kms,j)                     &
```
weiwangncar, line 341: @eahendricks Once you change above, these arrays will need to changed too, and those declaration inside ysu2d.

phys/module_bl_ysu.F
eahendricks: @weiwangncar @davegill Thanks for the clarification. As a new WRF user, I probably do not fully understand the differences between memory and tile indices. Do the memory indices include the halo cells for each tile in domain decomp, whereas tile indices do not (so tile size would always be smaller than memory size)? I think you are correct they should be tile sized since they are not sent out. I will have to make the modifications and run it again through my test suite to make sure the results look the same. If so I'll commit the changes.

weiwangncar: @eahendricks Tile sizes can be smaller than patches which are dimensioned by memory sizes, if another layer of parallelism, OpenMP, is activated. If one just uses dmpar, then tile and patch sizes are the same. The definition for a tile is section of a patch allocated to a shared-memory processor within a node. So the tiles do include halo regions. By changing your array dimensions, it helps to enable shared-memory parallelism to work properly, at least we hope.

eahendricks: @weiwangncar Thanks for the explanation. I was using dmpar, which is why I did not notice any problem. So glad you guys caught this error.

weiwangncar:
@eahendricks If you receive the results from Jenkins test, can you please post it?

davegill:
@weiwangncar @eahendricks 
Wei,
The last jenkins I have is 5 days ago, associated with the latest mods. All pass.

```
Please find result of the WRF regression test cases in the attachment. This build is for Commit ID: f1d9d7f862593b35e0f1e6ebfdf9e8d4903e9ee8, requested by: eahendricks for PR: https://github.com/wrf-model/WRF/pull/1305. For any query please send e-mail to David Gill.

    Failed Simulations are: 
    None
    None
```

eahendricks:
Thanks for the posting the jenkins @davegill. I also ran the new code through my test suite, and the physical results look the same as before the tile-size correction. I should note that all physical tests (making sure results are sensible) were done using INTEL (ifort/icc): SGI MPT configure=24 option (dmpar) on Cheyenne.

dudhia (approved):

weiwangncar:
@eahendricks May I ask for one more test? This is a test for OpenMP. You can compile this code using OpenMP (or smpar) only, and run the model twice (using 4 or more threads) and use diffwrf (built in external/io_netcdf/ directory to compare the results. If they are identical, then it is likely mean that this code is thread safe. I ask this because YSU+urban is not tested using Jenkins. If you would like to do more, a more formal test would be to compile the code with configure -d for both dmpar and smpar separately, and compare the model output using diffwrf (10 steps are we do in Jenkins tests). Thanks.

eahendricks:
@weiwangncar Sure, no problem. I configured on Cheyenne with option 23 (smpar). But I have never run WRF in this configuration before. Do you have an example smpar/OpenMP run script you can pass me (any special directives and the command)? (assuming mpiexec_mpt will not work). I tried mpirun and it did not work either. I tried googling all of this but no luck so thought in the interest of time I would just ask.

weiwangncar:
@eahendricks Understood. I just learnt that myself. Here are the two lines that are important in a script:
#PBS -l select=1:ncpus=4:mpiprocs=4
./wrf.exe
If you'd like to see a full script, try /glade/scratch/weiwang/runomp.csh. It uses 4 threads in this example.

eahendricks:
@weiwangncar Thanks. I did a run with 36 threads because of larger domain. With sf_urban_physics=0 it ran ok. With sf_urban_physics=3 (BEP+BEM), it crashed. So it looks like there is a problem with BEP+BEM and also likely BEP. I'll look into this more tomorrow. These are the last lines in "test_job.o6531165":

CALL rrtmg_sw
 call surface_driver
in SFCLAY
in NOAH DRV
wrf_abort

The directory of the run is here: /glade/scratch/erichend/WRF/ysu4smpar/wrf/. The error appears to be in "module_surface_driver.F".  But it is strange because I did not modify this file. If you have any suggestions on what may be causing this error, please let me know. Glad you suggested this extra test.

weiwangncar:
@eahendricks For the no urban run, did you run the model twice and compare the results?

weiwangncar:
@eahendricks Please don't turn on debug_levels, it's not very helpful. Good that you're using a single domain for the test. There may be a memory issue. Could you try to use 16 threads, and definitely comment out 'setenv OMP_NUM_THREADS 36' - this is not needed on Cheyenne according its documentation. I would also suggest that you split output files to one per file by setting  frames_per_outfile   = 1 and history_interval_s = 50 (instead of history_interval). This will make comparing files easier - at least it helps me. If you didn't compile the code with 'configure -d' (debug mode), it may be helpful to do so.

eahendricks:
@weiwangncar No I did not compare the model twice for the no urban run. I thought Jenkins already did that, and you said YSU+urban is the one we needed to test with OpenMP? I just tried with 16 threads. Segmentation fault. Would you recommend I try a much smaller number of grid points? I was using 298x298x60. I reduced the vertical levels to 40, and same crash. I could try something like 50x50x20 (and rerun WPS).

eahendricks:
Just ran with MYJ BEP, and it crashed with the same segmentation fault. So it is a BEP/BEP+BEM issue, and not related specifically to YSU BEP/BEP+BEM.

weiwangncar:
@eahendricks You are right. The no-urban case is probably safe because of the Jenkins test.

weiwangncar:
@eahendricks Try the case we use in Jenkins tests. I have wrfbdy_d01 and wrfinput_d01 created for this case in /glade/scratch/weiwang/v42/WRF/urban/ and you can try it for your code. It is a small case, easy to run. I used 4 threads for the tests.

eahendricks:
@weiwangncar Thanks. It ran fine with my executable and YSU BEP+BEM (/glade/scratch/erichend/WRF/ysu4smparsmall/wrf/). So I think we are good to go. It must have been some issue with my WPS/WRF setup that OMP did not like.

weiwangncar:
@eahendricks Good to know it is successful. 
@davegill @dudhia Any other comments?

eahendricks:
I ran two diffwrf tests. One with MYJ BEP+BEM and one with YSU BEP+BEM. You can find the differences from the last Jenkins file here: /glade/scratch/erichend/WRF/ysu4smparsmall/wrf/. MYJ: "DIFFWRFMYJBEPBEM.out", YSU: "DIFFWRFYSUBEPBEM.out". So neither is reproducible. This tells me maybe it is a BEP/BEM issue and not likely related to any issue with the implementation in YSU.

weiwangncar:
@eahendricks Are the comparisons between two OpenMP runs or between MPI and OpenMP runs? Did you compile the code with 'configure -d'? In other words, is the code compiled without optimization?

eahendricks:
@weiwangncar Two OpenMP runs. I just re-compiled with 'configure -d'. Still differences between runs for both MYJ and YSU with BEP+BEM. Have you verified previously that BEP/BEP+BEM is reproducible using OpenMP (say either with MYJ or BouLac)? From my tests, it appears not, but might be good to have another sanity check.

weiwangncar:
@eahendricks Using 'configure -d', two OpenMP runs (run for 30 min) are identical for MYJ PBL, but different only at 30 min for BouLac (output at 3 min interval). This is for BEP+BEM, and no radiation calls at 30 min.
When compared to MPI runs, all options show differences in TSK field after first time step.
After checking our current Jenkins tests, it looks like we are NOT testing urban + Noah, and we only test urban + NoahMP, and we only run the model for a 9 min forecast (this info comes from the namelist tar file: Namelists/weekly/em_real/MPI)
@dudhia @davegill Do you have any comments?

weiwangncar:
@eahendricks After talking to Dave, we find a namelist that we do test Noah + urban physics 2 with BouLac PBL, with nesting and 9 min of forecast time. Could you see if your code can pass this test with MPI and OpenMP? If you don't have a nest input, you can just do it for a single domain. Thanks.

eahendricks:
@weiwangncar OpenMP results below. I tested with YSU also in addition to BouLac. Will post another update once I finish the MPI runs. Looks like the issue is BEM. Tests pass with BEP, but not with BEP+BEM.
1. Noah + urban physics 2 with BouLac PBL, one nest. PASS: diffwrf has no differences between two OpenMP runs at 9 mins.
2. Noah + urban physics 2 with YSU PBL, one nest. PASS: diffwrf has no differences between two OpenMP runs at 9 mins.
3. Noah + urban physics 3 with BouLac PBL, one nest. FAIL: diffwrf has differences between two OpenMP runs at 9 mins.
4. Noah + urban physics 3 with YSU PBL, one nest. FAIL: diffwrf has differences between two OpenMP runs at 9 mins.

davegill:
@eahendricks @weiwangncar 
> OpenMP results below. I tested with YSU also in addition to BouLac. Will post another update once I finish the MPI runs. Looks like the issue is BEM. Tests pass with BEP, but not with BEP+BEM.

Folks,
We _know_ that Chemistry is not OpenMP safe. We can just put a check_a_mundo test for urban=3 with OpenMP. And we make sure not to add an OpenMP test for jenkins. We actually already have that suite of tests that are not OpenMP friendly.

There is no need to fix an existing issue with the WRF code with this PR, or to derail this PR due to an existing WRF OpenMP issue.

weiwangncar (approved):

weiwangncar:
@eahendricks Thanks for doing the tests. I think this PR is ready to be merged. 
@davegill Can we worry about the check after the branch cut-off?

davegill:
@weiwangncar 
> Can we worry about the check after the branch cut-off?

Wei,
Yes, the clean up of "if this is OpenMP and if we are selecting urban==3; then STOP" can come later.

I am with this PR

davegill (approved):

davegill:
@weiwangncar 
Jimy and I have approved these UCM + YSU PBL changes. You may merge.

davegill:
@weiwangncar @eahendricks 

Before merging, there is an outstanding request to update the README.namelist:

> Thanks for letting me know about that. Yes, the README.namelist should be changed to the bolded type. 

sf_urban_physics(max_dom)           = 0, ! activate urban canopy model (in Noah and Noah-MP LSMs only)
                                                           = 0: no
                                                           = 1: Single-layer, UCM 
                                                           = 2: Multi-layer, Building Environment Parameterization (BEP) scheme 
                                                                 **(works only with YSU, MYJ, and BouLac PBL)**
                                                           = 3: Multi-layer, Building Environment Model (BEM) scheme 
                                                                **(works only with YSU, MYJ, and BouLac PBL)**

davegill:
@weiwangncar 
>Before merging, there is an outstanding request to update the README.namelist:

OK, the file was modified, just not mentioned in the PR. I updated the PR message. OK to merge.

:shipit:

PR #1306, opened 2020-10-27
update namelist.input to current recommendations and newer case

kkeene44:
TYPE: no impact 

KEYWORDS: namelist, update, dates, e_we, e_sn, domain size

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
The default namelist.input file for real cases had several settings that go against best practices and recommendations. 
- e_we and e_sn default values were 74x61, 112x97. We tell users to never use a domain smaller than 100x100, and we continue to have many users running small domains, and then seeking help. 
- default case and dates were very old (2000). 
- default grid space was very coarse, which isn't used as often with finer resolution input data available.
- default num_metgrid_levels was set to 32, which is the number of levels in older GFS data
- default cudt was set to 5, and we recommend 0

Solution:
- Updated the case to the 2019 Hurricane Dorian case (Sept 2019). This included the case dates, run_hours (36) and domain specifications.
- Updated dx/dy to 15 km - which meant updating time_step and radt
- Added a comment after time_step to 1) show that comments are allowed in the namelist, when using a "!" and 2) to not set that value larger than 6xDX (a very common user problem we see)
- Updated domain size to larger, more realistic dimensions, which meant updating e_vert and i/j_parent_start
- Updated num_metgrid_levels to 34, which is the number of levels in the current GFS data
- Updated cudt to 0
- Also added a new file (namelist.input.default) that has descriptions for all the default parameters, including when NOT to add additional columns (a common user issue is when they try to add columns to non max_dom parameters).
*Note: WPS [PR#160](https://github.com/wrf-model/WPS/pull/160) was also created to modify corresponding values in namelist.wps

LIST OF MODIFIED FILES:
M    test/em_real/namelist.input
A    test/em_real/namelist.input.default

TESTS CONDUCTED: 
1. I ran a test for the new case to make sure it all looked good - used GFS 0.25 degree data.


kkeene44:
@dudhia @weiwangncar @davegill 
Let me know if any other lines should be modified, or if I should change anything.

davegill:
Kelly,
I always trip on the default GFS number of levels (now 34). Thanks

1. Jimy and Wei said that km_opt and diff_opt have better recommendations (4 and 2, respectively)
2. What is a reasonable number of vertical levels, 45? 50?

Maybe put the raw data for this new case in a "soon to be renamed general purpose area": https://www2.mmm.ucar.edu/wrf/users/special_code.html

Dave

kkeene44:
@davegill 
I've taken care of item 1) above (km_opt was already set to 4), and I added the data to the web page mentioned.

agumartina:
Great work @kkeene44 
>     2. What is a reasonable number of vertical levels, 45? 50?
We are currently discussing this topic
what value should it be? 45 or 50?

Agu

dudhia:
Good question, I think for this number of levels and 50 hPa top dzstretch_s
should be 1.1
but it uses the default 1.3.
Jimy

davegill:
@dudhia @kkeene44 
Jimy,
1. Are you saying that the default value of dztretch_s should be 1.1 instead of 1.3?
2. Is it better to let the user define the model lid and the dzstretch_s, and then let the real program determine the required levels?

dudhia:
There is an argument for changing the default to 1.1 but at least in the
example namelist it is useful to add this as 1.1 which also informs users
that these options exist to automatically set levels. For this number of
levels
1.3 is not optimal as it is designed for people who want few levels, like
28.
We also have dzstretch_u and I have this doc with some examples.

kkeene44:
@dudhia 
Okay, I'll add a line for dzstretch_s = 1.1 to the default namelist. 
Is there an equation for figuring out the appropriate number of levels and value for dzstretch_s?
 I assume users will ask about this.

dudhia:
There is the doc I shared which may be in the User Guide(?).

weiwangncar:
I have to think about this a bit more, but what hits me is that it is not entirely easy to play with dzstretch_s and dzstretch_s and dzstretch_u for inexperienced users. If we decide for a 50 mb ptop for general application, and 50 levels, whatever the appropriate values for dzstretch_s/u should be made default.

dudhia:
Yes, usually we need both dzstretch_s and dzstretch_u to get optimal
levels that not only change smoothly but also their derivative changes
smoothly.
The table gives some clues for a 50 hPa top.

kkeene44:
@dudhia 
I don't see a doc. Am I missing it somewhere?

weiwangncar:
@kkeene44 It is at the end of Chapter 4 under initialization.

davegill:
@dudhia @weiwangncar @kkeene44 
Folks,
This is an update to the default namelist settings. This can be done incrementally to the develop branch. I say, put in the ideas that Kelly has suggested so far, and then we'll see if we want to put in more updates.

davegill:
> This is an update to the default namelist settings. This can be done incrementally to the develop branch. I say, put in the ideas that Kelly has suggested so far, and then we'll see if we want to put in more updates.

Folks,
What is proposed here is good. If we want more, we add more later.

davegill (approved):
Almost 4 months, this has to be OK, sheesh

PR #1307, opened 2020-10-29
use tke_seed_value only when necessary

matzegoebel:
TYPE: bug fix

KEYWORDS: tke, isfflx, tke_heat_flux, tke_drag_coefficient, tke_seed

DESCRIPTION OF CHANGES:
Problem:
The tke seed value is used in the absence of surface drag or a surface heat flux, since there is no way to generate tke without pre-existing tke. So far the usage of the seed value is only tied to the namelist variables tke_drag_coefficient and tke_heat_flux being greater than 0. However, these namelist variables are only applied if no PBL scheme is used, diff_opt=2, and isfflx=0.

If a PBL scheme is used or diff_opt=1, then isfflx=0 is a sufficient condition to get zero fluxes (see also technical notes).

Solution:
Fixed the if statements to enforce a correct setting of the tke seed value that fits to the purpose of the seed.

LIST OF MODIFIED FILES: 
dyn_em/module_diffusion_em.F
wrftladj/module_diffusion_em_ad.F
wrftladj/module_diffusion_em_tl.F

TESTS CONDUCTED:
1. Jenkins testing is all pass.

RELEASE NOTE: corrected the tke seed value to be non-zero if and only if the surface heat and momentum fluxes are zero depending on isfflx, diff_opt, and bl_pbl_physics

### DIFF
--- a/dyn_em/module_diffusion_em.F
+++ b/dyn_em/module_diffusion_em.F
@@ -2160,9 +2160,18 @@ END SUBROUTINE smag2d_km
 ! tke_seed if the drag and flux are off.
 
     c_k = config_flags%c_k
-    tke_seed = tke_seed_value
-    if( (config_flags%tke_drag_coefficient .gt. epsilon) .or.  &
-        (config_flags%tke_heat_flux .gt. epsilon)  ) tke_seed = 0.
+    tke_seed = 0.
+    IF (config_flags%isfflx .eq. 0) THEN
+      IF ((config_flags%diff_opt .eq. 2) .and. (config_flags%bl_pbl_physics .eq. 0)) THEN
+        IF( (config_flags%tke_drag_coefficient .lt. epsilon) .and.  &
+            (config_flags%tke_heat_flux .lt. epsilon)  )  THEN
+          tke_seed = tke_seed_value
+        ENDIF
+      ELSE
+        !tke_drag_coefficient and tke_heat_flux are irrelevant here
+        tke_seed = tke_seed_value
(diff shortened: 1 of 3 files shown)


dudhia:
I would note that diff_opt=1 would not be a normal choice without a PBL because its vertical diffusion is fixed and would not use tke. Even mixing the tke scheme with a PBL scheme is a little hard to justify.

matzegoebel:
You're definitely right on the first one. The diff_opt condition is only there for completeness, because diff_opt=1 + km_opt=2 is in principle allowed. Mixing a PBL scheme with a TKE based scheme for horizontal diffusion, on the other hand, is used in a number of studies. Whether that's appropriate is a different question.
Anyway, this code modification just aims at finding out whether the surface fluxes are zero, in which case you need a tke seed value, or not. I don't know if the seed does any harm, but in case you use an LES with isfflx=1 or 2, you don't need any seed, no matter the values of tke_heat_flux or tke_drag_coefficient. In case you use a PBL scheme and isfflx=0, on the other hand, you do need a seed, no matter the values of tke_heat_flux or tke_drag_coefficient. Or am I mistaken?

dudhia (approved):

weiwangncar:
@dudhia Can I say this for a release message:
Corrected the tke seed value to be non-zero for km_opt 2 and 5 if and only if the surface heat and momentum fluxes are zero depending on isfflx, diff_opt, and bl_pbl_physics.

dudhia:
looks OK to me

PR #1314, opened 2020-11-04
Corner points

### LINKED ISSUE #1311: Symmetric and open BC: corner points missing
**Describe the bug**
In idealized simulations, the routine `set_physical_bc3d` sets the boundary zone points for periodic, symmetric, and open boundary conditions:
https://github.com/wrf-model/WRF/blob/f311cd5e136631ebf3ebaa02b4b7be3816ed171f/share/module_bc.F#L747-L1096
At first, the routine does not set the corner points (e.g., (ids-1,jds-1)): For the y-boundaries it iterates i only from `MAX(ids,its-1)` to  `MIN(ite+1,ide+istag)` and analogously for the x-boundaries (note that for open BC it actually also tries to set the corner points (see line 896), but not effectively, because the RHS values in lines 903-905 are still 0 there, as the y-BC has not been applied yet).

For doubly-periodic BC, the corner points are set correctly afterwards, starting at line 1100. In any other case, however, it seems that the corner points are not assigned and appear as zeros. 
At least that's what I see, when debugging the code with gdb.

These corner points are later used, e.g. in the diffusion module:
https://github.com/wrf-model/WRF/blob/f311cd5e136631ebf3ebaa02b4b7be3816ed171f/dyn_em/module_diffusion_em.F#L5511-L5517
Thus, because `rho(i_start-1,k,j_start-1)=0`, `xkxavg(i_start, k, j_start)` is much smaller than at the other locations.
Inaccuracies must obviously arise.

**Suggested solution**
The corner points should be set also for non-doubly-periodic boundary conditions. This should be easy to achieve by applying lines 747-1096 in module_bc.F twice, the second time only for the corner points. But someone needs to look at this in more detail.

matzegoebel:
Missing corner points are set when setting boundary conditions of 3D variables and all staggered variables are treated equally in symmetric BC. Thereby the periodic BC become cleaner and more consistent with set_physical_bc2d and the symmetric BC are fixed as horizontal deformation is treated now as a staggered variable which leads to correct BC for SGS momentum fluxes at the northern boundaries.

TYPE: bug fix

KEYWORDS: boundary conditions, symmetric, corners

DESCRIPTION OF CHANGES:
Problem:
In idealized simulations, the routine `set_physical_bc3d` sets the boundary zone points for periodic, symmetric, and open boundary conditions.
At first, the routine does not set the corner points (e.g., (ids-1,jds-1)): For the y-boundaries, it iterates i only from `MAX(ids,its-1)` to  `MIN(ite+1,ide+istag)` and analogously for the x-boundaries (note that for open BC it actually also tries to set the corner points (see line 896), but not effectively, because the RHS values in lines 903-905 are still 0 there, as the y-BC has not been applied yet).

For doubly-periodic BC, the corner points are set correctly afterward, starting at line 1100. In any other case, however, it seems that the corner points are not assigned and appear as zeros. 

These corner points are later used, e.g. in the SGS UV-flux (titau_21_12) in the diffusion module:
https://github.com/wrf-model/WRF/blob/f311cd5e136631ebf3ebaa02b4b7be3816ed171f/dyn_em/module_diffusion_em.F#L5511-L5517
Thus, because `rho(i_start-1,k,j_start-1)=0` and `xkx(i_start-1,k,j_start-1)=0`, we have `xkxavg(i_start, k, j_start)` much smaller than at the other locations.

This is usually not a problem, since the deformation defor12 and thus also the flux titau_21_12 should be 0 at these corner points when using symmetric boundary conditions on at least one of these boundaries.
I write "should" because that's a theoretical argument. In practice, only the deformation at the ids or jds boundaries is zero. Due to an inconsistency in the symmetric BC for the deformation, the deformation at the ide or jde boundaries is not 0. This problem is described in the following.

The deformation (variable = 'd') is treated like an unstaggered variable: https://github.com/wrf-model/WRF/blob/f311cd5e136631ebf3ebaa02b4b7be3816ed171f/share/module_bc.F#L1003-L1011
Thus, the deformation which was initially zero at the boundary is set to a non-zero value by `dat(i,k,jde) = dat(i,k,jde-1)`.
All staggered variables except the normal velocity (so also the deformation) should be treated equally in the symmetric boundaries since the location of the symmetric boundary is on the staggered grid. Therefore I adjusted the relevant if statement in the second commit.

If open boundary conditions are used on any of the boundaries, only the fluxes on the interior are calculated, so the corner points also don't matter. However, to avoid any current or future side effects due to the missing corner points, corner points are set anyway as it is done in `set_physical_bc2d`.

Solution:

1. commit:
The solution to set the corner points is the same as is already applied in the 2d routine `set_physical_bc2d`.

2. commit:
To treat all variables staggered in a certain direction equally, the variables `jstag` and `istag` are used in the if statement instead of checking the value of `variable`. This was done only for the symmetric BC. For the open BC, it might be necessary, too.

3. commit:
The corner point assignment for doubly-periodic BC is now redundant and was removed.

ISSUE:
Fixes #1311

LIST OF MODIFIED FILES: share/module_bc.F

TESTS CONDUCTED:
With the first commit, I tested, whether the corner points for doubly-periodic BC are set in the same way as in the original code. This test allowed removing this redundant part of the code in commit 3.

To compare the different model versions, I ran an idealized simulation with km_opt=2, diff_opt=2, tke_heat_flux=0.2, U = 5 m/s, periodic in x, symmetric in y.
The following figure shows the titau_21_12 flux at k=kds after 20 minutes for three model versions:
left: original, middle: fixed corners (commit 1), right: fixed corners and fixed symmetric BC for the deformation (commit 1 and 2)
fuy_sgs

All versions have titau_21_12=0 at j=jds. In the left and middle panel, the fluxes at jde and jde-1 are almost identical. However, the original version has fluxes with lower magnitudes at the upper left and right corners (as the corners of xkmh and rho are not set as explained above). In the right panel, the flux at j=jde is now also zero as it should be.

The following figure shows a time series of U at the upper left corner:
u_timeseries

We can see that fixing the symmetric BC to have zero fluxes at j=jde does have a significant effect.
Fixing the corners (commit 1) only has a minor effect, which is maximized at the corner points (j=jde, i=ids, and ide) as can be seen here:
u_diff

RELEASE NOTE: 
Missing corner points are set when setting boundary conditions of 3D variables and all staggered variables are treated equally in symmetric BC. Thereby the periodic BC become cleaner and more consistent with set_physical_bc2d and the symmetric BC are fixed as horizontal deformation is treated now as a staggered variable which leads to correct BC for SGS momentum fluxes at the northern boundaries.

### DIFF
--- a/share/module_bc.F
+++ b/share/module_bc.F
@@ -680,7 +680,8 @@ CONTAINS
       REAL,  DIMENSION( ims:ime , kms:kme , jms:jme ) :: dat
       TYPE( grid_config_rec_type ) config_flags
 
-      INTEGER  :: i, j, k, istag, jstag, itime, k_end
+      INTEGER  :: i, j, k, istag, jstag, itime, k_end, &
+                  i_start, i_end
 
       LOGICAL  :: debug, open_bc_copy
 
@@ -927,13 +928,27 @@ CONTAINS
 
 !  same procedure in y
 
+!  Set the starting and ending loop indexes in the 'i' direction, so that
+!  halo cells on the edge of the domain are also updated.  Begin with a default
+!  start and end index for inner tiles, and then modify if the tile is on the
+!  edge of the domain.
+
+      i_start = MAX(ids, its-1)
+      i_end = MIN(ite+1, ide+istag)
+      IF ( its .eq. ids) THEN
+        i_start = ims
+      END IF
+      IF ( ite .eq. ide) THEN
+        i_end = ime
+      END IF
+
       periodicity_y:  IF( ( config_flags%periodic_y ) ) THEN
         IF ( ( jds == jps ) .and. ( jde == jpe ) )  THEN      ! test if both north and south on processor
           IF( jts == jds ) then
 
             DO j = 0, -(bdyzone-1), -1
             DO k = kts, k_end
-            DO i = MAX(ids,its-1), MIN(ite+1,ide+istag)
+            DO i = i_start, i_end
               dat(i,k,jds+j-1) = dat(i,k,jde+j-1)
             ENDDO
             ENDDO
@@ -945,7 +960,7 @@ CONTAINS
 
             DO j = -jstag, bdyzone
             DO k = kts, k_end
-            DO i = MAX(ids,its-1), MIN(ite+1,ide+istag)
+            DO i = i_start, i_end
               dat(i,k,jde+j+jstag) = dat(i,k,jds+j+jstag)
             ENDDO
             ENDDO
@@ -964,7 +979,7 @@ CONTAINS
 
             DO j = 1, bdyzone
             DO k = kts, k_end
-            DO i = MAX(ids,its-1), MIN(ite+1,ide+istag)
+            DO i = i_start, i_end
               dat(i,k,jds-j) = dat(i,k,jds+j-1)
             ENDDO                               
             ENDDO
@@ -976,7 +991,7 @@ CONTAINS
 
               DO j = 1, bdyzone
               DO k = kts, k_end
-              DO i = MAX(ids,its-1), MIN(ite+1,ide+istag)
+              DO i = i_start, i_end
                 dat(i,k,jds-j) = - dat(i,k,jds+j)
               ENDDO              
               ENDDO
@@ -986,7 +1001,7 @@ CONTAINS
 
               DO j = 1, bdyzone
               DO k = kts, k_end
-              DO i = MAX(ids,its-1), MIN(ite+1,ide+istag)
+              DO i = i_start, i_end
                 dat(i,k,jds-j) = dat(i,k,jds+j)
               ENDDO              
               ENDDO
@@ -1007,7 +1022,7 @@ CONTAINS
 
             DO j = 1, bdyzone
             DO k = kts, k_end
-            DO i = MAX(ids,its-1), MIN(ite+1,ide+istag)
+            DO i = i_start, i_end
               dat(i,k,jde+j-1) = dat(i,k,jde-j)
             ENDDO                               
             ENDDO
@@ -1019,7 +1034,7 @@ CONTAINS
 
               DO j = 1, bdyzone
               DO k = kts, k_end
-              DO i = MAX(ids,its-1), MIN(ite+1,ide+istag)
+              DO i = i_start, i_end
                 dat(i,k,jde+j) = - dat(i,k,jde-j)
               ENDDO                               
               ENDDO
@@ -1029,7 +1044,7 @@ CONTAINS
 
               DO j = 1, bdyzone
               DO k = kts, k_end
-              DO i = MAX(ids,its-1), MIN(ite+1,ide+istag)
+              DO i = i_start, i_end
                 dat(i,k,jde+j) = dat(i,k,jde-j)
               ENDDO                               
               ENDDO
@@ -1050,7 +1065,7 @@ CONTAINS
                          ( jts == jds) .and. open_bc_copy )  THEN
 
             DO k = kts, k_end
-            DO i = MAX(ids,its-1), MIN(ite+1,ide+istag)
+            DO i = i_start, i_end
               dat(i,k,jds-1) = dat(i,k,jds)
               dat(i,k,jds-2) = dat(i,k,jds)
               dat(i,k,jds-3) = dat(i,k,jds)
@@ -1070,7 +1085,7 @@ CONTAINS
           IF (variable /= 'v' .and. variable /= 'y' ) THEN
 
             DO k = kts, k_end
-            DO i = MAX(ids,its-1), MIN(ite+1,ide+istag)
+            DO i = i_start, i_end
               dat(i,k,jde  ) = dat(i,k,jde-1)
               dat(i,k,jde+1) = dat(i,k,jde-1)
               dat(i,k,jde+2) = dat(i,k,jde-1)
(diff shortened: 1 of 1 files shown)


dudhia:
A lot to take in. Thanks for the detail.

davegill:
@weiwangncar @dudhia 
Folks,
Jimy already said that this was a lot to take in. Do we want Bill to take a look at this one, too?

dudhia:
I might look at it first, but not soon.

share/module_bc.F
```diff
@@ -927,13 +928,27 @@ SUBROUTINE set_physical_bc3d( dat, variable_in,        &
...
 
+!  Set the starting and ending loop indexes in the 'i' direction, so that
+!  halo cells on the edge of the domain are also updated.  Begin with a default
+!  start and end index for inner tiles, and then modify if the tile is on the
+!  edge of the domain.
+
+      i_start = MAX(ids, its-1)
+      i_end = MIN(ite+1, ide+istag)
+      IF ( its .eq. ids) THEN
+        i_start = ims
```
dudhia, line 939: I wonder about ims here, because ims includes the halo size. Is this deliberate? Can ips be used?

matzegoebel: I took this straight from set_physical_bc2d for consistency. But yes it shouldn't be necessary to include the complete halo. We could also write i_start = its -1, i_end = ite+1 (so without the max/min and if statements). Then all cases should be included as well. Is that what you had in mind?

dudhia:
I see now that this is a part of the code that is explicitly dealing with
halo cells at the outer boundary,
so it is OK, especially as bc2d does it. I will likely go ahead and approve
this PR soon.

dudhia (approved):

weiwangncar:
@matzegoebel Can you clarify what boundary condition or combination of boundary conditions are corrected in this PR (it was wrong before), and what are made cleaner and consistent?

matzegoebel:
> @matzegoebel Can you clarify what boundary condition or combination of boundary conditions are corrected in this PR (it was wrong before), and what are made cleaner and consistent?

I now tried to specify this in the overview description and release statement. But I'm not sure if it is clear now.

matzegoebel:
I fixed a problem in the symmetric BC, but the open BC also look a bit suspicious:
https://github.com/wrf-model/WRF/blob/12cd9b46cfe26ae400bcf335ec098b8a5276f77c/share/module_bc.F#L895-L920
and also later for the y-boundaries.

Instead of checking for the value of istag, only variables u and x are considered (so not the deformation d). There is even a comment from JM in 2002 which casts quite some doubt on that code section. Someone should have a look at this.

davegill (approved):

davegill:
@weiwangncar @dudhia @matzegoebel 
Folks,
This code is either causing or has uncovered an index bug. This PR went into the release-v4.2.2 branch, and it caused no troubles there. All of the jenkins regression tests passed for this PR going into release-v4.2.2 (and I verified that those tests still do pass). However, because of our gitflow strategy, we merge the minor release branches (because they are bug fixes) onto the develop branch. That is where the problem pops up.

After that merge, this PR breaks the bit-for-bit comparison for the real*8 super-cell case when comparing serial vs OpenMP, and only for a single particular case. 

Due to this bit-for-bit difference, any PR to the develop branch (and they are ALL to the develop branch now) will say that the jenkins testing failed.

Options:
1. We rip this code out. I do not recommend that.
2. We go through the namelist options that fail and pass to determine the cause. Long-term, yes I agree. But we need an immediate solution.
3. Drop the offending test from the jenkins suite during an interim period. I think that we have to go with this one. I do not have time to track down a bit-wise diff, particularly as it cannot be replicated on a Mac or on our classroom linux cluster.

matzegoebel:
Since the error doesn't occur in the Jenkins testing for this PR but only for the develop branch, it seems more likely that this PR rather uncovered a problem in one of the other new commits on the develop branch. Is it possible to include this PR right where the develop branch diverged from the master branch, check whether the tests pass and then, if they do, include all other commits on the develop branch one by one to see where the errors start ocurring? This could be done in a separate test branch by cherry-picking form the develop branch. The supercell case uses open boundary conditions, right? So the errors cannot be related to changes made for periodic and symmetric BC in this PR, but must have to do with the changes starting at line 936, where the start and end points of the loops are changed.

dudhia:
Is it only one test that failed? What could be the distinction between this
and the other tests?
If we need a quick thing to keep things going, I agree with the Jenkins
by-pass for now,
but we need to track it down and Matthias has said one way which is similar
to how
we do these, but that can be time-consuming. Would be good to know if
merging this into earlier
develop versions works.

davegill:
Jimy,
The only failure is with a single case of real*8 quarter_ss, OpenMP vs Serial. Here are the QSS cases that used to be run: 06 (left), 08 (middle), and 09 (right). This has now been reduced to testing only case 08 and 09.

Screen Shot 2021-01-22 at 12 47 47 PM

The differences are damp_opt, the advection option, the hydrostatic flag, and boundary conditions.

The big problem is that we can't test this on our Macs or on our Linux classroom machines. The differences do not appear there. The failure mode is ONLY on a jenkins server running on the AWS cloud. 

The cherry picking idea is valid, this mean the testing needs to be done via pull requests. That is what all of this nonsense was for:
Screen Shot 2021-01-22 at 12 52 56 PM

It takes about 35 minutes once a PR is submitted. You run multiple tests to make sure it is a robust response. If we only go back to v4.2.1, we would need about 6 binary cuttings in our git log output to determine the problematic code. However, the earlier develop branch (circa 4.2.1) does not work with the current set of namelist options. There is actually no reason to believe that the bug we are looking for was introduced within the last couple of months, though. The time out for the tests that hang is about 90 minutes. We need a way to determine if the failure reported by jenkins is due to insufficient bug fixes from the release-v4.2.2 branch or a legit interaction with the cherry picked corner point mods.

dudhia:
My suspicion there would be tke_adv_opt=4 but I would be surprised if
anyone uses it.
This is positive definite 5th order WENO advection. We could test other
tke_adv_opt choices like 3
which is also WENO. Also interested to know if *_adv_opt=4 is used in any
other tests for
other variables.
Jimy

PR #1338, opened 2020-12-15
Avoid double staggering of ww in vertical advection of geopotential

matzegoebel:
TYPE: enhancement

KEYWORDS: geopotential,vertical advection

DESCRIPTION OF CHANGES:
The geopotential equation in WRF is formulated in the following (advective) form where staggering/destaggering is 
marked with overbars:

The vertical advection term thus employs a double staggering of Omega. First, Omega is destaggered to mass levels 
and multiplied with the phi gradient. Then the product is staggered back to w-levels. This formulation is analogous to the horizontal terms, but the double averaging of Omega may introduce additional inaccuracies and is not necessary. 

The phi gradient can be first staggered to w-levels and then multiplied with the unaveraged Omega:
 
For consistency and to avoid numerical instabilities, the same thing has to be done in the acoustic step.

The new namelist option `phi_adv_z` allows switching between the two formulations: 
`phi_adv_z = 1`: old formulation (default)
`phi_adv_z = 2`: new formulation

LIST OF MODIFIED FILES: 
Registry/Registry.EM_COMMON
dyn_em/module_big_step_utilities_em.F
dyn_em/module_small_step_em.F
run/README.namelist 

TESTS CONDUCTED: 
Idealized test cases em_les and em_hill2d_x. 
The vertical velocity is affected by the modification. Updrafts and downdrafts in the em_hill2d_x case become ~10% 
stronger with the new formulation:
W_xz
In the LES case stronger updrafts and downdrafts occur more frequently:
hist_w

I came across this issue when trying to transform advective components of temperature, humidity, and momentum into 
Cartesian form in a consistent way, using w for the vertical advection. Without the proposed modifications I could not 
close the budget, because the vertical advection of phi was not consistent with the vertical advection in the other budget 
equations.

RELEASE NOTE: A new namelist option is included that allows a user to avoiding double stagger averaging of Omega in the vertical advection of geopotential. The new namelist option phi_adv_z allows switching between the two formulations: phi_adv_z = 1 (the old formulation, which remains the default), and phi_adv_z = 2 (the new formulation).

### DIFF
--- a/Registry/Registry.EM_COMMON
+++ b/Registry/Registry.EM_COMMON
@@ -2662,6 +2662,7 @@ rconfig   integer  chem_adv_opt           namelist,dynamics	max_domains    1    
 rconfig   integer  tracer_adv_opt         namelist,dynamics     max_domains    1       rh    "tracer_adv_opt"        "positive-definite RK3 transport switch"      ""
 rconfig   integer  scalar_adv_opt         namelist,dynamics	max_domains    1       rh    "scalar_adv_opt"        "positive-definite RK3 transport switch"      ""
 rconfig   integer  tke_adv_opt            namelist,dynamics	max_domains    1       rh    "tke_adv_opt"           "positive-definite RK3 transport switch"      ""
+rconfig   integer  phi_adv_z              namelist,dynamics	max_domains    1       h     "phi_adv_z"             "Vertical advection option for geopotential; 1: original 2: avoid double staggering of omega"   ""
 # switches for selectively deactivating 2nd and 6th order horizontal filters for specific scalar variable classes
 rconfig   logical  moist_mix2_off         namelist,dynamics	max_domains    .false. rh    "moist_mix2_off"        "de-activate 2nd-order horizontal mixing for moisture"      ""
 rconfig   logical  chem_mix2_off          namelist,dynamics	max_domains    .false. rh    "chem_mix2_off"         "de-activate 2nd-order horizontal mixing for chem species"  ""
--- a/dyn_em/module_big_step_utilities_em.F
+++ b/dyn_em/module_big_step_utilities_em.F
@@ -1458,8 +1458,7 @@ SUBROUTINE rhs_ph( ph_tend, u, v, ww,               &
 
      DO k = 2, kte
      DO i = its, itf
-          wdwn(i,k) = .5*(ww(i,k,j)+ww(i,k-1,j))*rdnw(k-1)               &
-                        *(ph(i,k,j)-ph(i,k-1,j)+phb(i,k,j)-phb(i,k-1,j))
+          wdwn(i,k) = rdnw(k-1)*(ph(i,k,j)-ph(i,k-1,j)+phb(i,k,j)-phb(i,k-1,j))
      ENDDO
      ENDDO
 
@@ -1468,7 +1467,7 @@ SUBROUTINE rhs_ph( ph_tend, u, v, ww,               &
      DO k = 2, kte-1
      DO i = its, itf
            ph_tend(i,k,j) = ph_tend(i,k,j)                           &
-                             - (fnm(k)*wdwn(i,k+1)+fnp(k)*wdwn(i,k))
+                             - ww(i,k,j)*(fnm(k)*wdwn(i,k+1)+fnp(k)*wdwn(i,k))
      ENDDO
      ENDDO
 
--- a/dyn_em/module_big_step_utilities_em.F
+++ b/dyn_em/module_big_step_utilities_em.F
@@ -1454,22 +1454,42 @@ SUBROUTINE rhs_ph( ph_tend, u, v, ww,               &
 
 ! advective form for the geopotential equation
 
+   !  RHS term 3 is: - omega partial d/dnu(phi)
+
    DO j = jts, jtf
+      IF (config_flags%phi_adv_z == 2) THEN
+         !  First get staggered partial d/dnu(phi) and then multiply with omega
+         !  to avoid double staggering of omega
 
-     DO k = 2, kte
-     DO i = its, itf
-          wdwn(i,k) = rdnw(k-1)*(ph(i,k,j)-ph(i,k-1,j)+phb(i,k,j)-phb(i,k-1,j))
-     ENDDO
-     ENDDO
+         DO k = 2, kte
+         DO i = its, itf
+              wdwn(i,k) = rdnw(k-1)*(ph(i,k,j)-ph(i,k-1,j)+phb(i,k,j)-phb(i,k-1,j))
+         ENDDO
+         ENDDO
 
-!  RHS term 3 is: - omega partial d/dnu(phi)
+         DO k = 2, kte-1
+         DO i = its, itf
+              ph_tend(i,k,j) = ph_tend(i,k,j)                           &
+                                - ww(i,k,j)*(fnm(k)*wdwn(i,k+1)+fnp(k)*wdwn(i,k))
+         ENDDO
+         ENDDO
+      ELSE
+         !  First destagger omega and multiply with partial d/dnu(phi), then stagger the product
 
-     DO k = 2, kte-1
-     DO i = its, itf
-           ph_tend(i,k,j) = ph_tend(i,k,j)                           &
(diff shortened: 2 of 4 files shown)


davegill:
@dudhia @weiwangncar 
Folks,
Would you guys review this please

dudhia:
also should bring in Bill Skamarock (github name?).

dudhia:
I will say it looks OK and this discretization is not detailed in the ARW Tech Note, so nothing to modify there.

weiwangncar:
@skamaroc Bill, can you take a look at this PR? Thanks!

dudhia:
@matzegoebel can you indicate how much effect this has on results? Is it random noise or systematic in the diffs?

dudhia:
Also would be helpful to show the change in discretization as an equation (not sure how but something like)
overbar [ overbar(w) (d phi d eta) ] 
changed to
w overbar (d phi d eta)
where overbar is the vertical average.

davegill:
@dudhia @matzegoebel 
Jimy,
Of all of the WRF contributors, you have asked the right person to show you the discretized equations!

dudhia:
I know github can do them, just don't know how, but I think Matthias does.
This would also be easier to show Bill who has not given his view yet,
Jimy

matzegoebel:
I updated the commit message to include the corresponding equations.

matzegoebel:
I ran the em_les test case with and without the modifications with 1 min output interval. Here are the histogram plots of the geopotential difference at three different heights:
hist_ph

Judging from these plots, the differences seem to be random rather than systematic. Or do you have something else in mind that I should look at?

matzegoebel:
When averaging over x and y the differences become quite small:
scatter_ph

dudhia:
Thanks for adding the equations. I was just thinking of an ncdiff plot to show a sample effect which would illustrate noise versus systematic change. If there is any evidence that the change does not introduce noise, that would be great too, since it now uses a less vertically averaged w.

matzegoebel:
I'm not quite sure what you mean exactly. Here are instantaneous time-height difference plots for different xy locations:
pcolor_ph
I cannot really see much in there. Only that there is some feature at bottom_top_stag=20 which is the boundary layer height.
And here is a kernel density estimation of the perturbation from the xy-average for all data points:
hist_all
The distributions are almost identical, so I think there should not be more noise (i.e. broader PDF) with the new version.

dudhia:
Initial feedback from Bill is that this should not go into a bug fix
release just in principle as it is not a bug, so let's
target this towards the develop branch for V4.3 while Bill thinks about it
further.

dudhia:
I have moved this to a develop branch PR and removed the bug label

matzegoebel:
Oh yes that's right, it's not actually a bug. I also changed the base branch of my PR.

dudhia:
Do you do any real-data testing of this change? Bill Skamarock is interested in the divergence field probably because it shows noise best.

matzegoebel:
Do you mean the grid variable "div"? I could output that for the LES case again if you want. No, I haven't done any real-data testing, yet. I'm a bit hesitant, I must say, because I've actually never run WRF in real-case mode.
Would you want me to run the WRF testing framework or what do you have in mind?

dudhia:
I am relaying things Bill has mentioned. Initially the easiest thing is to
assess the
noise level in an idealized case using the absolute change in mu for each
time step.
This is a measure of integrated divergence. If you can output div, we can
look at that too.
So we can start with that. If real-data testing is needed, someone at this
end can help.

matzegoebel:
Ok. I repeated the simulations with output at every timestep. Here are the KDE plots for the change of MUT from one timestep to the next. The new version has a slightly lower standard deviation.
mu_kde

dudhia:
Thanks for this. Which test case is this? Another useful thing is the time
series
of the average absolute delta mu per time-step, if possible.

matzegoebel:
Sure! Here is the full time-series of the absolute mu change averaged over xy and a close-up on the first 10 minutes:
dmu_timeseries
dmu_timeseries_start

matzegoebel:
it's the em_les test case

dudhia:
I see no problem there. Thanks.

davegill:
@dudhia @weiwangncar @matzegoebel 
Folks,
Is there a status on this yet? Matthias has provided about a zillion plots and equations.

dudhia:
Good question. I don't think Bill has seen the noise-level plots yet
and we need to ping him on it.

dudhia:
@matzegoebel I have heard back from Bill Skamarock who wants to be cautious but he is OK with this being added via a switch, so that people can test it more for themselves before we commit to it. I think a sample difference xy plot of fields at some time into the simulation may also be useful to post. Do you expect a random difference pattern in fields?

matzegoebel:
ok sounds good. Shall I include the switch and add it to the user guide and tech notes or will someone else do this? What fields are you referring to? I would guess that the differences are random, but I'm not sure.

dudhia:
Great if you can include the switch and add it to the README.namelist. We
can deal with
the other docs separately. Need to decide on a name and default setting.
A field like w and phi itself would be good to look at just some snapshots.
Maybe both
full fields and the diff posted together.

matzegoebel:
I included a switch called phi_adv_z with default value 1 (original formulation). Do you think that's an appropriate name?
Here are xy plots of phi perturbation and w after 1 hour integration at k=5 (test case em_les again). Looks pretty random but that's hard to say without an in-depth statistical analysis. Let me know if I can do anything else.
ph_xy
w

dudhia:
With such random flows it is difficult, and a more steady flow like the 2d
hill case
may be more instructive to display this way if you can. Thanks.

matzegoebel:
I just figured that I didn't pay attention to get the same initial perturbations for the two em_les runs. Now I manually set the random seed to achieve this. I reproduced the plots that I posted so far. But it doesn't make a big difference. The histogram plots of ph difference get a bit narrower.
hist_ph
scatter_ph
pcolor_ph
hist_all
dmu_timeseries
dmu_timeseries_start
w
ph_xy

matzegoebel:
> With such random flows it is difficult, and a more steady flow like the 2d hill case may be more instructive to display this way 

Yes, good point. Here are xz difference plots for the em_hill2d_x case (xy plots don't make sense since it's 2D) after 10 hours. The updrafts and downdrafts are about 10% stronger with the new version. This makes sense since the new version employs less averaging for omega. So the modification indeed has some systematic and non-negligible impact.
PH_xz
W_xz

:

matzegoebel:
I also pushed the commit now that includes the switch

dudhia:
Thanks, Matthias, that is interesting. 10% is fairly significant but your
idea seems reasonable.
PH shows fractionally less difference.
I am cc'ing Bill Skamarock.
Jimy

davegill:
@matzegoebel @dudhia 
There is a blocky difference at the top of the atmosphere for both ph and w, but more pronounced for ph. Since it is a difference, that points to either the previous or current computation as problematic, correct?

dudhia:
Probably was run without damp_opt. Not a problem.

matzegoebel:
The damping options for that case are:
 damp_opt                            = 3
 zdamp                                 = 20000. ( ztop = 30000.)
 dampcoef                            = .1

I also thought this looks suspicious. But I guess the damping is too weak to damp away all the waves leading to some strange behavior at the top? But I'm not so familiar with Rayleigh damping to really understand this. If Jimy says it's ok, then I guess it is.

dudhia:
I don’t think it is a case of the new one being worse just different.

matzegoebel:
yes, I think it has to do with omega=0 at k=kts which introduces a jump if not all waves have been damped away.
I also looked again at the em_les case to check the updraft velocity distributions. The difference between the histogram densities of the new and old version shows again the stronger up and downdrafts in the new version.
hist_w

dudhia:
This looks fine. The commit message needs to be updated with the addition of the phi_adv_z switch and extra modified files. The new method is 2 and the old one is 1. 1 remains the default.

matzegoebel:
ok I updated the commit message. I also included the last two figures.

dudhia (approved):

matzegoebel:
@davegill will you add this option to the User Guide?

dudhia:
@Kelly Gmail ***@***.***>  can this be done?

weiwangncar:
@matzegoebel @dudhia We can add this in UG.

PR #1353, opened 2020-12-29
Remove UM trajectory initialization from real

smileMchen:
TYPE: feature removed

KEYWORDS: UM trajectory, NCAR ACOM trajectory

SOURCE: Internal

DESCRIPTION OF CHANGES: 
The UM trajectory option has been replaced by NCAR ACOM trajectory option.  Therefore, the initialization for UM 
trajectory (in the real program) is no longer needed.

LIST OF MODIFIED FILES:

TESTS CONDUCTED: 
(1) A single test case has been conducted to demonstrate that the UM trajectory no longer works. NCAR ACOM trajectory option gives expected results. 
(2) Jenkins test is done successfully.

RELEASE NOTE: Not necessary

### DIFF
--- a/dyn_em/module_initialize_real.F
+++ b/dyn_em/module_initialize_real.F
@@ -4251,44 +4251,6 @@ endif
           END DO
       END IF
 
-      !  Template for initializing trajectories.  The i, j, and k starting locations
-      !  are specified.  Right now, a small plane in the middle of the domain is
-      !  selected.
-      
-      grid%traj_i    = -9999
-      grid%traj_j    = -9999
-      grid%traj_k    = -9999
-      grid%traj_lat  = -9999
-      grid%traj_long = -9999
-    
-      IF (config_flags%num_traj .gt. 0 .and. config_flags%traj_opt .gt. 0) THEN
-         icount = 1
-         DO j = (jde + jds)/2 - 2, (jde + jds)/2 + 2, 1
-            DO i = (ide + ids)/2 - 2, (ide + ids)/2 + 2, 1
-               IF ( its .LE. i    .and. ite .GE. i   .and.  jts .LE. j    .and. jte .GE. j ) THEN
-                  grid%traj_i   (icount) = i
-                  grid%traj_j   (icount) = j
-                  grid%traj_k   (icount) = 10
-                  grid%traj_lat (icount) = grid%xlat(i,j)
-                  grid%traj_long(icount) = grid%xlong(i,j)
-               END IF
-         
-#ifdef DM_PARALLEL
-               grid%traj_i   (icount) = wrf_dm_max_real ( grid%traj_i   (icount) )
-               grid%traj_j   (icount) = wrf_dm_max_real ( grid%traj_j   (icount) )
-               grid%traj_k   (icount) = wrf_dm_max_real ( grid%traj_k   (icount) )
-               grid%traj_lat (icount) = wrf_dm_max_real ( grid%traj_lat (icount) )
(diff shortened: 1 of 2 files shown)


smileMchen:
@davegill @weiwangncar @dudhia @kkeene44 
All,
Since WRFv3.9.1, the UM trajectory option has been replaced by the NCAR ACOM trajectory option. While the ACOM trajectory option works fine, the UM trajectory does have its advantages. 
Dave and I discussed this issue whether to recover the UM trajectory option. We are not sure yet how much coding work will involve. Our initial attempt failed due to problems possibly related to memory and MPI issues. We don't know yet the exact reason. 
We leave the discussion to the group whether to keep both trajectory options, or it is just fine to stay with the NCAR ACOM option.

dudhia:
Can you say what the advantages of the UM option are?
Thanks.

smileMchen:
@dudhia 
Jimy,
The UM trajectory is easily initialized and the output can be controlled by the history interval. 
The ACOM option needs wrfinput_traj as input and output is every time step.

davegill:
The original simple UM trajectory:
1. just i,j,k location
2. initialized in real

The new ACOM trajectory:
1. values of user selectable fields at each trajectory point
2. initialization is via lat/lon and a separate input file

dudhia:
Seems worth a try in my opinion

davegill:
Folks,
Ming pointed me to the original PR
https://github.com/wrf-model/WRF/pull/84

There was no attempt to provide traj_opt=2 and keep the original traj_opt=1. There are lots of places where the original trajectory code was overwritten. It would be a large effort to bring back the original capability.

weiwangncar:
@dudhia @davegill @smileMchen Apart from initialization, the ACOM code can compute meteorological values along the trajectory, which was the major enhancement.

dudhia:
Maybe just making it easier to use or having reasonable defaults and
example files is sufficient.

davegill:
Folks,
None of the new trajectory is handled via the registry. A number of mods overwrote the original code. There is no way to tease out traj_opt=1 vs traj_opt=2 (for example), they are all mixed together. I ran a simple trajectory. I got all of the info for the ACOM trajectory. However, I also got the original data in the standard wrfout_d01* file.

This seems like quite an effort to split them apart. This PR cleans up code that is no longer used. I think that should be done, as users are going to be very confused why no trajectories are initialized after they have put them into the real program.

Do we have utilities to process the ACOM traj output files?

weiwangncar:
No, we don't. We didn't have anything to plot the UM trajectories either, at least not that I'm aware of. I've already changed my Option talk to reflect the new option. Also should we leave this PR to 4.3?

davegill:
@weiwangncar @smileMchen @dudhia @kkeene44 
> should we leave this PR to 4.3?

That would give us time to decide what to do. However, I think that the decision is clear - UM traj is gone, and ACOM traj is the way we operate.

dudhia:
I agree with this. We need to try to make the ACOM trajectories easy to use
with some default settings and example files that work.
Jimy

davegill:
@smileMchen @weiwangncar @dudhia 
Folks,
Regardless of the documentation status for the ACOM trajectory, this PR should be merged. It gets rid of code that would be confusing, as it is not utilized. If we are in agreement, I'll approve it.

weiwangncar:
@davegill @smileMchen @dudhia I'm ok with it...

davegill (approved):

matzegoebel:
In module_initialize_ideal.F the template for initializing trajectories is still there. Shouldn't it also be removed then?

PR #1355, opened 2020-12-30
Modifications to improve WRF-CTSM coupling

negin513:
TYPE: enhancement

KEYWORDS: WRF-CTSM coupling, Community Terrestrial Systems Model (CTSM), Community Land Model (CLM),
coupled land-atmosphere models

SOURCE:  Negin Sobhani <sup>1</sup>  , Dave Lawrence <sup>1</sup>  , Sam Levis <sup>2</sup> , Bill Sacks<sup>1</sup>

<sup>1</sup> National Center for Atmospheric Research (NCAR)
<sup>2</sup> SLevis Consulting 

DESCRIPTION OF CHANGES:
This PR includes some modifications that improve WRF-CTSM coupling including the capability to handle lake points by CTSM instead of WRF.

Problem:
Previously lake points were not handled by the CTSM model; however, the CTSM community prefers lake points to be handled by CTSM lake model. This PR adds the capability to handle lake points by CTSM model. 

Solution:
For solving this problem in `module_surface_driver.F`, we used ~~`xland(i,j) .gt. 1.5 .and. ht(i,j).gt.lake_min_elev`~~ `lakemask(i,j).eq.1.`  logic. For CTSM cases we changed a local copy of `xland` (called `ctsm_xland`) for the lake points to 1, so that CTSM receives the lake points. 

LIST OF MODIFIED FILES: 
M	   share/module_soil_pre.F

TESTS CONDUCTED: 
1. Following are a couple of figures, depicting a particular period's lakemask over CONUS, for both the input and theWRF model output. Specifically, the Great Lakes were part of the concern with this PR, and they are indeed represented as part of the lake identified cells.

Model input:
image

Model output:
image

2. Figures of low-level temperature and moisture for a 1-week simulation, without spectral nudging.

2-meter temperature: Here is the plot showing ΔT2 for the first week of 2013-04. ΔT2 is old version-new version. 
The new version contains the local xland mods requested by the WRF committee.
image

2-meter RH. There is ~ a 5-10% difference in RH at 2m between the old and new versions over all the inland lakes.
ΔRH2 is old version-new version. This means that over the lakes, the new version is 5-10% dryer than the old version. 
image

3. Jenkins tests are passing.


davegill:
@negin513 @mgduda @dudhia @weiwangncar 
Negin,
Michael has had a "lake identifier" index in the metgrid output for years. We turn that into just a water category in real for passing into WRF. We could provide a switch that controls an extra 2d array of a lake mask, which would be packaged with the existing CTSM option.

phys/module_surface_driver.F
```diff
@@ -3264,7 +3264,7 @@ SUBROUTINE surface_driver(                                         &
                IF(PRESENT(ACGRDFLX))ACGRDFLX(I,J)=ACGRDFLX(I,J) + GRDFLX(I,J)*DT
 
                ! Check that SFCDIAGS can declare these as intent(out)
-               T2(I,J)  = -1.E36
+               T(I,J)  = -1.E36
```
davegill, line 3267: Negin,
Does this exist?
There is a 3d t_phy.

dudhia:
I had an email response that seems to have failed to be delivered to github. This was as follows.
Is this not using the lake mask because you want to include the Great Lakes?
The reason for not using lake mask could be mentioned.

negin513:
> I had an email response that seems to have failed to be delivered to github. This was as follows.
> Is this not using the lake mask because you want to include the Great Lakes?
> The reason for not using lake mask could be mentioned.

Thanks @dudhia  for your comment.
We are still using the lake mask from WRF, but we are letting CTSM handle lake points. 

For more clarification, I've added a few lines of comments to the code. 
https://github.com/wrf-model/WRF/blob/1acbe6a4dd6a0dc2bd10c4b708647de90be4fa39/phys/module_surface_driver.F#L3898-L3901

Specifically about the great lakes, before they were treated as ocean but now they are going through CTSM lake model.

phys/module_surface_driver.F
negin513: @davegill Thanks for pointing this out. This should have been T2 and unchanged. I've fixed in the module.

share/module_soil_pre.F
```diff
@@ -224,7 +224,7 @@ SUBROUTINE adjust_for_seaice_post ( xice , landmask , tsk_old , tsk , ivgtyp , v
 
          CASE ( SLABSCHEME )
 
-         CASE ( LSMSCHEME , NOAHMPSCHEME , CLMSCHEME, SSIBSCHEME )   !mchen add for ssib
+         CASE ( LSMSCHEME , NOAHMPSCHEME , CLMSCHEME, CTSMSCHEME, SSIBSCHEME )   !mchen add for ssib
```
davegill, line 227: @negin513 
Negin,
Likely this CTSMSCHEME (and at line 3219) are causing NMM troubles. This probably needs to be defined in the file Registry/Registry.NMM via a package. You can make sure that NMM never uses it (if you want) with a test in check_a_mundo.

negin513: @davegill can I use instead? I think this would be simpler and we used WRF_USE_CTSM in other places in the code. 

```Fortran
#ifdef WRF_USE_CTSM 
CASE ( LSMSCHEME , NOAHMPSCHEME , CLMSCHEME, CTSMSCHEME, SSIBSCHEME )   !mchen add for ssib
#else
CASE ( LSMSCHEME , NOAHMPSCHEME , CLMSCHEME, SSIBSCHEME )   !mchen add for ssib
#endif

Comment (davegill) [reply share/module_soil_pre.F:227]: Negin,
You bet. That would be a fine way to do it. 
Also ifdef the location at the top of the surface driver where the CTSMSCHEME is pulled in from module_state_description (if not done already).

Comment (negin513) [reply share/module_soil_pre.F]: Dave, 
Thanks for your comment. In the latest push, I made this change (adding #ifdef WRF_USE_CTSM), which resolved the issue with the tests. 

Also, in the surface driver, we have a similar logic to not use CTSMSCHEME for NMM cores. 
https://github.com/wrf-model/WRF/blob/1acbe6a4dd6a0dc2bd10c4b708647de90be4fa39/phys/module_surface_driver.F#L298-L344

Thanks for your help with this.

File: phys/module_surface_driver.F [hunk 3]
```diff
@@ -4045,6 +4059,27 @@ SUBROUTINE surface_driver(                                         &
...
+            IF(PRESENT(ACLHF))ACLHF(I,J)=ACLHF(I,J) + LH(I,J)*DT
+            IF(PRESENT(ACGRDFLX))ACGRDFLX(I,J)=ACGRDFLX(I,J) + GRDFLX(I,J)*DT
+         ENDDO
+      ENDDO
+
+        !CALL SFCDIAGS(HFX,QFX,TSK,QSFC,CHS2,CQS2,T2,TH2,Q2,      &
+        !           PSFC,CP,R_d,RCP,CHS,t_phy,qv_curr,ua_phys,    &
+        !           ids,ide, jds,jde, kds,kde,                    &
+        !           ims,ime, jms,jme, kms,kme,                    &
+        !            i_start(ij),i_end(ij), j_start(ij),j_end(ij), kts,kte    )
```
davegill, line 4077: @negin513 
If these are commented out, let's just remove them.

phys/module_surface_driver.F
```diff
@@ -3895,6 +3895,20 @@ SUBROUTINE surface_driver(                                         &
 
 
 !------------------------------------------------------------------
+        !For WRF-CTSM simulations, we would like the land model (CTSM)
+        ! to handle inland lake points.
+        ! Here, we are changing xland to include lake points, so that
+        ! CTSM can handle it. 
+         DO j=j_start(ij),j_end(ij)
+            DO i=i_start(ij),i_end(ij)
+                IF (xland(i,j) .GT. 1.5.and.ht(i,j).gt.lake_min_elev) then
```
davegill, line 3904: @dudhia @weiwangncar @smileMchen @negin513 
Folks,
Is there not an actual WRF lake flag that can be used? I thought that we had a 2d lake flag.

phys/module_surface_driver.F
negin513: @davegill thanks for your comment. I just removed these commented out lines in the recent commits. 
The reason I kept it there in the first place was to keep it consistent with other LSMs. For example, for NOAH-MP these lines are just commented out and not removed: 
https://github.com/wrf-model/WRF/blob/ca7092a948d327104eb0ae3b0224a70f3b21c2bc/phys/module_surface_driver.F#L3306-L3312
But, I agree with you that removing them makes more sense.

phys/module_surface_driver.F
negin513: @davegill 
Thanks for your comment.  I incorporated ``lakemask(i,j).eq.1. `` in this conditional, so now it looks like this:

https://github.com/wrf-model/WRF/blob/15981e1a283516c8bd151ad77457a9e4f0c92c5c/phys/module_surface_driver.F#L3904-L3908

I am not sure how to interpret the variable ``lake_min_elev``. Does that refer to the minimum lake depth for considering a point "lake"?

I've noticed it is used in different places in module_surface_driver.F and other modules where I assumed it is used as an indicator for defining lake points.

davegill: @negin513 @smileMchen @dudhia @weiwangncar 
Negin,
The value `lake_min_elev` has a default value set to 5 m. The idea is IF this is a water point AND the elevation of the grid cell is >= this minimum elevation, THEN this must be a lake point. This is likely an older way of identifying a lake point. 

There is a field called `lakemask`. We have an option for activating some lake physics, `sf_lake_physics`. 

Ming worked with these features.

phys/module_surface_driver.F
```diff
@@ -3895,6 +3895,20 @@ SUBROUTINE surface_driver(                                         &
...
+        !For WRF-CTSM simulations, we would like the land model (CTSM)
+        ! to handle inland lake points.
+        ! Here, we are changing xland to include lake points, so that
+        ! CTSM can handle it. 
+         DO j=j_start(ij),j_end(ij)
+            DO i=i_start(ij),i_end(ij)
+                IF (lakemask(i,j).EQ.1. .AND. xland(i,j).GT.1.5 .AND. ht(i,j).GT.lake_min_elev) THEN
+                    WRITE( message,* ) 'WRF-CTSM: for great lakes  xland ==1'
+                    CALL wrf_message ( message )
+                    xland (i,j)  = 1
```
dudhia, line 3907: Is lakemask=1 for the Great Lakes? This only works if it is, but I thought it would be 0 since these are too large to qualify for the lake mask.
Also does CTSM really need xland=1 to handle the Great Lakes? Does it have its own internal lake model? Normally xland=1 is only for the land and lakemask is used for lakes.

davegill:
Folks,
How do we get the `lakemask` field that I was recommending that Negin use?
In the Registry.EM, `LAKEMASK` is an `i1` for I/O, so it is supposed to come into real from metgrid.
I do not see this field in a few of the met_em* files that I have.
Do we have to input geogrid static data that includes lakes?

dudhia:
Yes, this mask should come with the correct geogrid MODIS choice.
21-category, I think.

weiwangncar:
@davegill @dudhia It is in real that the field LAKEMASK is created, and it is created based on LU_INDEX:
            IF ( grid%lu_index(i,j) .NE. grid%islake ) THEN
               grid%lakemask(i,j) = 0
            ELSE
               grid%lakemask(i,j) = 1
            END IF
So this appears to available in the model regardless of whether you use the lake option or not.

phys/module_surface_driver.F
dudhia: @negin513  still needs to answer my question here. Not sure if this is correct yet.

dudhia: I think our suggestion was to do this by not modifying xland and instead passing a temporary local parameter into the CTSM driver which is 1 for the lake points.

phys/module_surface_driver.F
```diff
@@ -3895,6 +3896,20 @@ SUBROUTINE surface_driver(                                         &
...
 
 !------------------------------------------------------------------
+        !For WRF-CTSM simulations, we would like the land model (CTSM)
+        ! to handle inland lake points.
+        ! Here, we are changing xland to include lake points, so that
+        ! CTSM can handle it.
+         DO j=j_start(ij),j_end(ij)
+            DO i=i_start(ij),i_end(ij)
+                xland_ctsm (i,j) = xland (i,j) 
+                !IF (lakemask(i,j).EQ.1. .AND. xland(i,j).GT.1.5 .AND. ht(i,j).GT.lake_min_elev) THEN
```
dudhia, line 3906: Can we just remove this commented line?

phys/module_surface_driver.F
negin513: I am going ahead to resolve this.

davegill:
@negin513 
Negin,
Do you have any figures that show the impact of with vs without these mods in your CTSM system? 

Even small diffs show:
1. The coupled model is stable
2. The diffs do indeed make an impact

phys/module_surface_driver.F
negin513: It seems like lakemask=1 for the Great Lakes. I am attaching the plots from wrfout that I showed you in the previous meeting which shows Great Lakes in the lakemask:
image

This shows the lakemask for wrfinput:
image


I am putting these here only for the sake of future reference.

phys/module_surface_driver.F
negin513: @dudhia Thanks! I just removed this extra commented-out line. I will go ahead and mark this as resolved.

phys/module_surface_driver.F
negin513: > I think our suggestion was to do this by not modifying xland and instead passing a temporary local parameter into the CTSM driver which is 1 for the lake points.

@dudhia Correct! The suggestion by the WRF committee was that instead of directly changing `xland` variable, we should make a local variable called `ctsm_xland` which has 1 for the lake points. Then we pass `ctsm_xland` instead of `xland` to CTSM.

dudhia (approved):

negin513:
@davegill 

Here is the plot showing ΔT2 for the first week of 2013-04:
image


ΔT2 is old version-new version. The new version contains the mods that WRF committee asked.  
p.s.: These results are WRF-CTSM not using spectral nudging.

negin513:
@davegill and @dudhia :
The following plot is similar to the above but for RH at 2m. There is ~ a 5-10% difference in RH at 2m between the old and new versions over all the inland lakes. Does that look okay to you? 
image

ΔT2 is old version-new version. The new version contains the mods that WRF committee asked. This means that over the lakes the new version is 5-10% dryer than the old version.

negin513:
> @davegill and @dudhia :
> The following plot is similar to the above but for RH at 2m. There is ~ a 5-10% difference in RH at 2m between the old and new versions over all the inland lakes. Does that look okay to you?
> image
> 
> ΔT2 is old version-new version. The new version contains the mods that WRF committee asked. This means that over the lakes the new version is 5-10% dryer than the old version.

Just for sanity check, I compared RH at 2m from both the old version and the new version of WRF-CTSM ( with and without recent mods) with NOAH-MP and this is the results:
Screen Shot 2021-03-05 at 4 01 57 PM

By the "old" version of WRF-CTSM, I mean the version where xland = 1 was changed to 1 over the lakes. The new version of WRF-CTSM,  we did not  modify xland and instead we used a temporary local parameter called (xland_ctsm) in the CTSM driver which is 1 for the lake points.

weiwangncar:
I'm ok with this PR.

negin513:
# Summary Part 1 
Here is a summary of discussions and thoughts on this PR. 
Since most parts of the discussions on this PR were in meetings and over calls rather than in the comments of this PR, I am making this summary for future reference. 

@dlawrenncar and  Fei Chen suggested that they would like CTSM lake model to handle lake points rather than WRF slab model. https://github.com/ESCOMP/CTSM/issues/1092

At first, we used the logic `xland(i,j) .gt. 1.5 .and. ht(i,j).gt.lake_min_elev .and. lakemask(i,j).eq.1. `in the `module_surface_driver.F` to identify lakes and changed `xland` to 1 for lake points.   In WRF-CTSM coupling infrastructure only land points are passed from WRF to CTSM; therefore, by changing `xland` (land mask) to include lake points we can pass lake points to CTSM. Similar logic for identifying lake was used in `module_surface_driver.F`.

At this point the code looked like the following:
https://github.com/wrf-model/WRF/blob/15981e1a283516c8bd151ad77457a9e4f0c92c5c/phys/module_surface_driver.F#L3898-L3910


Next, @davegill  recommended not using the first two parts of the logic (`xland(i,j) .gt. 1.5 .and. ht(i,j).gt.lake_min_ele`) As this was an older method for identifying lake points.

> The value lake_min_elev has a default value set to 5 m. The idea is IF this is a water point AND the elevation of the grid cell is >= this minimum elevation, THEN this must be a lake point. This is likely an older way of identifying a lake point.

After applying this comment to the code, it looks like the following:
```FORTRAN
        !For WRF-CTSM simulations, we would like the land model (CTSM)
        ! to handle inland lake points.
        ! Here, we are changing xland to include lake points, so that
        ! CTSM can handle it. 
         DO j=j_start(ij),j_end(ij)
            DO i=i_start(ij),i_end(ij)
                IF (lakemask(i,j).eq.1. ) THEN
                    WRITE( message,* ) 'WRF-CTSM: for great lakes  xland ==1'
                    CALL wrf_message ( message )
                    xland (i,j)  = 1  
                ENDIF
            ENDDO
         ENDDO
```
 

At this point, @dudhia suggested making sure that the Great Lakes are indeed included in the `lakemask`. 
For addressing this comment, I've provided plots from both wrfout and wrfinput showing that indeed `lakemask=1` for the Great Lakes. Copied from the above:

>  It seems like lakemask=1 for the Great Lakes. I am attaching the plots from wrfout that I showed you in the previous meeting which shows Great Lakes in the lakemask:
> image
> 
> This shows the lakemask for wrfinput:
> image
>

negin513:
# Summary Part 2
------------------------------------------------------
Next, we had a meeting with the WRF committee (@davegill , @dudhia ,@weiwangncar, and @smileMchen) for approving this PR. In this meeting, Ming Chen and @weiwangncar suggested not modifying `xland` and instead passing a temporary local parameter into the CTSM driver which is 1 for the lake points. They mentioned that `xland` might be used in other places in the code and it is not advised to change it directly. 

At this point, we introduced a new local parameter called `ctsm_xland`. `ctsm_xland` is exactly similar to `xland` but with 1 for the lake points. Next, we passed `ctsm_xland` to `ctsm_run` subroutine. 

I accepted this change as I initially thought the results will be BFB. 
However, the results showed differences compared to the original way (changing `xland` directly and passing it `ctsm_run`). 

After discussions with @slevisconsulting and @Davegill, I was convinced that the resulting change was normal and should have been expected. 

Not changing the original `xland`  and using the temporary `ctsm_xland` made sense as there might be other points in the code with special treatments over water points that still need to be done, and by modifying `xland` directly we are skipping those treatments. 

Below shows the difference between T2 and RH for one week in 2013-04:

> Here is the plot showing ΔT2 for the first week of 2013-04:
> image
> 
> 
> ΔT2 is old version-new version. The new version contains the mods that WRF committee asked.  
> p.s.: These results are WRF-CTSM not using spectral nudging. 

> The following plot is similar to the above but for RH at 2m. There is ~ a 5-10% difference in RH at 2m between the old and new versions over all the inland lakes. Does that look okay to you? 
> image
> 

> Just for a sanity check, I compared RH at 2m from both the old version and the new version of WRF-CTSM ( with and without recent mods) with NOAH-MP and this is the results:
> Screen Shot 2021-03-05 at 4 01 57 PM
> 
> By the "old" version of WRF-CTSM, I mean the version where xland = 1 was changed to 1 over the lakes. The new version of WRF-CTSM,  we did not  modify xland and instead we used a temporary local parameter called (xland_ctsm) in the CTSM driver which is 1 for the lake points.


--------------------------
This change was made because WRF will handle the transferred CTSM fluxes differently over water than over land points. 
We would like CTSM fluxes over lake points in the WRF.

negin513:
@davegill , @dudhia , @weiwangncar , @smileMchen, I would like to once again confirm if the followings are true with this change:

- With the current changes over the lake points, fluxes transferred from CTSM are being used in the WRF model. 
- Based on your comments, there are not any additional treatments done over the lake points that need to be turned off when we are directly using CTSM fluxes.

negin513:
For confirming that in the WRF-CTSM settings, CTSM values over lake points are being transferred correctly and not being over-written by WRF,  @dlawrenncar  suggested the following experiment:

**In this experiment, over the lake points in CTSM initial conditions, we set values higher than ordinary (For example 60 °C = 333.15 K). If the fluxes from CTSM over lake points are being used in WRF, we should see its impact in the WRF output files.** 

Following this experiment, we can clearly see the fluxes from CTSM over the lake points in the WRF output, which confirms that CTSM values over the lake points are not being over-written by WRF. 

Here I attached a few plots showing different variables.

First, the following plot shows upward heat flux at the surface (HFX) from wrfout files for a one-week of simulation (first week of 2013-04). Here we can clearly see the signals from CTSM initial conditions over lakes.

image


The following plot shows the temperature at 2m (T2) from wrfout files for a one-week simulation:
image


This plot shows skin surface temperature (TSK) from wrfout for one day of simulation:
image


**In conclusion, we can confirm that WRF-CTSM coupling is working correctly over lakes and fluxes from CTSM over the lake points are being used in WRF.**

PR #1387, opened 2021-01-27
Bug fix in Deng shcu parameterization for updraft top calculation

twjuliano:
TYPE: bug fix

KEYWORDS: Deng shcu, limit updraft, cloud

SOURCE: Timothy W. Juliano and Pedro A. Jimenez (NCAR/RAL), Jimy Dudhia (NCAR/MMM)

DESCRIPTION OF CHANGES:
Problem:
The issue is in `loop60` of the code. There are two exit statements within that loop. If either of these conditional statements are satisfied, then the loop exits and there is no bug. The problem arises if there is a clean exit of `loop60`. In this loop, variable `NK` (the index of the loop) goes from just below cloud base (`KLCL-1`) to #layers-1 (`KLM`). If there is a clean exit of the loop, then `NK` = #layers. This causes variable `LTOP1` to have a value of #layers and therefore `LTOP` also equals #layers. When this happens, in line 2250 of the official WRF release [`TMA = QQG(NK+1)*EMS(NK+1)`], QQG goes out of bounds since QQG is defined from 1 to #layers and `NK` in this loop goes from 1 to `LTOP` which is #layers. Thus, it tries to access index `NK+1` which is #layers+1.

Solution:
The solution is to limit the value of `LTOP1` at the exit of loop60 to #layers-1 (`KLM`).

LIST OF MODIFIED FILES:
M phys/module_shcu_deng.F

TESTS CONDUCTED: 
1. We have run a 3km CONUS simulation with and without the bug fix. Without the bug fix, the simulation crashes as 
reported above. With the bug fix, the simulation integrates through without an issue.
2. Jenkins tests all PASS.

RELEASE NOTE: A fix is introduced to the Deng shallow cumulus parameterization scheme. When the updraft top reached the domain top, the model would crash due to accessing out of bounds memory. The modification is to limit the vertical index of the updraft top.


davegill:
jenkins:
```
Please find result of the WRF regression test cases in the attachment. This build is for Commit ID: 4e0ef8d56020e27288ea3bd07ec833773a3aa464, requested by: twjuliano for PR: https://github.com/wrf-model/WRF/pull/1387. For any query please send e-mail to David Gill.

    Failed Simulations are: 
    None
    None
```

weiwangncar:
Do we need to have AJ Deng review this? I'm ok with this.

dudhia:
Should be OK not to. Just addresses an out-of-bounds that can happen
sometimes.

davegill:
@twjuliano @dudhia @weiwangncar 
Folks,
Everyone seems OK with this bounds fix. We should approve and merge.

davegill:
@twjuliano 
Tim,
Since you are doing lots of PRs now, I would ask a conventional item. 
Would you make all of the lines of the commit message (except for the RELEASE NOTE part) to be split lines.

This gives a much nice look to the commit message when viewed in an editor.

weiwangncar (approved):

twjuliano:
> @twjuliano
> Tim,
> Since you are doing lots of PRs now, I would ask a conventional item.
> Would you make all of the lines of the commit message (except for the RELEASE NOTE part) to be split lines.
> 
> This gives a much nicer look to the commit message when viewed in an editor.

@davegill Sure, I can do that. Since I type the message using the online interface, does that just require me to hit return at the end of line instead of automatically continuing the next line?

PR #1406, opened 2021-02-17
Remove requirement to run 'run_me_first' scripts for ideal cases

kkeene44:
TYPE: enhancement

KEYWORDS: ideal, run_me_first.csh, link, Makefile, 

SOURCE: Internal

DESCRIPTION OF CHANGES:
Problem:
Several ideal cases (em_b_wave, em_convrad, em_quarter_ss, em_scm_xy, em_seabreeze2d_x, and em_tropical_cyclone) 
need additional linked files to run. Previously it was necessary for users to run a script called 'run_me_first.csh' to 
link-in the necessary files.  

Solution:
Removed all 'run_me_first.csh' scripts from ideal case directories and modified the top-level Makefile so that now the 
files are linked during the compile.

LIST OF MODIFIED FILES: 
M       Makefile
D       test/em_b_wave/run_me_first.csh
D       test/em_convrad/run_me_first.csh
D       test/em_quarter_ss/run_me_first.csh
D       test/em_scm_xy/run_me_first.csh
D       test/em_seabreeze2d_x/run_me_first.csh
D       test/em_tropical_cyclone/run_me_first.csh

TESTS CONDUCTED: 
1. Ensured that code compiles and required files are linked for all modified ideal cases.
2. Jenkins is all pass.

RELEASE NOTE: It is no longer necessary to execute a 'run_me_first.csh' script for any of the idealized cases. All necessary physics input files and look-up tables are linked to the corresponding directory during the compilation step.


davegill:
I just added the new p3 data to the run_me_first scripts. The MP schemes tend to have / need extra data that a number of the idealized cases could use. I am not sure about the land surface model or the radiation schemes.

kkeene44:
@davegill 
Can you point me to the PR where you added that?

kkeene44:
@davegill 
I updated the Makefile to include all the p3 links. Thanks!

Makefile
```diff
@@ -229,7 +229,9 @@ em_quarter_ss : wrf
 	( cd test/em_quarter_ss ; /bin/rm -f kernels.asc_s_0_03_0_9 ; ln -s ../../run/kernels.asc_s_0_03_0_9 . )
 	( cd test/em_quarter_ss ; /bin/rm -f kernels_z.asc ; ln -s ../../run/kernels_z.asc . )
 	( cd test/em_quarter_ss ; /bin/rm -f masses.asc ; ln -s ../../run/masses.asc . )
-	( cd test/em_quarter_ss ; /bin/rm -f termvels.asc ; ln -s ../../run/termvels.asc . )
+	( cd test/em_quarter_ss ; /bin/rm -f termvels.asc ; ln -s ../../run/termvels.asc . ; \
+		ln -sf ../../run/RRTM_DATA . ; \
```
weiwangncar, line 233: @kkeene44 It looks like we don't need to link this file - it is not used by the default options.

Makefile
```diff
@@ -229,7 +229,9 @@ em_quarter_ss : wrf
...
+		ln -sf ../../run/p3* . )
```
weiwangncar, line 234: @kkeene44 @davegill The p3* data were not used by the default namelists, perhaps we should not link them.

Makefile
```diff
@@ -330,7 +332,9 @@ em_b_wave : wrf
 	( cd main ; $(MAKE) RLFLAGS="$(RLFLAGS)" MODULE_DIRS="$(ALL_MODULES)" SOLVER=em IDEAL_CASE=ideal em_ideal )
 	( cd test/em_b_wave ; /bin/rm -f wrf.exe ; ln -s ../../main/wrf.exe . )
 	( cd test/em_b_wave ; /bin/rm -f ideal.exe ; ln -s ../../main/ideal.exe . )
-	( cd test/em_b_wave ; /bin/rm -f README.namelist ; ln -s ../../run/README.namelist . )
+	( cd test/em_b_wave ; /bin/rm -f README.namelist ; ln -s ../../run/README.namelist . ; \
+		ln -sf ../../run/RRTM_DATA . ; \
```
weiwangncar, line 336: @kkeene44 @davegill The same comments apply here.

Makefile
```diff
@@ -396,7 +400,11 @@ em_seabreeze2d_x : wrf
 	( cd main ; $(MAKE) RLFLAGS="$(RLFLAGS)" MODULE_DIRS="$(ALL_MODULES)" SOLVER=em IDEAL_CASE=ideal em_ideal )
 	( cd test/em_seabreeze2d_x ; /bin/rm -f wrf.exe ; ln -s ../../main/wrf.exe . )
 	( cd test/em_seabreeze2d_x ; /bin/rm -f ideal.exe ; ln -s ../../main/ideal.exe . )
-	( cd test/em_seabreeze2d_x ; /bin/rm -f README.namelist ; ln -s ../../run/README.namelist . )
+	( cd test/em_seabreeze2d_x ; /bin/rm -f README.namelist ; ln -s ../../run/README.namelist . ; \
+		ln -sf ../../run/LANDUSE.TBL . ; \
+		ln -sf ../../run/RRTM_DATA . ; \
+		ln -sf ../../run/wind-turbine-1.tbl . ; \
+		ln -sf ../../run/p3* . )
```
weiwangncar, line 407: @kkeene44 Here too.

weiwangncar:
@davegill Unlike real data case directory, there is only limited physics files that linked to the idealized case directories, and they often depend on the default options used. If we are not linking all of the physics files, we should not add any particular ones, such as those for p3.

dudhia:
I agree with Wei. I don't want to have all the physics files linked into the ideal test directories and having run_me_first leaves the directory cleaner. So I am not sure about removing run_me_first.

kkeene44:
@dudhia @weiwangncar @davegill 
I can agree with not having all the physics files linked in; however, if a user is going to use an idealized case, they are instructed to run 'run_me_first' before using it, so I'm not sure why it needs to be a separate task for them if the files are going to be linked-in anyway before they run anything.

dudhia:
Few idealized cases have physics in their namelists. The sea-breeze case is
one
that is an exception. We could remove the need for run-me-first for the
cases
that always need it. Check which ones they are. Is it possible to link for
individual ideal cases?
Jimy

dudhia:
convrad is another one I can think of.

kkeene44:
@dudhia 
Yes, it is possible to link for individual cases. That's how I've set it up in the Makefile. I'm only linking the files that were in the 'run-me-first' scripts when the cases are built.

weiwangncar:
@kkeene44 @dudhia It looks like we have some agreement here. My other comments are saying that some of the files currently are linked by run_me_first.csh are not used, and they don't have to be linked. If you could go through each idealized case and see what default options are used, and only link those files, and that would be nice.

dudhia:
I think few of those don't really need the link to happen automatically as
the
default namelist has no physics.

kkeene44:
@dudhia @weiwangncar 
I went through and determined which files were needed for the default namelists and only linked those in the Makefile. I'm still removing the 'run_me' scripts.

davegill:
> the default namelist has no physics

We usually only care about input look-up tables for the LSM, MP, and radiation. Here are the idealized cases that have physics in those areas activated by default.

```
em_convrad/namelist.input: sf_surface_physics                  = 1,
em_scm_xy/namelist.input: sf_surface_physics                  = 2,
em_seabreeze2d_x/namelist.input: sf_surface_physics                  = 1,
em_tropical_cyclone/namelist.input: sf_surface_physics                  = 1,
```

```
em_convrad/namelist.input: mp_physics                          = 6,
em_quarter_ss/namelist.input: mp_physics                          = 1,     1,     1,
em_scm_xy/namelist.input: mp_physics                          = 2,
em_seabreeze2d_x/namelist.input: mp_physics                          = 14,
em_squall2d_x/namelist.input: mp_physics                          = 2,
em_squall2d_y/namelist.input: mp_physics                          = 2,
em_tropical_cyclone/namelist.input: mp_physics                          = 1,
```

```
em_convrad/namelist.input: ra_lw_physics                       = 4,
em_scm_xy/namelist.input: ra_lw_physics                       = 1,
em_seabreeze2d_x/namelist.input: ra_lw_physics                       = 1,
```

I can see not wanting to drop a ton of files in a directory. Our error messages when look-up tables are missing are terse: 
```
module_ra_rrtmg_lw: error opening RRTMG_LW_DATA on unit 15
module_sf_noahlsm.F: soil_veg_gen_parm: failure opening VEGPARM.TBL
```
Some of the Fortran OPEN statements do not have an error return, which would be a message to the user with a failure to open unit such and such.

dudhia:
Probably those microphysics options don't need tables?
So seabreeze, scm, and convrad would be left.

kkeene44:
@davegill @dudhia @weiwangncar 
For this last commit (the current proposed code), I specifically tested each case. I removed all linked files and ran ideal and wrf, adding in files as they were requested via fatal errors. Without the files I added, the case would not run. With this test, only convrad, seabreeze, and tropical cyclone needed any linked files and those are the only ones I left in the Makefile.

kkeene44:
Oops, SCM needs it too. Fixing that now.

Makefile
```diff
@@ -330,7 +330,7 @@ em_b_wave : wrf
 	( cd main ; $(MAKE) RLFLAGS="$(RLFLAGS)" MODULE_DIRS="$(ALL_MODULES)" SOLVER=em IDEAL_CASE=ideal em_ideal )
 	( cd test/em_b_wave ; /bin/rm -f wrf.exe ; ln -s ../../main/wrf.exe . )
 	( cd test/em_b_wave ; /bin/rm -f ideal.exe ; ln -s ../../main/ideal.exe . )
-	( cd test/em_b_wave ; /bin/rm -f README.namelist ; ln -s ../../run/README.namelist . )
+	( cd test/em_b_wave ; /bin/rm -f README.namelist ; ln -s ../../run/README.namelist . )
```
weiwangncar, line 333: @kkeene44 Removing the last space will probably remove the difference here.

davegill:
Folks,
Let's put in the P3 mods, THEN this PR.

davegill:
Replaced by #1415 "Remove requirement to run 'run_me_first' scripts for ideal cases"
Every run script was a conflict, so regenerated PR with latest develop branch.

PR #901, opened 2019-05-08
Fix vertical refinement, broken from v4.0 through v4.1

davegill:
TYPE: bug fix

KEYWORDS: hybrid vertical coordinate, vertical refinement

SOURCE: Katie Lundquist (LLNL) and internal

DESCRIPTION OF CHANGES:
Problem:
Since the formal introduction of the hybrid vertical coordinate option in
WRF (with version v4.0), the vertical refinement option (also referred to
as vertical nesting) has been broken for both of the options when vert_refine_method
= 1 or 2 (ndown or WRF, respectively). The symptomology is that model 
dies in the first time step, whether in the dynamics, physics, or initialization.

Solution:
Katie discovered that the 1d arrays (c1, c2, c3, c4 which are used to
construct the pressure level of the vertical levels) were never set for
vertical refinement.

Editor's note: These values were set for traditional nests. Since the number of
vertical levels in the parent and child domains were the same, the already 
existing code was sufficient:
```
 nest%c1h(1:parent%e_vert) = parent%c1h(1:parent%e_vert)
 nest%c2h(1:parent%e_vert) = parent%c2h(1:parent%e_vert)
 nest%c3h(1:parent%e_vert) = parent%c3h(1:parent%e_vert)
 nest%c4h(1:parent%e_vert) = parent%c4h(1:parent%e_vert)
 nest%c1f(1:parent%e_vert) = parent%c1f(1:parent%e_vert)
 nest%c2f(1:parent%e_vert) = parent%c2f(1:parent%e_vert)
 nest%c3f(1:parent%e_vert) = parent%c3f(1:parent%e_vert)
 nest%c4f(1:parent%e_vert) = parent%c4f(1:parent%e_vert)
```
However, when the two domains have a different number of vertical levels these assignments 
are incorrect. When there is no input for the second domain, there is no opportunity inside of
the WRF model to correct these wrong assignments by overwriting the nest values of the 1d 
arrays with a subsequent _read_.

The source code that is required to compute the vertical 1d arrays already exists.
For the real program, the computation of the 1d arrays is handled in the 
`dyn_em/module_initialize_real.F` file. The entirety of those computations are now removed from 
the `module_initialize_real.F` subroutine and placed in `dyn_em/nest_init_utils.F`, as a new 
subroutine. This refactoring does two things:
1. The computations necessary for the vertical refinement are available and consistent with the
existing computations done elsewhere in the real program.
2. A single shared routine is in place instead of duplicated software.

[...]

PR #1435, opened 2021-03-16
Fix to enable vertical nesting with hybrid vertical coordinate

rsarthur:
TYPE: bug fix

KEYWORDS: vertical nesting, hybrid vertical coordinate, vert_refine_method, hybrid_opt

SOURCE: Robert Arthur and Katie Lundquist (LLNL)

DESCRIPTION OF CHANGES:
Problem:
When vertical nesting is used with the hybrid vertical coordinate option (hybrid_opt=2), large errors occur in the 
vertically-refined domain.

Solution:
An incorrect calculation of pb (base pressure) was fixed in subroutine force_domain_em_part2 (module_dm.F). The pb 
calculation was changed to match similar code in start_em. This fix is within a block of code that is only called if 
vert_refine_method .ne. 0, in order to rebalance on the "intermediate" nested grid. The error was likely a result of the 
switch to the new hybrid coordinate definition, which was not updated correctly in this particular subroutine.

The check-a-mundo preventing vertical nesting from being used with the hybrid coordinate (see #1154) has also 
been removed.

This PR is related to #901, which addressed vertical nesting with the hybrid coordinate (but did not fully fix the issue), 
as well as #1408, which addressed vertical nesting with use_theta_m=1. 

LIST OF MODIFIED FILES:
external/RSL_LITE/module_dm.F
share/module_check_a_mundo.F

TESTS CONDUCTED: 
1. The tests conducted here are similar to those in #1408. A 4-domain nested case was completed with vertical 
refinement on d04. All domains were initialized from a wrfinput file and eta levels were not provided. Pertinent 
namelist options are:
```
input_from_file = .true., .true., .true., .true.,
e_vert = 45, 45, 45, 55,
vert_refine_method = 0, 0, 0, 2,
rebalance = 1,
hybrid_opt = 0 or 2, depending on the case
use_theta_m = 0
```

Results are summarized in the plot below, which shows U at the first grid point above the surface on d04 at various 
output times.

U

With the original code, vertical nesting works with hybrid_opt=0, but not hybrid_opt=2. With the updated code, results 
look reasonable for hybrid_opt=2 and are roughly the same as for hybrid_opt=0. Similar patterns are also seen for other 
variables, such as V and MU, although they are not shown here.

2. Several similar tests were also conducted, as above but with
   * d04 initialized via interpolation from d03 during runtime, rather than from wrfinput_d04
   * eta levels specified for each domain in the namelist
   * use_theta_m=1

These tests provided qualitatively similar results, showing that the fix works for a range of possible setups.

3. Finally, the tests above were completed with hybrid_opt=0, but using the updated code. These showed bit-for-bit 
identical results (using diffwrf) to the same cases using the original code, confirming that previous vertical nesting 
functionality with hybrid_opt=0 is maintained. This was done for both use_theta_m options, 0 and 1.

4. Jenkins testing is OK

RELEASE NOTE: Vertical nesting now works with the hybrid vertical coordinate.

### DIFF
--- a/external/RSL_LITE/module_dm.F
+++ b/external/RSL_LITE/module_dm.F
@@ -4257,7 +4257,7 @@ endif
             DO i = MAX(ids,ips),MIN(ide-1,ipe)
 
                DO k = 1, kpe-1
-                  grid%pb(i,k,j) = ngrid%c3h(k)*(ngrid%c1h(k)*grid%mub(i,j)+ngrid%c2h(k)) + ngrid%c4h(k) + ngrid%p_top
+                  grid%pb(i,k,j) = ngrid%c3h(k) * grid%mub(i,j) + ngrid%c4h(k) + ngrid%p_top
              
                   !  If this is a real run, recalc t_init.
    
--- a/share/module_check_a_mundo.F
+++ b/share/module_check_a_mundo.F
@@ -2418,26 +2418,6 @@
 
 !-----------------------------------------------------------------------
 !  Consistency checks for vertical refinement:
-!  must use the terrain following vertical coordinate
-!-----------------------------------------------------------------------
-      oops = 0 
-      DO i = 2, model_config_rec % max_dom
-        IF ( .NOT. model_config_rec % grid_allowed(i) ) CYCLE
-        IF (model_config_rec%vert_refine_method(i) .NE. 0) THEN
-          IF ( model_config_rec%hybrid_opt .NE. 0 ) THEN
-            oops = oops + 1
-          END IF
-        END IF
-      END DO
-
-      IF ( oops .GT. 0 ) THEN
-        wrf_err_message = '--- ERROR: vert_refine_method=2 only works with hybrid_opt = 0 '
-        CALL wrf_debug ( 0, TRIM( wrf_err_message ) )
-        count_fatal_error = count_fatal_error + 1
-      END IF
-
-!-----------------------------------------------------------------------
-!  Consistency checks for vertical refinement:
 !  feedback has to be turned off
 !-----------------------------------------------------------------------
       oops = 0 


davegill:
jenkins
```
Please find result of the WRF regression test cases in the attachment. This build is for Commit ID: 84cefc3bd7d54801219a8ec960ddfb0127bc1ff5, requested by: rsarthur for PR: https://github.com/wrf-model/WRF/pull/1435. For any query please send e-mail to David Gill.

    Failed Simulations are: 
    None
    None
```

external/RSL_LITE/module_dm.F
```diff
@@ -4257,7 +4257,7 @@ SUBROUTINE force_domain_em_part2 ( grid, ngrid, pgrid, config_flags    &
             DO i = MAX(ids,ips),MIN(ide-1,ipe)
 
                DO k = 1, kpe-1
-                  grid%pb(i,k,j) = ngrid%c3h(k)*(ngrid%c1h(k)*grid%mub(i,j)+ngrid%c2h(k)) + ngrid%c4h(k) + ngrid%p_top
+                  grid%pb(i,k,j) = ngrid%c3h(k) * grid%mub(i,j) + ngrid%c4h(k) + ngrid%p_top
```
davegill, line 4260: @rsarthur 
Instead of your fix being this:
```
grid%pb(i,k,j) = ngrid%c3h(k) * grid%mub(i,j) + ngrid%c4h(k) + ngrid%p_top
```
how about:
```
grid%pb(i,k,j) = grid%c3h(k) * grid%mub(i,j) + grid%c4h(k) + grid%p_top
```

Usually, the `ngrid` refers to the nest and the `grid` refers to either the parent or nest, depending on context. If we have `grid` on the LHS, perhaps all of the other parts of the struct should be from `grid`?

rsarthur: @davegill 
I see your point, and have sort of been wondering myself why `ngrid` is used here. It is my impression that in this portion of the code, `ngrid` refers to the nest and `grid` refers to the "intermediate" grid, which has the horizontal resolution of the parent and the vertical levels of the nest. Is that correct? If so, then it seems like we could make the change you suggest. 

However, `ngrid` variables are used similarly throughout this subroutine, so if we change this one line, we might want to make the change throughout. Either way, I'm happy to make the change(s) and rerun my test cases.

Note also that in previous versions of the code, prior to the addition of the hybrid coordinate, `ngrid` was used in a similar way on the RHS in this subroutine.

@kalundquist, do you have any thoughts on this?

kalundquist:
Here, grid refers to the intermediate grid which has the vertical resolution of the nest and the horizontal resolution of the parent.  I doubt that variables such as c3h were ever initialized for the intermediate grid.  But, the point of this subroutine is to be filling the intermediate grid (the grid data structure) with values from the parent and/or nest.

From: rsarthur ***@***.***>
Reply-To: wrf-model/WRF ***@***.***>
Date: Friday, March 19, 2021 at 2:33 PM
To: wrf-model/WRF ***@***.***>
Cc: "Lundquist, Katie" ***@***.***>, Mention ***@***.***>
Subject: Re: [wrf-model/WRF] Fix to enable vertical nesting with hybrid vertical coordinate (#1435)


@rsarthur commented on this pull request.

________________________________

In external/RSL_LITE/module_dm.F<https://urldefense.us/v3/__https:/github.com/wrf-model/WRF/pull/1435*discussion_r597985798__;Iw!!G2kpM7uM-TzIFchu!jbRmxG_hguCyNBWLzrG6jKJqJlUzhWyykW15i8IdM0RNZkgLiYMJIkXm3HKwlRFVH_el$>:

> @@ -4257,7 +4257,7 @@ SUBROUTINE force_domain_em_part2 ( grid, ngrid, pgrid, config_flags    &

             DO i = MAX(ids,ips),MIN(ide-1,ipe)



                DO k = 1, kpe-1

-                  grid%pb(i,k,j) = ngrid%c3h(k)*(ngrid%c1h(k)*grid%mub(i,j)+ngrid%c2h(k)) + ngrid%c4h(k) + ngrid%p_top

+                  grid%pb(i,k,j) = ngrid%c3h(k) * grid%mub(i,j) + ngrid%c4h(k) + ngrid%p_top

@davegill<https://urldefense.us/v3/__https:/github.com/davegill__;!!G2kpM7uM-TzIFchu!jbRmxG_hguCyNBWLzrG6jKJqJlUzhWyykW15i8IdM0RNZkgLiYMJIkXm3HKwlZVBHmaj$>
I see your point, and have sort of been wondering myself why ngrid is used here. It is my impression that in this portion of the code, ngrid refers to the nest and grid refers to the "intermediate" grid, which has the horizontal resolution of the parent and the vertical levels of the nest. Is that correct? If so, then it seems like we could make the change you suggest.

However, ngrid variables are used similarly throughout this subroutine, so if we change this one line, we might want to make the change throughout. Either way, I'm happy to make the change(s) and rerun my test cases.

Note also that in previous versions of the code, prior to the addition of the hybrid coordinate, ngrid was used in a similar way on the RHS in this subroutine.

@kalundquist<https://urldefense.us/v3/__https:/github.com/kalundquist__;!!G2kpM7uM-TzIFchu!jbRmxG_hguCyNBWLzrG6jKJqJlUzhWyykW15i8IdM0RNZkgLiYMJIkXm3HKwlQfhUTfZ$>, do you have any thoughts on this?

external/RSL_LITE/module_dm.F
davegill: @kalunquist @rsarthur 
Thanks for digging in and solving these issues. It is much appreciated.

davegill (approved):

external/RSL_LITE/module_dm.F
rsarthur: You're welcome, @davegill. Thank you for your careful reviews. Hopefully this will allow people to use vertical nesting successfully in v4.

PR #1456, opened 2021-04-04
Noah-MP Dynamic Irrigation Packaged 2d fields

prasanthvkrishna:
TYPE: bug fix

KEYWORDS: Noah MP dynamic irrigation packaging 2d variables.

SOURCE: Prasanth Valayamkunnath (NCAR)

DESCRIPTION OF CHANGES:
Purpose: To package Noah MP dynamic irrigation variables.

LIST OF MODIFIED FILES: 
          Registry/Registry.EM_COMMON
          Registry/Registry.NMM 
          phys/module_sf_noahmplsm.F (resolved issue with ISNAN in PGI)

TESTS CONDUCTED: 
1. Compiled the model using intel/19.0.5. No further testing is conducted as these are minor changes in the code.
2. Jenkins testing OK


dudhia:
Typically packaging needs a test to make sure not using this option is unaffected. If the arrays are still accessed in any way when the option is off, it could cause problems with unallocated space. Make sure these arrays are not in any executed code lines when the option is off.

phys/module_sf_noahmplsm.F
```diff
@@ -8027,8 +8027,8 @@ SUBROUTINE SPRINKLER_IRRIGATION(parameters,NSOIL,DT,SH2O,SMC,SICE,& !in
     ELSE ! Equation (4)
        IRRLOSS   = 4.337*(EXP(0.077*WINDSPEED))*(((ESAT1-EAIR)*0.01)**(-0.098)) ! [%]
     END IF
-    IF(ISNAN(IRRLOSS))IRRLOSS=4.0 ! In case if IRRLOSS is NaN
-    
+    !IF(ISNAN(IRRLOSS))IRRLOSS=4.0 ! In case if IRRLOSS is NaN
```
davegill, line 8030: @prasanthvkrishna 
It is possible that you should keep the existing "if(isnan(irrlos))" test. The standard does not guarantee what happens if you try to do a arithmetic test on a NaN value. I think BOTH tests should be kept.

prasanthvkrishna: Hi Dave, 

We are in the process to make Noah MP consistent across different models. @Cenlin_He has reported that he is getting HRLDAS compile errors while using the PGI compiler. This was the reason I removed the ISNAN check. Do the WRF compile test considers PGI? Please let me know.
Thanks

prasanthvkrishna:
> Typically packaging needs a test to make sure not using this option is unaffected. If the arrays are still accessed in any way when the option is off, it could cause problems with unallocated space. Make sure these arrays are not in any executed code lines when the option is off.

Sure, I will do a test. Thanks, Jimy

phys/module_sf_noahmplsm.F
cenlinhe: Hi @davegill 
ISNAN is not supported by PGI compiler (at least during my test in cheyenne). In order to avoid limiting the use of Noah-MP source code across different host models and compiler platforms, I think it is better to remove the ISNAN check or use something equivalent but works for different compilers. 
Also, based on the current code, it seems that the IRRLOSS variable is unlikely to have NaN values. 
What do you think?

davegill:
@prasanthvkrishna @cenlinhe 
> ISNAN is not supported by PGI compiler

This intrinsic became part of the standard with Fortran2003. We should be able to rely on isnan being present (not a slam on you, but on the PGI compiler that you are using).

I will defer to your decision.

phys/module_sf_noahmplsm.F
davegill: I am OK with this decision

prasanthvkrishna:
> @prasanthvkrishna @cenlinhe
> 
> > ISNAN is not supported by PGI compiler
> 
> This intrinsic became part of the standard with Fortran2003. We should be able to rely on isnan being present (not a slam on you, but on the PGI compiler that you are using).
> 
> I will defer to your decision.
@davegill 
So you suggest using ISNAN? Should I make it active then?

jamiebresch:
> @prasanthvkrishna @cenlinhe
> 
> > ISNAN is not supported by PGI compiler
> 
> This intrinsic became part of the standard with Fortran2003. We should be able to rely on isnan being present (not a slam on you, but on the PGI compiler that you are using).
> 
> I will defer to your decision.

isnan can be implemented like 
https://github.com/wrf-model/WRF/blob/fb60d61cc44e2a2e8b8311f0b79185724010d510/phys/module_mp_p3.F#L5311-L5315
That function appears not actually used by P3, though.

cenlinhe:
@davegill @prasanthvkrishna Thanks. Not sure why the PGI compiler in Cheyenne does not include this intrinsic (I tested it using the latest PGI compiler pgi/20.4). How about this: let's remove this ISNAN in noahmplsm here for this WRF release, if we find any potential NaN issue that might occur to this variable in the future applications, we will revisit it through bug fix.

prasanthvkrishna:
> > @prasanthvkrishna @cenlinhe
> > > ISNAN is not supported by PGI compiler
> > 
> > 
> > This intrinsic became part of the standard with Fortran2003. We should be able to rely on isnan being present (not a slam on you, but on the PGI compiler that you are using).
> > I will defer to your decision.
> 
> isnan can be implemented like
> https://github.com/wrf-model/WRF/blob/fb60d61cc44e2a2e8b8311f0b79185724010d510/phys/module_mp_p3.F#L5311-L5315
> 
> 
> That function appears not actually used by P3, though.

@jamiebresch Thanks :)

cenlinhe:
> > > @prasanthvkrishna @cenlinhe
> > > > ISNAN is not supported by PGI compiler
> > > 
> > > 
> > > This intrinsic became part of the standard with Fortran2003. We should be able to rely on isnan being present (not a slam on you, but on the PGI compiler that you are using).
> > > I will defer to your decision.
> > 
> > 
> > isnan can be implemented like
> > https://github.com/wrf-model/WRF/blob/fb60d61cc44e2a2e8b8311f0b79185724010d510/phys/module_mp_p3.F#L5311-L5315
> > 
> > That function appears not actually used by P3, though.
> 
> @jamiebresch Thanks :)

This would also be an option.

cenlinhe:
@prasanthvkrishna @davegill OK, it seems that we could implement the isnan function to resolve this issue. Prasanth, could you please implement this function into module_sf_noahmplsm.F and do a quick test if it works?

prasanthvkrishna:
Sure, I will do that.

cenlinhe:
@davegill  So if we implement the logical function as suggested above following module_mp_p3.F, we will use another function name "isnan_lsm" to avoid conflicts with "isnan" that is already an intrinsic function name for INTEL currently. Do you think this works for you?

davegill:
@prasanthvkrishna @dudhia @weiwangncar 
Prasanth,
Take a look at #1460
What if we removed the ISNAN part of this PR, and put the ISNAN change into my PR?

phys/module_sf_noahmplsm.F
```diff
@@ -8040,6 +8043,13 @@ SUBROUTINE SPRINKLER_IRRIGATION(parameters,NSOIL,DT,SH2O,SMC,SICE,& !in
     IREVPLOS = IRSIRATE*IRRLOSS*(1./100.) 
     IRSIRATE = IRSIRATE-IREVPLOS
   END SUBROUTINE SPRINKLER_IRRIGATION
+
+    logical function isnan_lsm(arg1)
+        real,intent(in) :: arg1
+       isnan_lsm = (arg1 .ne. arg1)
+       return
+    end function isnan_lsm
+
```
davegill, line 8052: This "roll your own" ISNAN() function is the technique used in other physics packages (P3 MP, TEMF BL). I am entirely OK with this solution.

davegill:
> @davegill So if we implement the logical function as suggested above following module_mp_p3.F, we will use another function name "isnan_lsm" to avoid conflicts with "isnan" that is already an intrinsic function name for INTEL currently. Do you think this works for you?

Yes, this is just fine. It is done elsewhere in the model.

davegill (approved):
1. Packaging is a good idea
2. ISNAN() functionality is implemented as is already done in P3 and TEMF

phys/module_sf_noahmplsm.F
cenlinhe: Hi @davegill, OK thanks. We will go with our solution here by defining the isnan_lsm function as done in other physics packages.

prasanthvkrishna:
> 1. Packaging is a good idea
> 2. ISNAN() functionality is implemented as is already done in P3 and TEMF

@davegill Thanks, Dave.

davegill:
@prasanthvkrishna @cenlinhe @dudhia
> Make sure these arrays are not in any executed code lines when the option is off.

Folks,
Jimy made the request above. Has this been verified?

prasanthvkrishna:
> @prasanthvkrishna @cenlinhe @dudhia
> 
> > Make sure these arrays are not in any executed code lines when the option is off.
> 
> Folks,
> Jimy made the request above. Has this been verified?

@dudhia @davegill Yes Dave. I tested it by opting the unified Noah instead of Noah-MP. None of the dynamic irrigation 2d fields are in the WRF output.  So, the packaging worked.

dudhia:
Another test should make sure NoahMP results are not changed when the
irrigation scheme is off before and after packaging.

davegill:
@prasanthvkrishna 
> Another test should make sure NoahMP results are not changed when the
irrigation scheme is off before and after packaging.

What is the status of Jimy's requested test?

prasanthvkrishna:
> @prasanthvkrishna
> 
> > Another test should make sure NoahMP results are not changed when the
> > irrigation scheme is off before and after packaging.
> 
> What is the status of Jimy's requested test?

@dudhia @davegill 
I checked the results. No difference between versions. Thanks.

PR #1457, opened 2021-04-05
Fix for fractional seaice using MYJ and QNSE surface layer and MYNN PBL

weiwangncar:
TYPE: bug fix

KEYWORDS: fractional seaice, MYJ and QNSE surface layer, MYNN PBL

SOURCE: internal, reported by Lou Wicker (NSSL)

DESCRIPTION OF CHANGES:
Problem:
When MYNN PBL is used with MYJ surface layer with the fractional seaice option on, an input field wspd is not defined. 

Solution:
Define wspd just the same as when the fractional seaice option is off. Hence, the loop to define wspd is moved outside the existing `if (fractional_seaice == 1)` block. The same change is applied to QNSE surface layer option too.

M    phys/module_surface_driver.F

TESTS CONDUCTED: 
1. This allows MYJ surface layer option to work with MYNN PBL when fractional seaice is on.
2. Jenkins tests all pass.

RELEASE NOTE: When MYNN PBL is used with MYJ surface layer with the fractional seaice option on the model failed immediately, due to an input field wspd is not being defined. This has been fixed.

### DIFF
--- a/phys/module_surface_driver.F
+++ b/phys/module_surface_driver.F
@@ -2313,6 +2313,7 @@ CONTAINS
               ims,ime, jms,jme, kms,kme,                           &
               i_start(ij),i_end(ij), j_start(ij),j_end(ij),     &
               kts,kte,scm_force_flux    )
+           ENDIF
 #if ( EM_CORE==1)
          DO j = j_start(ij),j_end(ij)
             DO i = i_start(ij),i_end(ij)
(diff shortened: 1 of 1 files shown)


davegill:
@weiwangncar @dudhia 
This looks good to me.

davegill (approved):

PR #1458, opened 2021-04-05
turn ysu_topdown_pblmix on by default

weiwangncar:
TYPE: enhancement

KEYWORDS: top-down mixing, YSU PBL

SOURCE: internal

DESCRIPTION OF CHANGES:
The `ysu_topdown_pblmix` option was introduced in WRF 3.7, with the default set to 0. The option mixes the cooler 
and moister air down into the PBL from cloud-topped boundary layer driven by radiation. This option has been shown 
to produce better results in a variety of cases. Hence, we would like to turn it on by default. 

LIST OF MODIFIED FILES: 
M       Registry/Registry.NMM

TESTS CONDUCTED: 
1. Jenkins tests OK

RELEASE NOTE: The namelist option ysu_topdown_pblmix for the YSU PBL scheme has been turned on by default in the ARW and NMM main Registry files. To turn if off, set ysu_topdown_pblmix=0 in the &physics namelist record (single entry).


davegill:
@weiwangncar 
>  The namelist option ysu_topdown_pblmix for the YSU PBL scheme has been turned on by default in the ARW and NMM main Registry files. 

Wei,
Would you add one more sentence concerning what this option actually does, super brief.

weiwangncar:
@davegill Added.

davegill (approved):
Approved one-character change

PR #1462, opened 2021-04-07
Added two missing END PARALLEL DO directives for do_avgflx_em=1

matzegoebel:
TYPE: bug fix

KEYWORDS: OMP, do_avgflx_em

DESCRIPTION OF CHANGES:
The missing OMP directive END PARALLEL DO was added at two places: where the time-averaging of fluxes is 
initialized and updated (relevant for do_avgflx_em=1).

LIST OF MODIFIED FILES: 
dyn_em/solve_em.F

TESTS CONDUCTED:
1. Jenkins is OK, and a large number of jenkins tests are with OpenMP.

RELEASE NOTE: Added missing OMP directives in ARW solve for the option: do_avgflx_em=1. Many OpenMP implementations assume the "END PARALLEL DO", but it is more conventional within WRF to have it.

### DIFF
--- a/dyn_em/solve_em.F
+++ b/dyn_em/solve_em.F
@@ -3458,6 +3458,7 @@ BENCH_END(bc_end_tim)
                  &   grid%avgflx_efu1,grid%avgflx_dfd1,grid%avgflx_efd1 )
             CALL wrf_debug(200,'In solve_em, after zero_avgflx call')
          ENDDO
+         !$OMP END PARALLEL DO
       ENDIF
 
 ! Update avgflx quantities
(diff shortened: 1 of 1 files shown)


matzegoebel:
I noticed that this OMP directive is usually placed after the tile loops, but not at these two places. But still, it compiles and seems to work right?  So is the directive redundant in general?

davegill (approved):
Approved

davegill:
@matzegoebel 
> So is the directive redundant in general?

Not redundant, but optional. There are some optimizations that are easier for the compilers to organize when the explicit blocks are user-defined. In WRF, we put the "END PARALLEL DO" at the end of the DO loop for all of the "PARALLEL DO" blocks (well, _almost_ all of the time). :)

PR #1463, opened 2021-04-07
Removed redundant code in vertical_diffusion_implicit

matzegoebel:
TYPE: no impact

DESCRIPTION OF CHANGES:
Removed redundant code in vertical_diffusion_implicit

The case isfflx=1 in the routine vertical_diffusion_implicit contained do loops with no impact. In these loops, only 
unindexed variables that are not used within the do loop are set. There are no other assignments in the do loops, so 
the loops can be safely removed.

LIST OF MODIFIED FILES:  
dyn_em/module_diffusion_em.F

TESTS CONDUCTED:
1. Before vs after is identical.

### DIFF
--- a/dyn_em/module_diffusion_em.F
+++ b/dyn_em/module_diffusion_em.F
@@ -8099,13 +8099,7 @@ END SUBROUTINE phy_bc
     ENDDO
  
   CASE (1) ! use surface heat flux computed from surface routine
-    DO j = j_start, j_end
-    DO i = i_start, i_end
-       cpm = cp * (1. + 0.8 * moist(i,kts,j,P_QV))
-       heat_flux = hfx(i,j)/cpm/rho(i,1,j)
-    ENDDO
-    ENDDO
- 
+    ! do nothing
   CASE DEFAULT
     CALL wrf_error_fatal( 'isfflx value invalid for diff_opt=2' )
   END SELECT hflux


dyn_em/module_diffusion_em.F
```diff
@@ -8099,13 +8099,7 @@ SUBROUTINE vertical_diffusion_implicit(ru_tendf, rv_tendf, rw_tendf, rt_tendf,&
...
  
   CASE (1) ! use surface heat flux computed from surface routine
-    DO j = j_start, j_end
-    DO i = i_start, i_end
-       cpm = cp * (1. + 0.8 * moist(i,kts,j,P_QV))
-       heat_flux = hfx(i,j)/cpm/rho(i,1,j)
-    ENDDO
-    ENDDO
- 
+    ! do nothing
```
davegill, line 8102: @dudhia 
Jimy,
We need you to evaluate this change.

davegill:
@matzegoebel 
> The case isfflx=1 in the routine vertical_diffusion_implicit contained do loops with no impact 

Would you add a sentence explaining that those assignments are to local variables that are not indexed, etc.

davegill (approved):
These are local, unindexed variables that are not used within the do loop. They can be removed. There are no other assignments in the do loop, so the loop can be removed.

dyn_em/module_diffusion_em.F
dudhia: correct: neither heat_flux nor cpm are needed below. This will have no effect.

matzegoebel:
ok done

PR #1467, opened 2021-04-13
set fractional_seaice=0 for ideal namelists with default of SLAB scheme

kkeene44:
TYPE: no impact

KEYWORDS: fractional_seaice, ideal, convrad, seabreeze2d_x, tropical_cyclone, SLAB

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem: 
The namelist parameter "fractional_seaice" was recently modified in the Registry.EM_COMMON file to be set to "on" 
(1) by default. However, if using the SLAB scheme, fractional_seaice must be turned off. Three of the idealized cases 
(em_convrad, em_seabreeze2d_x, and em_tropical_cyclone), by default, use the SLAB scheme, but fractional_seaice 
was not included in the namelists, and since the default is "on," an error occurs when running wrf.

Solution:
Modified the namelists for the 3 cases, setting fractional_seaice = 0.

LIST OF MODIFIED FILES:
M  test/em_convrad/namelist.input
M  test/em_seabreeze2d_x/namelist.input
M  test/em_tropical_cyclone/namelist.input

TESTS CONDUCTED:
1. wrf runs without the fractional_seaice error with the mods.
2. Waiting on Jenkins

### DIFF
--- a/test/em_convrad/namelist.input
+++ b/test/em_convrad/namelist.input
@@ -62,7 +62,7 @@
  cudt                                = 0,
  bucket_mm                           =10.,
  bucket_J                            =1.e8,
- fractional_seaice		     = 0,
+ fractional_seaice		                 = 0,
  /
 
  &fdda
--- a/test/em_convrad/namelist.input
+++ b/test/em_convrad/namelist.input
@@ -62,6 +62,7 @@
  cudt                                = 0,
  bucket_mm                           =10.,
  bucket_J                            =1.e8,
+ fractional_seaice		     = 0,
  /
 
  &fdda
--- a/test/em_convrad/namelist.input
+++ b/test/em_convrad/namelist.input
@@ -62,7 +62,7 @@
  cudt                                = 0,
  bucket_mm                           =10.,
  bucket_J                            =1.e8,
- fractional_seaice		      = 0,
+ fractional_seaice		       = 0,
  /
 
  &fdda
--- a/test/em_convrad/namelist.input
+++ b/test/em_convrad/namelist.input
@@ -62,7 +62,7 @@
  cudt                                = 0,
  bucket_mm                           =10.,
  bucket_J                            =1.e8,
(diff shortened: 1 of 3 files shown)


weiwangncar:
@kkeene44 Thanks for catching this!

@kkeene44 @davegill @dudhia I would like to re-propose the change: 1. Turn off fractional_seaice in the Registry - considering turning that on helped to fix a number of problems  associated with that option in combination with others (sigh!); 2. Add fractional_seaice = 1 to namelist.input files in test/em_real/ to alert user the option; 3. leave namelists in idealized cases alone... Let me know what you all think...

kkeene44:
@weiwangncar,
I'm okay with your proposal. I'll wait to see what the others think.

davegill:
@kkeene44 
This is picky - would you line up the entries please? I am OK after that

kkeene44:
@davegill 
I want to understand what I do wrong with my code that causes this every time I submit a PR! It's lined up perfectly in my actual code, yet it always comes out off a space or 2 in github. Do you have a trick, or know how I can stop this from happening? As you can see, it also took me EIGHT tries to get it right!

dudhia:
I am fine with this.

weiwangncar:
@dudhia Are you agreeing with my proposal?

dudhia:
Yes, I think you are saying to change the default back and just change some
relevant
namelists instead.

weiwangncar:
@dudhia Thanks. It could mean adding fractional_seaice option to several namelists in test/em_real/ directory.
@kkeene44 Would you like to make the new PR?

kkeene44:
@weiwangncar 
Yes, and I will close this one.

weiwangncar:
@kkeene44 Thanks! I would suggest that we add fractional_seaice = 1 to these namelist.input* files in test/em_real/ directory:
namelist.input
namelist.input.jan00
namelist.input.global

PR #1422, opened 2021-02-28
Fix for fractional_seaice option, again

weiwangncar:
TYPE: bug fix

KEYWORDS: fractional seaice, MM5 and revised MM5 surface layer schemes, uninitialized local arrays

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
Intermittent model failure when fractional_seaice is turned with nesting, with errors like 'invalid memory reference'.

Solution:
Two local arrays, qsfc_sea (which causes the problem), and znt_sea are not initialized, and somehow the problem only shows up in nesting. This should fix one of the comparison test, 60NE, but may not fix test 07NE, due to use of urban physics option 2.

LIST OF MODIFIED FILES: 
M     phys/module_surface_driver.F

TESTS CONDUCTED: 
1. Do mods fix problem? Partially. It should fix model failure, and nesting bit-for-bit test with seaice flag on. 
2. Jenkins tests are passing.

RELEASE NOTE: Part of fix in order to turn on fractional_seaice flag by default.

PR #1401, opened 2021-02-11
Change default fractional_seaice option from 0 to 1

weiwangncar:
TYPE: enhancement

KEYWORDS: namelist fractional_seaice, default, replacing original PR#1128

SOURCE: Internal

DESCRIPTION OF CHANGES:
Since GFS, the most commonly used input data source, has been providing fractional seaice
data for a while, it seems reasonable to make the default option for fractional_seaice to be 1.
This should also help our general users. Both ERA5 and ERA-Interim provide fractional seaice
data, too. 

Turning on the fractional seaice flag broke many MPI vs OpenMP comparison tests. This led to the finding that the 
option is not thread safe for a number surface layer physics options: revised MM5, original MM5, MYJ, QNSE and PX (sf_sfclay_physics = 1, 91, 2, 4, and 7). This PR additionally provides the fix for these options.

LIST OF MODIFIED FILES:
M Registry/Registry.EM_COMMON
M phys/module_surface_driver.F

TESTS CONDUCTED:
1. Regression test all pass.

RELEASE NOTE: The default option for using fractional_seaice has been changed from 0 (off) to 1 (on) starting with release v4.3. In addition, several surface layer schemes are made thread safe if fractional seaice option is used. These are revised MM5, original MM5, MYJ, QNSE and PX surface layer scheme (sf_sfclay_physics = 1, 91, 2, 4, and 7).

PR #1468, opened 2021-04-14
fractional_seaice turned off in registry, turned on in namelists for real cases

kkeene44:
TYPE: no impact

KEYWORDS: fractional_seaice, registry, Registry.EM_COMMON, namelist.input, SLAB

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
The namelist parameter "fractional_seaice" was recently modified in the Registry.EM_COMMON file to be set to "on"
(1) by default (commit 0fff69a1b117424d, PR #1401 "Change default fractional_seaice option from 0 to 1". However, if 
using the SLAB scheme, fractional_seaice must be turned off, which causes problems for some of the default 
idealized cases. Additionally, some other cases where users are running with the SLAB scheme (and not realizing it's 
turned on by default), would cause lots of user difficulty and confusion.

Solution:
Set fractional_seaice back to 0 (off) in Registry.EM_COMMON, but add it (and turned on) to three of the default 
namelists in the test/em_real directory as examples.

LIST OF MODIFIED FILES:
M test/em_real/namelist.input
M test/em_real/namelist.input.global
M test/em_real/namelist.input.jan00
M Registry/Registry.EM_COMMON

TESTS CONDUCTED:
1. The namelist changes have no impact on the build or run phases of the model.
2. When fractional seaice was activated by default, this uncovered several problems that were previously addressed. 
Reverting the default fractional seaice setting back to zero (OFF) is not problematic.
   * f87b449e6c04,  #1422 "Fix for fractional_seaice option, again"
   * 476990af1aa3, #1457 "Fix for fractional seaice using MYJ and QNSE surface layer with MYNN PBL"

### DIFF
--- a/Registry/Registry.EM_COMMON
+++ b/Registry/Registry.EM_COMMON
@@ -2578,7 +2578,7 @@ rconfig   integer slope_rad               namelist,physics      max_domains    0
 rconfig   integer topo_shading            namelist,physics      max_domains    0       -     "topo_shading" "1: apply topographic shading to radiation, 0:not" ""
 rconfig   integer topo_wind               namelist,physics      max_domains    0       -     "topo_wind"  "2: Use Mass sfc drag scheme, 1: improve effects topography over surface wind, 0:not" ""
 rconfig   integer no_mp_heating           namelist,physics      1              0       -     "no_mp_heating" "switch to turn of latent heating in mp schemes"   ""
-rconfig   integer fractional_seaice       namelist,physics      1              1       -     "fractional_seaice" "Fractional sea-ice option"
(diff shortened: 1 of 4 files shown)


weiwangncar (approved):

davegill:
@kkeene44 @weiwangncar 
Folks,
I am good with this decision to make fractional seaice OFF by default, and to show some examples in the namelists. We need to be careful when we put the RELEASE NOTES together, as there is a section about the default value for fractional seaice changing.

davegill (approved):
Approved

kkeene44:
@davegill 
Noted. I'll try to look out for that in the release notes and NOT include that.

PR #1469, opened 2021-04-14
NML check: fractional seaice + SLAB not permitted

davegill:
TYPE: enhancement

KEYWORDS: namelist, fractional seaice, slab

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
When a user specifies both the SLAB thermal diffusion land model and the fractional seaice option, the model fails.

Solution:
Put in a simple check to find this namelist inconsistency, and report this to the user.

LIST OF MODIFIED FILES:

TESTS CONDUCTED: 

RELEASE NOTE: When a user specifies both the SLAB thermal diffusion land model and the fractional seaice option, the model fails. A simple check to find this namelist inconsistency has been added.


davegill:
@kkeene44 
Mostly this is to see if I can trigger a jenkins test. I also could not find any record of a jenkins test associated with your new PR #1468

weiwangncar:
@davegill Shall we remove these lines from module_surface_driver.F around lines 2599?
           IF ( FRACTIONAL_SEAICE == 1 ) THEN
              CALL wrf_error_fatal('SLAB scheme cannot be used with fractional seaice')
           ENDIF

davegill:
@weiwangncar 
> Shall we remove these lines from module_surface_driver.F around lines 2599?

Wei,
Good one.
I'll put that in the PR.

davegill:
RE-DO THIS
I based this from an old master branch. Oops

PR #1471, opened 2021-04-15
Switched off optimization for KPP compilation, to stop buffer overflow.

douglowe:
TYPE: bug fix

KEYWORDS: KPP, GCC

SOURCE: Douglas Lowe (University of Manchester), Will Hatheway

DESCRIPTION OF CHANGES:
Problem:
Compiling of WRF-Chem with KPP chemistry schemes fails on
ubuntu systems, due to a buffer overflow error for the few largest
chemical schemes. This does not happen on HPC systems, so might
be a ubuntu specific issue (however we were unable to get to the root
cause of what compiler flag differences there might be).

Solution:
We switched off all optimization of the compilers for KPP only (using -O0).
This stopped the buffer overflow error. We did try minor optimization (-O1),
but this caused the error again, so we've left all optimization off. As this only
impacts the speed of the KPP program during compilation of WRF-Chem
(and marginally, at that), we felt this was an acceptable solution, rather than
trying to find the exact optimization flag which caused the problem. 

LIST OF MODIFIED FILES: 
chem/KPP/configure_kpp

TESTS CONDUCTED: 
1. Compilation on ubuntu now works.
2. Timing test - entire model with KPP:

| Compiler Optimization  | Compile Time|
| :----------------------: |:-------------:| 
|            -O                         | 55 min           |
|          -O0                         | 54 min           |  
|         -O3                          | 53 min           | 

3. Timing test - KPP mechanism isolated: To get a better idea of the KPP processing time I isolated the KPP chemical 
mechanism building step, and ran the compiler with each of these flags 20 times. I don't think there's any systematic 
differences in build time between each of the optimization options - probably the KPP run time is short enough for it 
not to be noticeable. The results with standard deviations:

| Compiler Optimization  | Compile Time|
| :----------------------: |:-------------:| 
|            -O                         | 55.6± 2 s       |
|          -O0                         | 58.4 ± 6.7 s   |  
|         -O3                          | 57.6 ± 1.9 s    | 

4. Jenkins tests are all PASS.

RELEASE NOTE: KPP is now compiled unoptimized. This will help with compiling WRF-Chem on ubuntu (and other smaller memory) systems. Timing tests indicate that this does not cause a significant timing impact on the WRF build time (less than 5%).

### DIFF
--- a/chem/KPP/configure_kpp
+++ b/chem/KPP/configure_kpp
@@ -61,11 +61,11 @@ if test -z "$FLEX_LIB_DIR" ; then
 fi
 
 
-#    Platform independent C compiler flags. By default "-O" is used which 
-#    turns on optimisation. If you are experiencing problems you may try 
+#    Platform independent C compiler flags. By default "-O0" is used which 
+#    turns off all optimisation. If you are experiencing problems you may try 
 #    "-g" to include debuging informations.
 
-CC_FLAGS="-O"
+CC_FLAGS="-O0"
 
 
 


davegill:
@douglowe 
Doug,
Do you have access to an HPC system? What I'd like to see is the performance difference. Would you be able to measure a before vs after to give us a sense of the impact that this would have. I understand that this is just for building the code. However, a KPP build already takes significantly longer than a traditional Chem build.

davegill:
@douglowe 
Doug,
Also, we are unable to understand this word that has an "s" instead of a "z": optimisation. :)

douglowe:
> @douglowe
> Doug,
> Do you have access to an HPC system? What I'd like to see is the performance difference. Would you be able to measure a before vs after to give us a sense of the impact that this would have. I understand that this is just for building the code. However, a KPP build already takes significantly longer than a traditional Chem build.

Yes - the KPP build does take a lot longer. I think a lot of that is due to building the kpp_interface module, rather than running KPP itself. But I'll run a couple of test builds on our local HPC, and will report back on the difference this makes to build time.

douglowe:
> @douglowe
> Doug,
> Also, we are unable to understand this word that has an "s" instead of a "z": optimisation. :)

Sure thing - I'll switch to US spellings.

douglowe:
I've checked the compile times on our local HPC. With the standard `-O` flag compilation took ~55 minutes. With the `-O0` flag it was ~54 minutes, and using a `-O3` flag it took ~53 minutes.

To get a better idea of the KPP processing time I isolated the KPP chemical mechanism building step, and ran the compiler with each of these flags 20 times. The results were (with standard deviations):
`-O`: 55.6 +/- 2 seconds
`-O3`: 57.6 +/- 1.9 seconds
`-O0`: 58.4 +/- 6.7 seconds (think I had some conflict for resources for one of the runs).

I don't think there's any systematic differences in build time between each of the optimization options - probably the KPP run time is short enough for it not to be noticable.

davegill:
@jordanschnell 
I have reviewed this change. I am OK with this proposed mod. It does not impact anyone except for KPP users. For KPP users, the impact seems relatively small - a few percent for build time. There is no impact on the KPP execution.

jordanschnell:
@davegill @douglowe Thanks to both of you. Will Hatheway (who should receive some credit here) contacted me about this problem in a build script he was writing for the community - figured it was memory related but thankfully Doug was able to figure it out.

jordanschnell (approved):
Approved by /chem

ruikangma:
Hello,

I was wondering if it is safe to change the optimization level from "CC_FLAGS="-O"" to "CC_FLAGS="-O0""? I am currently using a highly modified WRF-Chem model based on version 4.1.

I have observed that when I compile WRF using gnu/9.1.0 with "CC_FLAGS="-O0"", the results differ from those obtained when compiling with intel/19.1.1 using "CC_FLAGS="-O"". I would appreciate any insights you can provide on this matter.

Thank you,
Ruikang

PR #1479, opened 2021-04-21
Fix hybrid vertical coordinate 1d array assignment

davegill:
TYPE: bug fix

KEYWORDS: hybrid vertical coordinate, HVC, 1d

SOURCE: Bill Skamarock (NCAR), internal

DESCRIPTION OF CHANGES:
Problem:
The hybrid vertical coordinate has a default option (hyb_opt=2). For the not so frequently used option where the 
stretching uses a sin^2() function (hyb_opt=3), the bounding value for the weighting term at the top of the 
atmosphere should be identically zero. It is likely computed to be nearly zero, but an IF test cleanly assigns the 
value. However, that IF test uses the wrong logic for which k-index to check.

Solution:
All we do is swap `kds` with `kde` in the IF test:
```
         ELSE IF ( k .EQ. kds ) THEN
```
with
```
         ELSE IF ( k .EQ. kde ) THEN
```

LIST OF MODIFIED FILES: 
modified:   nest_init_utils.F

TESTS CONDUCTED: 
1. This option is not advertised, so the incorrect bounding value was not consequential. However, the modification is 
    correct based on surrounding code, and on the actual derivation of the purpose of the term.
   * The previous IF test for hybrid option 2 uses the correct logic, and the corrected IF test for the sin^2() function is 
      now identical to the logic used in hybrid option 2.
   * The term in question, c3f, is the full level weighting term. At the top of the atmosphere, this needs to be identically 
      zero, so that the pressure is entirely isobaric.
2. Jenkins is all pass.

RELEASE NOTE: A bug fix was found to the hybrid vertical coordinate. Those using the default with the hybrid option activated (hyb_opt=2), or those reverting back to a terrain following coordinate (hyb_opt=0) are not impacted. Basically, you are unlikely to be impacted.

### DIFF
--- a/dyn_em/nest_init_utils.F
+++ b/dyn_em/nest_init_utils.F
@@ -1090,7 +1090,7 @@ SUBROUTINE compute_vcoord_1d_coeffs ( ht, etac, znw, &
          c3f(k) = znw(k)*sin(0.5*3.14159*znw(k))**2
          IF      ( k .EQ. kds ) THEN
             c3f(k) = 1.
-         ELSE IF ( k .EQ. kds ) THEN
+         ELSE IF ( k .EQ. kde ) THEN
             c3f(kde) = 0.
          END IF
       ELSE


davegill:
@skamaroc @dudhia @weiwangncar 

Folks,
Bill found a bug in WRF!
The error is only activated with hybrid_opt=3, which uses a sin^2() stretching function for the weighting term (1=terrain following, 0=isobaric). This hyb_opt=3 option is not used in WRF, so we do not expect any impact for users. The snippet below shows the corrected code within context. We need to have IF tests for k=KDS and k=KDS as we want the value for C3F to be identically ONE at the surface and identically ZERO at the model lid, not computationally close to those values.

```
   DO k=kts, kte
      IF      ( hybrid_opt .EQ. 0 ) THEN
         c3f(k) = znw(k)
      ELSE IF ( hybrid_opt .EQ. 1 ) THEN
         c3f(k) = znw(k)
      ELSE IF ( hybrid_opt .EQ. 2 ) THEN
         B1 = 2. * etac**2 * ( 1. - etac )
         B2 = -etac * ( 4. - 3. * etac - etac**3 )
         B3 = 2. * ( 1. - etac**3 )
         B4 = - ( 1. - etac**2 )
         B5 = (1.-etac)**4
         c3f(k) = ( B1 + B2*znw(k) + B3*znw(k)**2 + B4*znw(k)**3 ) / B5
         IF ( znw(k) .LT. etac ) THEN
            c3f(k) = 0.
         END IF
         IF ( k .EQ. kds ) THEN
            c3f(k) = 1.
         ELSE IF ( k .EQ. kde ) THEN
            c3f(k) = 0.
         END IF
      ELSE IF ( hybrid_opt .EQ. 3 ) THEN
         c3f(k) = znw(k)*sin(0.5*3.14159*znw(k))**2
         IF      ( k .EQ. kds ) THEN
            c3f(k) = 1.
         ELSE IF ( k .EQ. kde ) THEN
            c3f(kde) = 0.
         END IF
      ELSE
         CALL wrf_message     ( 'ERROR: --- hybrid_opt' )
         CALL wrf_message     ( 'ERROR: --- hybrid_opt=0    ==> Standard WRF terrain-following coordinate' )
         CALL wrf_message     ( 'ERROR: --- hybrid_opt=1    ==> Standard WRF terrain-following coordinate, hybrid c1, c2, c3, c4' )
         CALL wrf_message     ( 'ERROR: --- hybrid_opt=2    ==> Hybrid, Klemp polynomial' )
         CALL wrf_message     ( 'ERROR: --- hybrid_opt=3    ==> Hybrid, sin^2' )
         CALL wrf_error_fatal ( 'ERROR: --- Invalid option' )
      END IF
   END DO
```

davegill:
jenkins
```
Please find result of the WRF regression test cases in the attachment. This build is for Commit ID: 5c5b94d05eb91fb82c7c95de83e3dc7e90398861, requested by: davegill for PR: https://github.com/wrf-model/WRF/pull/1479. For any query please send e-mail to David Gill.

    Failed Simulations are: 
    None
    None
```

weiwangncar:
I'm ok with the fix.

davegill:
@weiwangncar 
Wei,
If you are OK with this, would you approve it please? We are on pretty solid footing with this one.

weiwangncar (approved):

PR #1486, opened 2021-04-23
Bug fix FARMS to use correct cloud transmittance

pedro-jm:
TYPE: bug fix

KEYWORDS: FARMS, undefined variable, cloud transmittance

SOURCE: Pedro A. Jimenez and Jimy Dudhia (NCAR), and Jaemo Yang and Yu Xie (NREL)

DESCRIPTION OF CHANGES:
Problem: Incorrect treatment of transmittance of diffuse radiation by clouds in FARMS

Solution:
Change the cloud transmittance and initialize variables before using them

LIST OF MODIFIED FILES: 

TESTS CONDUCTED: 
1. The impact of these mods should be small.
There is no impact on the results in the test we run. We used Intel compiler which I think initializes values to 0 if the variables are not defined. And 0 happens to be the right value in this case.

2. Jenkins tests are all passing.

RELEASE NOTE: Corrected the incorrect treatment of cloud transmittances of diffuse radiation by clouds in the FARMS system. The impact of these mods are small.

### DIFF
--- a/phys/module_ra_farms.F
+++ b/phys/module_ra_farms.F
@@ -347,7 +347,7 @@
       if (tau_qv > TAU_MIN) then
         Call Watermodel (tau_tot, de_cloud2, solarangle, Tducld_water, Ruucld_water)
       else
-        Tducld_water = 1.0
+        Tducld_water = 0.0
         Ruucld_water = 0.0
         tau_qv2 = 0.0
       end if
@@ -356,7 +356,7 @@
       if (tau_qi > TAU_MIN) then
         call Icemodel (tau_tot, de_ice2, solarangle, Tducld_ice, Ruucld_ice)
       else
-        Tducld_ice = 1.0
+        Tducld_ice = 0.0
         Ruucld_ice = 0.0
         tau_qi2 = 0.0
       end if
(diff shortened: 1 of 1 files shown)


dudhia:
state something about effect on results if any.

dudhia:
Confirmation of test result needs to be added for approval. Also note that this was mainly to pass a configure -D test as it had no effect on results.

pedro-jm:
@dudhia
There is no impact on the results in the test we run. We used Intel compiler which I think initializes values to 0 if the variables are not defined. And 0 happens to be the right value in this case.
Pedro.

pedro-jm:
Would it be possible to incorporate this fix in v4.3?

dudhia:
Yes, it is already in that queue.

pedro-jm:
wonderful.

dudhia:
@pedro-jm can you update the test result message before I approve this?

weiwangncar:
I'm ok with this PR.

pedro-jm:
@dudhia 
I just updated the result test message.

dudhia (approved):

dudhia:
OK, I approved it.

PR #1417, opened 2021-02-25
Adding the MAD-WRF model

davegill:
TYPE: New feature

KEYWORDS: MAD-WRF, MADCast, WRF-Solar, cloud initialization, cloud advection/diffusion, nudging

SOURCE: Pedro Jimenez, Jared Lee, and Jimy Dudhia (NCAR), and Greg Thompson (JCSDA)

DESCRIPTION OF CHANGES:
Problem: There is room for improvement in the short range shortwave irradiance predictions

Solution:
MAD-WRF is a solar irradiance nowcasting system that combines:
1. the benefits of data assimilation and the philosophy of advecting cloud properties of MADCast, and
2. the cloud physics of WRF-Solar. 

The initial cloud field can be enhanced based on relative humidity and the model can impose the cloud mask and the cloud top / base height from available earth observations. The hydrometeor content is advected and diffused as tracers, and the resolved hydrometeors are nudged towards these tracers at the beginning of the simulation to facilitate the cloud spin up. This results in improved solar irradiance prediction in the short-range (the first hours). There is an option to run MAD-WRF with just advecting and diffusing the hydrometeors.

We extensively tested MAD-WRF performance using an implementation in an older version of WRF. Our results show the improved performance at the beginning of the simulation. 

The following technical description follows the README.madwrf file in the doc directory. There are two options to run MAD-WRF:

1. madwrf_opt = 1: The initial hydrometeors are advected and diffused with the model dynamics without accounting for any microphysical processes. 
   * The user should set mp_physics = 96 and use_mp_re = 0 in the physics block of the WRF namelist.

2. madwrf_opt = 2: There is a set of hydrometeor tracers that are advected and diffused with the model dynamics. 
   * At initial time the tracers are equal to the standard hydrometeors.
   * During the simulation the standard hydrometeors are nudged towards the tracers.
   * The namelist variable madwrf_dt_nudge sets the temporal period for hydrometeor nudging [min]
   * The namelist madwrf_dt_relax sets the relaxation time for hydrometeor nudging [s]

MAD-WRF has an option to enhance the cloud initialization. To turn on (off) the cloud initialization set the namelist variable madwrf_cldinit to 1 (0)

[...]

PR #1488, opened 2021-04-24
Fix restart problem with 3 or more domains when o3input = 2

weiwangncar:
TYPE: bug fix

KEYWORDS: restart, at least 3 domains with feedback on, ozone input option 2

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
With o3input = 2 as a default option for RRTMG, and feedback on, the restart capability is not bit for bit when using three or more total domains. This is due to different XLAT and XLONG are used in the initial interpolation of ozone data to the model grid at model start time versus model restart time.

Solution:
Removing 'u' (the Registry "feedback" option) for XLAT and XLONG fixes the problem. While we were making these nesting changes in the Registry, the 'd' and 'u' options for cloud fraction are also removed. There is no need to do either horizontal interpolation or feedback for this variable.

LIST OF MODIFIED FILES:

TESTS CONDUCTED: 
1. This PR fixes the none bit-for-bit restart problem with nests with at least 3 domains. 1. The restart now also works at non lateral boundary times - restart times at 2h 45 m and 1 h were tested and worked with 3-h lateral boundary condition interval. The code is also tested with no input files from domain 2 and 3.
2. All tests were conducted with feedback on, but smooth_option = 0.
3. The Jenkins tests are all passing.

RELEASE NOTE: Fixed a restart problem when using 3 or more total domains with feedback on and using RRTGM with o3input set to 2 (default). Restart now works at all times, though with smooth_option off.

### DIFF
--- a/Registry/Registry.EM_COMMON
+++ b/Registry/Registry.EM_COMMON
@@ -49,8 +49,8 @@
 # table entries are of the form
 #<Table> <Type> <Sym>         <Dims>   <Use>   <NumTLev> <Stagger> <IO>     <DNAME>             <DESCRIP>     <UNITS>
 #
-state    real  XLAT             ij      misc        1         -     i0123rh0156{22}{23}du=(copy_fcnm)      "XLAT"                "LATITUDE, SOUTH IS NEGATIVE"                                          "degree_north"
-state    real  XLONG            ij      misc        1         -     i0123rh0156{22}{23}d=(interp_fcn_blint_ll:xlat,input_from_file)u=(copy_fcnm)      "XLONG"               "LONGITUDE, WEST IS NEGATIVE"                                          "degree_east"
+state    real  XLAT             ij      misc        1         -     i0123rh0156{22}{23}d      "XLAT"                "LATITUDE, SOUTH IS NEGATIVE"                                          "degree_north"
+state    real  XLONG            ij      misc        1         -     i0123rh0156{22}{23}d=(interp_fcn_blint_ll:xlat,input_from_file)      "XLONG"               "LONGITUDE, WEST IS NEGATIVE"                                          "degree_east"
 
 # It is required that LU_INDEX appears before any variable that is
 # interpolated with a mask, as lu_index supplies that mask.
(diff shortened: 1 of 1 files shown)


weiwangncar:
Passed all Jenkins tests:

    Failed Simulations are: 
    None
    None

davegill:
jenkins
```
Please find result of the WRF regression test cases in the attachment. This build is for Commit ID: 2cf91659ea2f01593393e8cca3de977501dcc0c3, requested by: weiwangncar for PR: https://github.com/wrf-model/WRF/pull/1488. For any query please send e-mail to David Gill.

    Failed Simulations are: 
    None
    None
```

davegill (approved):

dudhia:
Interesting that cldfra is also not fed back now. Any reason? Comment needs to be added in PR. CLDFRA should be fed back only if radiative tendencies are, but they are not so I agree with this change too.

weiwangncar:
@dudhia CLDFRA had just 'rh' until MAD WRF PR (#1417). 
@pedro-jm Do you know if making CLDFRA 'rhdu' a must for MAD WRF?

pedro-jm:
@weiwangncar 
I never remember the meaning of those letters. For MAD-WRF we would like to have the cloud fraction in the output from Real.

dudhia:
Yes, u just means feedback from the nest. Not needed here.

weiwangncar:
@dudhia The letter 'd' (interpolating parent domain field to the nest) for CLDFRA may only provide 'cloud fraction' at the time when nest starts and radiation is NOT called, but it doesn't do anything to the model results. The next time radiation is called, CLDFRA will be computed on the nest. For this reason, and for the reason that 'd' and 'u' were not associated with CLDFRA prior to MAD WRF code, should we keep 'd' with this field (CLDFRA)?

dudhia:
I agree that d is not needed for CLDFRA either and can be removed.

weiwangncar:
@dudhia Thanks. Can you approve this PR?

dudhia:
It is already approved but I also approve of the d removal for CLDFRA.

PR #1499, opened 2021-04-28
Fix nest feedback smoothing with restarts

davegill:
TYPE: bug fix

KEYWORDS: restart, smooth, MPI

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
When an MPI job uses restarts for a nested run that had smoothing activated for the feedback of the nest, the results were not bit-for-bit.

Solution:
1. A number of fields associated with GWD were incorrectly labelled with the `s` (feedback smoothing) flag in Registry. Those were unnecessary, and have been removed. 
2. The remaining fields that do have the `s` flag in the Registry have been added to a HALO for feedback.

LIST OF MODIFIED FILES: 
modified:   share/interp_fcn.F

TESTS CONDUCTED: 
1. Without the mods, the restart tests for both smooth_opt = 1 , 2 are not bit for bit. With the mods, smooth_opt = 0, 1, 2, 3 are all bit for bit with a restart (and using MPI). 

2. Jenkins tests all passing.

RELEASE NOTE: Registry clean-up to reduce the fields that are subject to feedback smoothing in the parent domain, and to include all fields with feedback smoothing in the FEEDBACK HALO. These fixes gives bit-wise identical results with a distributed memory restart when doing smoothing on the feedback area on the coarse grid (in the area of the nest).


dudhia:
Do we really apply the smoother to all those GWD arrays? Perhaps we don't need to.

davegill:
@dudhia 
> Do we really apply the smoother to all those GWD arrays? 

Jimy, yep.

dudhia:
This is a questionable need since they are likely low-res fields anyway (?),
and GWD is more likely to only be used on an outer domain.

davegill:
@kkeene44 
Kelly,
This works for smooth option 0, 1, 2, with restarts, with MPI.

davegill:
jenkins
```
Please find result of the WRF regression test cases in the attachment. This build is for Commit ID: e3548820395f54d7467604f78ea7f520cc48e76f, requested by: davegill for PR: https://github.com/wrf-model/WRF/pull/1499. For any query please send e-mail to David Gill.

    Failed Simulations are: 
    None
    None
```

kkeene44:
@davegill 
I've tested this and you are correct - with these mods smooth_opt = 0, 1, 2 are all identical with restarts.

davegill:
@weiwangncar @dudhia 
Wei, 
I included your plots in the PR.
Looking at the figures, I think that removing the U and V fields is too big of an impact. I think that we hold this PR until after the release.

weiwangncar:
@davegill I'd be ok without this fix in 4.3 - we've had this problem forever, it seems.

dudhia:
I think it may be OK to have the fix with no u and v smoothing even for this release.
Just to make restarts work.

davegill:
@weiwangncar @dudhia 
The latest changes reflects suggestions from Jimy (the "u" nest options from the GWD fields), and re-include U/V in feedback.

weiwangncar (approved):

davegill:
closing this, replaced with #1500 and #1501 - in that order

PR #1501, opened 2021-05-01
Remove ARW nesting "us" (feedback and smoothing) from all GWD variables

davegill:
TYPE: bug fix

KEYWORDS: feedback, smoothing, GWD

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
The GWD fields should not have information from the fine grid fed back up to the parent. 

Solution:
Remove the feedback option `u` from the GWD state definitions in the Registry file. Since we are getting rid of feedback, then we must get rid of the smoothing option `s` that is available only for feedback.

LIST OF MODIFIED FILES:

TESTS CONDUCTED: 
1. The purpose of this change is physical meaningfulness - the coarse grid GWD fields should not be getting info from a child domain. The code builds and runs, that is sufficient testing.
2. Jenkins tests are all PASS.

RELEASE NOTE: Remove feedback from all of the GWD fields in the ARW Registry.


dudhia (approved):
This should be committed before the restart feedback smoothed fields fix #1502 because that fix also requires these changes.

davegill:
@dudhia @weiwangncar 
> This should be committed before the restart feedback smoothed fields fix #1502 because that fix also requires these changes.

Jimy,
Completely right. We are waiting for some jenkins issues to be resolved.

davegill:
jenkins
```
Please find result of the WRF regression test cases in the attachment. This build is for Commit ID: 2af28d55fff61741d4f7db110002284d2632d146, requested by: davegill for PR: https://github.com/wrf-model/WRF/pull/1501. For any query please send e-mail to David Gill.

    Failed Simulations are: 
    None
    None
```

davegill:
Folks,
Jimy has already approved these changes. I simply did some code alignment to force an additional jenkins test to verify that the tests were working. The jenkins tests are working.  However, those mods invalidate Jimy's previous "approval", so we need another one.

weiwangncar:
Please go ahead. I'm ok with this. I just wonder how we can quantify the differences seen due to this change.

weiwangncar (approved):

PR #1534, opened 2021-07-14
Updated P3 code with bug fixes and cleanup

hmorrison100:
TYPE: bug fixes along with general P3 update (update from P3v4.0.7 to P3 v4.2.2)

KEYWORDS: P3, microphysics

SOURCE: Hugh Morrison (NCAR)

DESCRIPTION OF CHANGES:
Problem: 
Bug in solve_em.F so that 3-moment P3 (MP option 53) was not working properly. Some additional minor bug fixes to the P3 module and lookup tables as well as code clean up (affecting MP options 50, 51, 52, and 53). Also updated the two lookup table generation codes (which do not affect WRF, but are included for users as a reference). This also required changing the head Makefile to update the 3-moment lookup table file name. Rather than back porting the minor bug fixes to the older version of P3 module (v4.0.7, in the release WRFv4.3.1), we have just swapped to a newer version of P3 (v4.2.2). Thus, it includes some other minor changes and clean up as well.

Solution:
Uncommented 1 line in solve_em.F in the call to microphysics driver. Made several changes to the P3 module and lookup tables to address minor bugs. Required modifying head Makefile. See above.

Fixes #123

LIST OF MODIFIED FILES: 
Modified: solve_em.F, module_mp_p3.F, Makefile
Added: p3_lookupTable_1.dat-5.3-3momI.gz, p3_lookupTable_2.dat-2momI_v5.2.2, p3_lookupTable_1.dat-5.3-2momI, create_p3_lookupTable_1.f90-v5.3, create_p3_lookupTable_2.f90-v5.2.2
Deleted: create_p3_lookupTable_1.f90-v5.1.6_b01, p3_lookupTable_1.dat-2momI_v5.1.6_oldDimax, p3_lookupTable_1.dat-3momI_v5.1.6.gz, p3_lookupTable_2.dat-4.1, create_p3_lookupTable_2.f90-v4.1

TESTS CONDUCTED: 
1. Idealized tests confirm the code is working. Additional real-case tests using the GEM model and same P3 module + lookup tables confirm reasonable results.

RELEASE NOTE: Include a stand-alone message suitable for the inclusion in the minor and annual releases. A publication citation is appropriate.


hmorrison100:
@dudhia @weiwangncar Let's try this again -- I fixed the issue with the Makefile and added the updated 3-moment lookup table file. It should compile/run fine now.

weiwangncar:
@hmorrison100 Can you say anything about the kind of differences one expects to see with this PR?

hmorrison100:
@weiwangncar no problem -- I modified the comment above to add much more detail, including expected impacts.

hmorrison100:
@weiwangncar @dudhia Hi, just checking in about this PR. It looks still active? Is there anything else needed on my end to finish it? Thanks.

weiwangncar:
@hmorrison100 We should start working on 4.3.1 release, and go thorough the open PRs in the next few weeks. Thanks for keeping an eye on it.

hmorrison100:
@weiwangncar Great! thanks Wei

davegill:
@hmorrison100 @weiwangncar @dudhia 
Folks,
Are we targeting this for 4.3.1? This has quite a few new tables. Opinions?

weiwangncar:
Yes, we are targeting 4.3.1.

weiwangncar:
@smileMchen Could you run a test to see if this PR works?

smileMchen:
@weiwangncar 
Yes , I will do soon.

hmorrison100:

Hi, is it still possible to slightly modify module_mp_p3.F? We just found a bug a few days ago -- the only change is that 4 lines need to commented out (this bug was actually introduced recently, it's not in the release WRFV4.3 code). Sorry for bringing this up last minute. Extensive tests show the impact is generally small, however the bug leads to ice-rain collection being far too small so this will introduce error at the process level. If it's still possible I'd really like to get the fix into WRFV4.3.1. The updated module_mp_p3.F is ready right now. What's the best approach? Should I copy this PR to a new one and update module_mp_p3.F, and then close this PR?

@davegill yes the lookup table files were all updated. However, the changes compared to previous table versions (in WRFV4.3) are generally small (and a few cosmetic things like removing some unused columns of data). The only exception is the lookup table 2 used by MP option 52, where process rates for ice-ice collection were in incorrect before by up to a factor of 2-3.

weiwangncar:
@hmorrison100 Please go ahead and make the change.

hmorrison100:
@weiwangncar do I need to start a new PR?

weiwangncar:
@hmorrison100 If the code still exists in your local computer, you should be able to edit the same branch and push the change to this branch on Github. In fact, since changes in this PR have not been merged, you probably cannot make any changes to the code proposed in this PR. You can also edit this PR directly from Github by clicking 'Files changed' on top of this paper, then use the '...' to the right of the each changed file to find 'edit' button. After you modified the code, you will be asked to enter a commit message similar to what you do on a local computer. If none of the these work for you, let me know your change, and I can help editing it on Github.

hmorrison100:
@weiwangncar thanks for the detailed instructions, I replaced with the updated module_mp_p3.F. All should be good now.

weiwangncar:
@hmorrison100 That's great! It looks like it passed reg test too.

kkeene44:
@hmorrison100 
Can you include the correct links in the Makefile so that the p3_lookupTable* files are correctly linked to the appropriate location during the build?

hmorrison100:
@kkeene44 Hi Kelly, thanks for noticing that. I replaced the old Makefile with a new one with the correct table filenames for all the links.

kkeene44:
Thanks @hmorrison100! I just ran a quick test and confirmed the tables link correctly now. (@weiwangncar)

weiwangncar (approved):

dudhia:
Need to check name spelling in Release Note

dudhia:
Was this tested versus the results we got for V4.3? The 3-moment scheme was an outlier and I wonder if it still is.

weiwangncar:
@hmorrison100 Can you address Jimy's comment?

hmorrison100:
@weiwangncar @dudhia I did not test versus those old results -- I don't have the WRF configuration files for this but it would be great to see it done. We did test extensively for the idealized cases in WRF and for many real cases in GEM (same version of P3).

weiwangncar:
@hmorrison100 Remember when Ming tested option 53 for 4.3, we saw it behaving differently from other P3 options? Are any of the fixes you have in this PR aimed at changing that behavior?

hmorrison100:
@weiwangncar yes this is what I assumed JImy was referring to and I remember Ming's tests. However this used a pretty severely buggy version of option 53 (the line that was incorrectly commented out in solve_em.F so that the 3-moment moment was not running as fully 3-moment code). I'm therefore hesitant to put too much stock into those runs. So the short answer is "no", we didn't address that specifically, and anticipate it will be different with that major bug fixed. That said, in some idealized offline 1D column and real case simulations in GEM there were still some hints of this behavior, but not to the extreme that Ming's WRF tests were showing. Again, it would be great to see what this does in those tests now that option 53 is configured properly.

smileMchen:
I am just back from PTO and need to catch up with many stuff. But I will remember this issue and will manage to run some test cases and compare with our previous run. I will keep you updated about the results. 
At the same time, I suppose it should be fine to merge this PR into WRFV4.3.1.

smileMchen (approved):

hmorrison100:
@smileMchen Thanks! we'll keep this in mind as well since Jason and Melissa (the other P3 developers) will soon be doing another big set of real case runs in GEM

hmorrison100:
@smileMchen @dudhia thanks for showing this. Looks reasonable -- in particular, the profile of QCLOUD is much more similar (if somewhat smaller QCLOUD magnitude), which was very different in the buggy code before. There are large difference in QICE between P3 and std43 but this is because P3 includes *all* ice as QICE, so an apples-to-apples comparison with other schemes would have to sum all the ice categories (which I'm assuming you didn't do for std43, judging by the profile of QICE).

PR #1535, opened 2021-07-15
Add note for WRF ARM build

yip111:
TYPE：WRF 4.2 Porting Guide for ARM

KEYWORDS: ARM, aarch64, Porting

DESCRIPTION OF CHANGES:
Problem:
WRF cannot run directly on the ARM machine.

Solution:
Edit the arch/configure.defaults file and add the aarch64 GNU & OPEN MPI compile options

LIST OF MODIFIED FILES:
M arch/configure.defaults
M README_ARM.md


martin-g:
Hello!

I am also interested in support for Linux ARM64!
If the project maintainers would need a Linux ARM64 VM for long-term testing/CI purposes then please let me know!

davegill:
@weiwangncar @dudhia @kkeene44 @smileMchen @yip111 
Folks,
What do we think about this PR?

Specifically, I am wondering if this should be broken into 2 pieces. The ARM specific update to the arch/configure.defaults file would be one PR. The README could be a second PR. We have never really thought about putting build info in the repository before.

Opinions?

weiwangncar:
@yip111 @davegill I'd suggest putting the README file in the doc/ directory.

dudhia:
Is the README needed? I notice it refers to V4.2. Is there anything special
about this install?

jamiebresch:
Seems a useful note, but probably should not be committed to the repository, especially not with the name README_ARM.
Is `-march=armv8.2-a `the only ARM specific setting? `armv8.2-a` will be architecture-dependent. It will be enough to just point out the compiler ARM options.

weiwangncar:
Should have read the README file. I agree as the file stands now, it probably should not be committed to the repository. It looks like general info for installing software before compiling WRF. If it can be made general, it may be useful for general users.

arch/configure.defaults
```diff
@@ -1975,6 +1975,47 @@ LIB_BUNDLED     = \
                      $(WRF_SRC_ROOT_DIR)/frame/pack_utils.o
 
 #insert new stanza here
+################################################## #########
+    #ARCH   Linux   aarch64,gnu OpenMPI #serial smpar dmpar dm+sm
```
martin-g, line 1979: I think the next line should be just `#`.
Also all lines below the stanza should not have any indentation.
Without these corrections I was not able to build it.

davegill:
@yip111 
Have you tried to see if version 4.3 works? I was able to build WRF on a raspberry pi, for example.

martin-g:
@davegill I've just tested this PR with WRF-4.3 instead of 4.2 and all is fine!

yip111:
> Have you tried to see if version 4.3 works? I was able to build WRF on a raspberry pi, for example.

@davegill Yes, it also works on version 4.3.

yip111:
> Specifically, I am wondering if this should be broken into 2 pieces. The ARM specific update to the arch/configure.defaults file would be one PR. The README could be a second PR. We have never really thought about putting build info in the repository before.

@davegill @martin-g Thanks for your suggestion, and I update and submit a sepated PR on https://github.com/wrf-model/WRF/pull/1557 .

davegill:
Replaced by #1557 Add new stanza for ARM aarch64/openmpi

PR #1542, opened 2021-08-11
Bug fix for aerosol-aware Thompson Scheme

smileMchen:
TYPE: bug fix

KEYWORDS:  aerosol aware Thompson scheme

SOURCE: internal

DESCRIPTION OF CHANGES: 
Problem:
When running a nested case with the aerosol aware Thompson scheme, the surface aerosol number concentration 
emissions (QNWFA2D) differ by almost one order of magnitude between the parent and child domains (even when using
a recommended nest ratio such as 3:1). 

Solution:
This PR fixes the problem by removing the grid-size dependence of airmass, but still keeping the layer thickness 
dependence.

LIST OF MODIFIED FILES:
M   phys/module_mp_thompson.F

TESTS CONDUCTED:
1. A nested case with the original code is run to demonstrate the large difference in QNWFA2D between the parent 
and child domains (results shown with left figure).
2. The same nested case is done with the modified code. These results show a more reasonable and consistent 
behavior between the two domains (results shown with right figure).
3. Jenkins test is all pass.

image

RELEASE NOTE: Since V4.0 the aerosol aware Thompson (mp=28) option has not been initializing the surface aerosol source array consistently between a parent and nested domain due to grid-size dependence. This has been removed and the results will change if dx is not 20 km, but now nests will have the same source.

### DIFF
--- a/phys/module_mp_thompson.F
+++ b/phys/module_mp_thompson.F
@@ -419,7 +419,7 @@
 
 
       INTEGER:: i, j, k, l, m, n
-      REAL:: h_01, airmass, niIN3, niCCN3, max_test, z1
+      REAL:: h_01, niIN3, niCCN3, max_test, z1
       LOGICAL:: micro_init, has_CCN, has_IN
 
       is_aerosol_aware = .FALSE.
@@ -463,7 +463,6 @@
             endif
             niCCN3 = -1.0*ALOG(naCCN1/naCCN0)/h_01
             nwfa(i,1,j) = naCCN1+naCCN0*exp(-((hgt(i,2,j)-hgt(i,1,j))/1000.)*niCCN3)
-            airmass = 1./orho(i,1,j) * (hgt(i,2,j)-hgt(i,1,j))*dx*dy      ! kg
             z1=hgt(i,2,j)-hgt(i,1,j)
             nwfa2d(i,j) = nwfa(i,1,j) * 0.000196 * (50./z1)
             do k = 2, kte


smileMchen:
Jenkins test passed:
Please find result of the WRF regression test cases in the attachment. This build is for Commit ID: fa0e2498c90f45ddc9c4aef2ab407f2f7c58ed88, requested by: smileMchen for PR: https://github.com/wrf-model/WRF/pull/1542. For any query please send e-mail to David Gill.

    Failed Simulations are: 
    None
    None

dudhia:
Looks good to me. Need @gthompsnWRF to look at it.

gthompsnWRF:
> Looks good to me. Need @gthompsnWRF to look at it.


Yes, I like this approach a lot.  Thank you Jimy!!!

phys/module_mp_thompson.F
```diff
@@ -419,7 +419,7 @@ SUBROUTINE thompson_init(hgt, orho, nwfa2d, nwfa, nifa,           &
 
 
       INTEGER:: i, j, k, l, m, n
-      REAL:: h_01, airmass, niIN3, niCCN3, max_test
+      REAL:: h_01, airmass, niIN3, niCCN3, max_test, z1 !mchen
```
dudhia, line 422: Ming you can take your name off this line

dudhia:
Need to add release note. I believe this is only a problem in V4.3.

dyn_em/module_initialize_real.F
```diff
@@ -4333,11 +4333,7 @@ SUBROUTINE init_domain_rk ( grid &
          do i = its, min(ide-1,ite)
             z1 = (grid%phb(i,2,j)-grid%phb(i,1,j))/g
             airmass = 1./grid%alt(i,1,j) * z1 * config_flags%dx*config_flags%dy          ! kg
```
dudhia, line 4335: airmass no longer needed

phys/module_mp_thompson.F
```diff
@@ -464,7 +464,8 @@ SUBROUTINE thompson_init(hgt, orho, nwfa2d, nwfa, nifa,           &
             niCCN3 = -1.0*ALOG(naCCN1/naCCN0)/h_01
             nwfa(i,1,j) = naCCN1+naCCN0*exp(-((hgt(i,2,j)-hgt(i,1,j))/1000.)*niCCN3)
             airmass = 1./orho(i,1,j) * (hgt(i,2,j)-hgt(i,1,j))*dx*dy      ! kg
```
dudhia, line 466: airmass no longer needed

smileMchen: Jimy,
Many thanks for the comments.  I will modify the codes accordingly.
Ming

phys/module_mp_thompson.F
smileMchen: Jimy,
It has been changed.

dyn_em/module_initialize_real.F
smileMchen: Jimy,
this code has been changed based on your suggestion.

phys/module_mp_thompson.F
```diff
@@ -419,7 +419,7 @@ SUBROUTINE thompson_init(hgt, orho, nwfa2d, nwfa, nifa,           &
 
 
       INTEGER:: i, j, k, l, m, n
-      REAL:: h_01, airmass, niIN3, niCCN3, max_test
+      REAL:: h_01, airmass, niIN3, niCCN3, max_test, z1
```
dudhia, line 422: airmass can also be removed from this list

dudhia, line 422: airmass can also be removed from this code including the later calculation line (466)

davegill:
@smileMchen @dudhia 
Ming,
Would you put together a RELEASE NOTE for this aerosol aware bugfix PR please? It seems like a really important bug fix. Would you also indicate how long this has been in the code. Jimy mentioned maybe just recently.

dudhia:
I will add the RELEASE NOTE

dudhia:
I looked back and this should have been a problem as far back as V4.0.
Maybe no one ran nested thompson aerosol?

smileMchen:
@dudhia 
We are aware of this issue quite a while ago, although I don't remember exactly which version of WRF we use.

smileMchen:
@davegill 
Dave,
This issue has bene there for quite a while. Jimy just said that it started from WRFv4.0. I haven't checked the code yet.

dudhia:
This will affect results unless people are running ~20 km. Ming's parent
domain may have been close to 20 km because the plotted values are a bit
different even though the pattern is the same.

smileMchen:
@dudhia @davegill 
Jimy and Dave,
My test case  is 15-3km nesting.

weiwangncar (approved):

dudhia (approved):

PR #1543, opened 2021-08-18
SFCLAY=1, add shallow water roughness calculation

phawbeck:
TYPE: new feature

KEYWORDS: shallow water, roughness, bathymetry

SOURCE: Patrick Hawbecker, Pedro A. Jiménez, and Jimy Dudhia (NCAR)

DESCRIPTION OF CHANGES:
This incorporates the shallow water drag module into the revised M-O surface layer scheme for over-water roughness calculation.

The new scheme is meant to incorporate the effects of roughness increasing over shallower waters (between 10 
and 100 m deep), mostly near coastlines. In order to include the shallow water roughness parameterization 
from Jiménez and Dudhia (2018) into the revised M-O surface layer scheme, a bathymetry dataset was 
downloaded and converted to the WPS tiled format, new variables and flags for the bathymetry data and 
which scheme is to be used over the ocean are added, and the source code is included into a module with 
module_sf_sfclayrev.F. 

The bathymetry data is open source from GEBCO Compilation Group (2021) GEBCO 2021 Grid ([doi:10.5285/c6612cbe-50b3-0cff-e053-6c86abc09f8f](https://www.gebco.net/data_and_products/gridded_bathymetry_data/)). Several checks are included to 
ensure users without the bathymetry data can still run with this new scheme using a namelist-defined 
variable (shalwater_depth) to designate the bathymetry over all water cells. If users have bathymetry data, 
they can run with this scheme (shalwater_z0 = 1; shalwater_depth <=0) or overwrite the bathymetry data 
with a user-defined value (shalwater_depth > 0).

LIST OF MODIFIED FILES: 
M Registry/Registry.EM_COMMON
M dyn_em/module_first_rk_step_part1.F
M dyn_em/nest_init_utils.F
M dyn_em/start_em.F
M phys/module_physics_init.F
M phys/module_sf_sfclayrev.F
M phys/module_surface_driver.F
M run/README.namelist 
M share/module_optional_input.F
M test/em_real/namelist.examples

TESTS CONDUCTED: 
Tests were conducted in 2 phases: within (1) real.exe and (2) wrf.exe.

1. Real.exe tests conducted:
image

For the case producing a warning (no bathymetry, shalwater_z0 = 1.0, shalwater_depth <= 0.0), the following 
warning is presented in rsl.error.* files:
```
Warning: No bathymetry data found for shallow water roughness model.
Warning: shalwater_depth must be greater than 0.0 for WRF to run.
```
The tests all performed as expected producing the following results for the WATER_DEPTH variable:
image

With incorrect namelist settings (use shallow water roughness drag, sfclay scheme other than revised MO):
```
--- ERROR: The shallow water roughness drag only works with sfclay_physics = 1
             Turn off the shallow water option, or change the surface layer scheme
FATAL CALLED FROM FILE:  <stdin>  LINE:    2513
```

2. The tests performed for wrf.exe are as follows:
image
The depth = 0.0 case produces the following error message:
```
No bathymetry data detected and shalwater_depth not greater than 0.0. Re-run WPS to get bathymetry data 
or set shalwater_depth > 0.0
```

The remaining cases produce WATER_DEPTH fields as follows:
image

After running these cases for 36 hours and discarding the first 12 hours as model-spinup. The averaged field 
(output every 2 hours) for ZNT and wind speed can be seen to show the expected impacts from the new 
scheme. When the true bathymetry is used, ZNT is higher for depths between 10 and 100 m (green lines). 
However, when a constant shalwater_depth is used, the value is set over the whole domain and ZNT is 
increased (in this case) over all water cells. Additionally, when shalwater_depth is set to a valid number, it will 
overwrite the bathymetry data within wrf.exe and produce the same result as if you did not have bathymetry 
data.
image
image

Additionally, in order to remain consistent with the lake model (sf_lake_physics > 0) we allow water_depth to 
be overwritten by lake_depth if sf_lake_physics > 0:
image

3. The code passed bit-for-bit check and restart test.

RELEASE NOTE: The shallow-water roughness scheme from Jiménez and Dudhia (2018) is included for offshore roughness adjustment in water depths between 10 and 100 m. The bathymetry data is open source from GEBCO Compilation Group. Please acknowledge the following in presentations and publications: GEBCO Compilation Group (2021) GEBCO 2021 Grid (doi:10.5285/c6612cbe-50b3-0cff-e053-6c86abc09f8f).
Jiménez, P. A., & Dudhia, J. (2018). On the need to modify the sea surface roughness formulation over shallow waters. Journal of Applied Meteorology and Climatology, 57(5), 1101-1110, DOI: [10.1175/JAMC-D-17-0137.1 ](https://doi.org/10.1175/JAMC-D-17-0137.1)

### DIFF
--- a/Registry/Registry.EM_COMMON
+++ b/Registry/Registry.EM_COMMON
@@ -2569,7 +2569,7 @@ rconfig   integer sf_ocean_physics        namelist,physics      1            0  
 rconfig   integer traj_opt                namelist,physics      1            0         h     "traj_opt"    "activate trajectory calculation  0=no, 1=on"   ""
 rconfig   logical dm_has_traj             namelist,physics      max_domains  .false.   rh    "has_traj"    "activate trajectory calculation per domain"   ""
 rconfig   integer tracercall              namelist,physics      1            0         h     "tracercall"  "activate tracer calculation  0=no, 1=on"   ""
-rconfig   integer shalwater_rough         namelist,physics      max_domains  0         rh    "shalwater_rough"     "shallow water roughness flag"      ""
+rconfig   integer shalwater_z0            namelist,physics      max_domains  0         rh    "shalwater_z0"     "shallow water sea surface roughness flag"      ""
 rconfig   real    shalwater_depth         namelist,physics      1            -1.0      rh    "shalwater_depth"     "water depth for shallow water scheme"      "m"
 rconfig   real    OMDT                    namelist,physics      1            1         h     "OMDT"        "Timestep of ocean model"      "s"
 rconfig   real    oml_hml0                namelist,physics      1            50        h     "oml_hml0"    "oml initial mixed layer depth value"   "m"
--- a/Registry/Registry.EM_COMMON
+++ b/Registry/Registry.EM_COMMON
@@ -883,8 +883,8 @@ state    real   XLAIDYN          ij     misc        1         -     -         "X
 state    real   SSTSK            ij     misc        1         -     rhd=(interp_mask_field:lu_index,iswater)u=(copy_fcnm)   "SSTSK"              "SKIN SEA SURFACE TEMPERATURE" "K"
 state    real   lake_depth            ij     misc        1    -     i012rd=(interp_mask_water_field:lu_index,islake)   "lake_depth"              "lake depth" "m"
 state    real     water_depth      ij     misc        1       -     i0rhd  "water_depth"  "global water depth" "m"
-rconfig  integer  shalwater_rough         namelist,physics      max_domains   0         rh    "shallow water roughness flag"     ""      ""
-rconfig  real     shalwater_depth         namelist,physics      1   -1.0       rh    "water depth for shallow water scheme"     ""      ""
+rconfig  integer  shalwater_rough         namelist,physics      max_domains   0         rh    "shalwater_rough"     "shallow water roughness flag"      ""
+rconfig  real     shalwater_depth         namelist,physics      1   -1.0       rh    "shalwater_depth"     "water depth for shallow water scheme"      "m"
 state    real   DTW              ij     misc        1         -     r   "DTW"              "WARM LAYER TEMP DIFF" "C"
 # Ocean surface currents
 state    real   UOCE            ij     misc        1         -     i0124rd=(interp_mask_water_field:lu_index,iswater)  "UOCE"  "SEA SURFACE ZONAL CURRENTS" "m s-1"
--- a/Registry/Registry.EM_COMMON
+++ b/Registry/Registry.EM_COMMON
@@ -171,6 +171,10 @@ state    real   utrop          ij       dyn_em      1        X     i1  "UTROP"  
 state    real   vmaxw          ij       dyn_em      1        Y     i1  "VMAXW"      "V-component of the max wind speed"  "m s-1"
 state    real   vtrop          ij       dyn_em      1        Y     i1  "VTROP"      "V-component of the tropopause wind" "m s-1"
 state    real   erod           ij.      misc        1        -     i012rd  "EROD" "fraction of erodible surface in each grid cell (0-1)"       "none"
+# PSH - begin
+state    real   bathymetry     ij       dyn_em      1        -     i1  "bathymetry" "Bathymetry and topography" "m"
+state    integer  BATHYMETRY_FLAG   -     misc        1         -     i0        "BATHYMETRY_FLAG"   "Flag for bathymetry in the global attributes for metgrid data"
+# PSH - end  end   
 
 #-----------------------------------------------------------------------------------------------------------------------------------------------------------------
 #                                               
@@ -880,6 +884,11 @@ state    real   XLAIDYN          ij     misc        1         -     -         "X
 # SKIN SST
 state    real   SSTSK            ij     misc        1         -     rhd=(interp_mask_field:lu_index,iswater)u=(copy_fcnm)   "SSTSK"              "SKIN SEA SURFACE TEMPERATURE" "K"
 state    real   lake_depth            ij     misc        1    -     i012rd=(interp_mask_water_field:lu_index,islake)   "lake_depth"              "lake depth" "m"
+# PSH - begin
+state    real     water_depth      ij     misc        1       -     i0rhd  "water_depth"  "global water depth" "m"
+rconfig  integer  shalwater_rough         namelist,physics      max_domains   0         rh    "shallow water roughness flag"     ""      ""
+rconfig  real     shalwater_depth         namelist,physics      1   0.0       rh    "water depth for shallow water scheme"     ""      ""
+# PSH - end  
 state    real   DTW              ij     misc        1         -     r   "DTW"              "WARM LAYER TEMP DIFF" "C"
 # Ocean surface currents
 state    real   UOCE            ij     misc        1         -     i0124rd=(interp_mask_water_field:lu_index,iswater)  "UOCE"  "SEA SURFACE ZONAL CURRENTS" "m s-1"
--- a/Registry/Registry.EM_COMMON
+++ b/Registry/Registry.EM_COMMON
@@ -887,7 +887,7 @@ state    real   lake_depth            ij     misc        1    -     i012rd=(inte
 # PSH - begin
 state    real     water_depth      ij     misc        1       -     i0rhd  "water_depth"  "global water depth" "m"
 rconfig  integer  shalwater_rough         namelist,physics      max_domains   0         rh    "shallow water roughness flag"     ""      ""
-rconfig  real     shalwater_depth         namelist,physics      1   0.0       rh    "water depth for shallow water scheme"     ""      ""
+rconfig  real     shalwater_depth         namelist,physics      1   -1.0       rh    "water depth for shallow water scheme"     ""      ""
 # PSH - end  
 state    real   DTW              ij     misc        1         -     r   "DTW"              "WARM LAYER TEMP DIFF" "C"
 # Ocean surface currents
--- a/Registry/Registry.EM_COMMON
+++ b/Registry/Registry.EM_COMMON
@@ -171,10 +171,8 @@ state    real   utrop          ij       dyn_em      1        X     i1  "UTROP"  
 state    real   vmaxw          ij       dyn_em      1        Y     i1  "VMAXW"      "V-component of the max wind speed"  "m s-1"
 state    real   vtrop          ij       dyn_em      1        Y     i1  "VTROP"      "V-component of the tropopause wind" "m s-1"
 state    real   erod           ij.      misc        1        -     i012rd  "EROD" "fraction of erodible surface in each grid cell (0-1)"       "none"
-# PSH - begin
 state    real   bathymetry     ij       dyn_em      1        -     i1  "bathymetry" "Bathymetry and topography" "m"
 state    integer  BATHYMETRY_FLAG   -     misc        1         -     i0        "BATHYMETRY_FLAG"   "Flag for bathymetry in the global attributes for metgrid data"
-# PSH - end  end   
 
 #-----------------------------------------------------------------------------------------------------------------------------------------------------------------
 #                                               
@@ -884,11 +882,9 @@ state    real   XLAIDYN          ij     misc        1         -     -         "X
 # SKIN SST
 state    real   SSTSK            ij     misc        1         -     rhd=(interp_mask_field:lu_index,iswater)u=(copy_fcnm)   "SSTSK"              "SKIN SEA SURFACE TEMPERATURE" "K"
 state    real   lake_depth            ij     misc        1    -     i012rd=(interp_mask_water_field:lu_index,islake)   "lake_depth"              "lake depth" "m"
-# PSH - begin
 state    real     water_depth      ij     misc        1       -     i0rhd  "water_depth"  "global water depth" "m"
 rconfig  integer  shalwater_rough         namelist,physics      max_domains   0         rh    "shallow water roughness flag"     ""      ""
 rconfig  real     shalwater_depth         namelist,physics      1   -1.0       rh    "water depth for shallow water scheme"     ""      ""
-# PSH - end  
 state    real   DTW              ij     misc        1         -     r   "DTW"              "WARM LAYER TEMP DIFF" "C"
 # Ocean surface currents
 state    real   UOCE            ij     misc        1         -     i0124rd=(interp_mask_water_field:lu_index,iswater)  "UOCE"  "SEA SURFACE ZONAL CURRENTS" "m s-1"
--- a/Registry/Registry.EM_COMMON
+++ b/Registry/Registry.EM_COMMON
@@ -172,7 +172,7 @@ state    real   vmaxw          ij       dyn_em      1        Y     i1  "VMAXW"  
 state    real   vtrop          ij       dyn_em      1        Y     i1  "VTROP"      "V-component of the tropopause wind" "m s-1"
 state    real   erod           ij.      misc        1        -     i012rd  "EROD" "fraction of erodible surface in each grid cell (0-1)"       "none"
 state    real   bathymetry     ij       dyn_em      1        -     i1  "bathymetry" "Bathymetry and topography" "m"
-state   integer BATHYMETRY_FLAG -       misc        1        -     i   "BATHYMETRY_FLAG"   "Flag for bathymetry in the global attributes for metgrid data"
+state   integer BATHYMETRY_FLAG -       misc        1        -     i0rh   "BATHYMETRY_FLAG"   "Flag for bathymetry in the global attributes for metgrid data"
 
 #-----------------------------------------------------------------------------------------------------------------------------------------------------------------
 #                                               
--- a/phys/module_physics_init.F
+++ b/phys/module_physics_init.F
@@ -188,6 +188,7 @@ CONTAINS
                          lakemask,   lakeflag,                                   & !lake
 #endif
                          lake_depth_flag, use_lakedepth,                         & !lake
+                         water_depth, bathymetry_flag, shalwater_rough,       & ! PSH -bathymetry
                          sf_surface_mosaic, mosaic_cat, NLCAT,               & ! Noah tiling
 !CLM variables
                          maxpatch,                                           &
@@ -756,6 +757,7 @@ CONTAINS
   LOGICAL, DIMENSION( ims:ime, jms:jme ),intent(out)                      :: lake2d
 !  REAL,    DIMENSION( ims:ime, jms:jme ), INTENT(IN)                      ::  HT
   REAL, OPTIONAL,    DIMENSION( ims:ime, jms:jme ), INTENT(IN)    ::  lake_depth
+  REAL, OPTIONAL,    DIMENSION( ims:ime, jms:jme ), INTENT(IN)    ::  water_depth ! PSH
   real, intent(in)      ::      lakedepth_default, lake_min_elev
 #if ( EM_CORE == 1 )
   REAL,              dimension(ims:ime,jms:jme )      ::  lakemask
@@ -763,6 +765,8 @@ CONTAINS
 #endif
   INTEGER, INTENT(INOUT)      ::   lake_depth_flag
   INTEGER, INTENT(IN)      ::   use_lakedepth
+  INTEGER, INTENT(INOUT)   ::   bathymetry_flag ! PSH
+  INTEGER, INTENT(IN)      ::   shalwater_rough  ! PSH
  
 
 !CLM
@@ -1446,6 +1450,7 @@ CONTAINS
                 lakemask, lakeflag,                                    & !lake
 #endif
                 lake_depth_flag, use_lakedepth,                        & !lake
+                water_depth, bathymetry_flag, shalwater_rough,   & ! PSH
                 te_temf,cf3d_temf,wm_temf,                      & ! WA
                 DZR, DZB, DZG,                                  & !Optional urban
                 TR_URB2D,TB_URB2D,TG_URB2D,TC_URB2D,QC_URB2D,   & !Optional urban
@@ -2470,6 +2475,7 @@ CONTAINS
                 lakemask, lakeflag,                                      & !lake
 #endif
                 lake_depth_flag, use_lakedepth,                          & !lake
+                water_depth,bathymetry_flag, shalwater_rough,    & ! PSH
                 te_temf,cf3d_temf,wm_temf,                      & ! WA
 !                num_roof_layers,num_wall_layers,num_road_layers,& !Optional urban
                 DZR, DZB, DZG,                                  & !Optional urban
@@ -2939,12 +2945,15 @@ CONTAINS
  
   logical,    dimension(ims:ime,jms:jme ),intent(out)                        :: lake2d
   REAL, OPTIONAL,    DIMENSION( ims:ime, jms:jme ), INTENT(IN)    ::  lake_depth
+  REAL, OPTIONAL,    DIMENSION( ims:ime, jms:jme ), INTENT(IN)    ::  water_depth ! PSH
 #if ( EM_CORE == 1 )
   REAL,              dimension(ims:ime,jms:jme ),intent(inout)      ::  lakemask
   INTEGER, INTENT(IN)      ::  lakeflag
 #endif
   INTEGER, INTENT(IN)      ::   use_lakedepth
   INTEGER, INTENT(INOUT)      ::   lake_depth_flag
+  INTEGER, INTENT(IN)      ::   shalwater_rough  ! PSH
+  INTEGER, INTENT(INOUT)   ::   bathymetry_flag ! PSH
 !  REAL,    DIMENSION( ims:ime, jms:jme ), INTENT(IN)                         ::  HT
 
    REAL,  DIMENSION( ims:ime , jms:jme ) , OPTIONAL, INTENT(INOUT) ::    &
--- a/phys/module_physics_init.F
+++ b/phys/module_physics_init.F
@@ -758,7 +758,7 @@ CONTAINS
   LOGICAL, DIMENSION( ims:ime, jms:jme ),intent(out)                      :: lake2d
 !  REAL,    DIMENSION( ims:ime, jms:jme ), INTENT(IN)                      ::  HT
   REAL, OPTIONAL,    DIMENSION( ims:ime, jms:jme ), INTENT(INOUT) ::  lake_depth
-  REAL, OPTIONAL,    DIMENSION( ims:ime, jms:jme ), INTENT(INOUT) ::  water_depth 
+  REAL,              DIMENSION( ims:ime, jms:jme ), INTENT(INOUT) ::  water_depth 
   real, intent(in)      ::      lakedepth_default, lake_min_elev
 #if ( EM_CORE == 1 )
   REAL,              dimension(ims:ime,jms:jme )      ::  lakemask
@@ -2949,7 +2949,7 @@ CONTAINS
  
   logical,    dimension(ims:ime,jms:jme ),intent(out)                        :: lake2d
   REAL, OPTIONAL,    DIMENSION( ims:ime, jms:jme ), INTENT(INOUT) ::  lake_depth
-  REAL, OPTIONAL,    DIMENSION( ims:ime, jms:jme ), INTENT(INOUT) ::  water_depth 
+  REAL,              DIMENSION( ims:ime, jms:jme ), INTENT(INOUT) ::  water_depth 
 #if ( EM_CORE == 1 )
   REAL,              dimension(ims:ime,jms:jme ),intent(inout)      ::  lakemask
   INTEGER, INTENT(IN)      ::  lakeflag
--- a/phys/module_physics_init.F
+++ b/phys/module_physics_init.F
@@ -188,8 +188,8 @@ CONTAINS
                          lakemask,   lakeflag,                                   & !lake
 #endif
                          lake_depth_flag, use_lakedepth,                         & !lake
-                         water_depth, bathymetry_flag, shalwater_rough,       & ! PSH -bathymetry
-                         shalwater_depth,                                    & ! PSH - bathymetry
+                         water_depth, bathymetry_flag, shalwater_rough,       & !bathymetry
+                         shalwater_depth,                                    & ! bathymetry
                          sf_surface_mosaic, mosaic_cat, NLCAT,               & ! Noah tiling
 !CLM variables
                          maxpatch,                                           &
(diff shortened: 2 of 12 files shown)


Registry/Registry.EM_COMMON
```diff
@@ -171,6 +171,10 @@ state    real   utrop          ij       dyn_em      1        X     i1  "UTROP"
 state    real   vmaxw          ij       dyn_em      1        Y     i1  "VMAXW"      "V-component of the max wind speed"  "m s-1"
 state    real   vtrop          ij       dyn_em      1        Y     i1  "VTROP"      "V-component of the tropopause wind" "m s-1"
 state    real   erod           ij.      misc        1        -     i012rd  "EROD" "fraction of erodible surface in each grid cell (0-1)"       "none"
+# PSH - begin
```
dudhia, line 174: we can remove these identifier lines from the Registry maybe use generic 
for shallow water roughness option

weiwangncar:
@phawbeck If you'd like to try to compile NMM, instructions are available from
https://www2.mmm.ucar.edu/wrf/users/supports/support.html. Look for 'Code Preparation and Testing Requirement' in the section for developers. Likely the new variables are either to be added to the NMM registry, passed along the 'pipeline', or declared optional in the code and used for ARW only.

smileMchen:
@phawbeck 
If we run a case with, for example  shalwater_rough = 1 and shalwater_depth = 50, does this indicate that the water depth is 50m everywhere? Thanks.

phawbeck:
> @phawbeck If you'd like to try to compile NMM, instructions are available from
> https://www2.mmm.ucar.edu/wrf/users/supports/support.html. Look for 'Code Preparation and Testing Requirement' in the section for developers. Likely the new variables are either to be added to the NMM registry, passed along the 'pipeline', or declared optional in the code and used for ARW only.

Perfect - I will go ahead and try to compile NMM and see what the issues are. Thanks!

phawbeck:
> @phawbeck
> If we run a case with, for example shalwater_rough = 1 and shalwater_depth = 50, does this indicate that the water depth is 50m everywhere? Thanks.

That is correct - setting shalwater_rough = 1 tells WRF you want to run with this parameterization, then shalwater_depth = [some number greater than 0] will define the water depth over all water cells as this value. If you have shalwater_depth set to a value less than or equal to 0.0 and you have bathymetry data, then it will use the bathymetry data.

smileMchen:
@phawbeck 
Thanks for the clarification.

phawbeck:
Hi - I am just checking to see what the status of this is. I noticed that the checks now say they have passed, but I was not able to compile NMM on my initial attempt but I can give it another go if needed. Thanks a lot!

weiwangncar:
@phawbeck Not sure why it says the checks have passed. If no change has been made to the code since you submitted it a while back, it shouldn't. But we sometimes the automatic tests go haywire...

weiwangncar:
@phawbeck No need to work on making the code to compile NMM. We will drop NMM support in 4.4.

phawbeck:
@weiwangncar great - thanks a lot! So is there anything else I should do about this? Or am I good to just sit and wait :-)

weiwangncar:
@phawbeck Basically, yes. But I do have a question: it seems that you have access to a bathymetry dataset. Can that be shared with the community? What is the resolution of the data?

davegill:
@phawbeck @weiwangncar @dudhia 
1. We probably need mods to share/module_check_a_mundo.F so that only the sfclay=1 tries to use these options. Do we want a stop, or a message that this is not being used? Opinions Wei and Jimy?
2. Since there are new namelist vars, let's put some descriptions in test/em_real/examples.namelist and run/README.namelist.
3. How does the namelist option for depth work? Is it an on/off flag, or do you set a depth?
4. Outside the scope of this PR ... what is the status of getting the bathymetry into the WPS static data tar files?
5. Would you add the DOI to the publication (I moved the reference to the RELEASE NOTES section).

dudhia:
some of this was addressed as README.namelist is updated even though not
listed in the PR.

phawbeck:
> @phawbeck Basically, yes. But I do have a question: it seems that you have access to a bathymetry dataset. Can that be shared with the community? What is the resolution of the data?

Yes - I am hoping to make this available but am not sure the best way to do that. It is 30 second data from GEBCO and it is open sourced data, so I think that means we could share it. I have code to convert the netCDF that you can download to the WPS tiled format, but it is probably easier to simply share the tiles themselves. Any thoughts on how to make this data available? Can it be included as optional download in WPS geog?

weiwangncar:
@phawbeck That's great. I think getting the data into WPS tiles is the best to go. And if you provide the data, we can definitely add it to the optional download. Thanks!

phawbeck:
> @phawbeck @weiwangncar @dudhia
> 
> 1. We probably need mods to share/module_check_a_mundo.F so that only the sfclay=1 tries to use these options. Do we want a stop, or a message that this is not being used? Opinions Wei and Jimy?
> 2. Since there are new namelist vars, let's put some descriptions in test/em_real/examples.namelist and run/README.namelist.
> 3. How does the namelist option for depth work? Is it an on/off flag, or do you set a depth?
> 4. Outside the scope of this PR ... what is the status of getting the bathymetry into the WPS static data tar files?
> 5. Would you add the DOI to the publication (I moved the reference to the RELEASE NOTES section).

Hi Dave - for (2) I have added some lines in test/em_real/examples.namelist. Are the existing changes to run/README.namelist sufficient? 
3. There are 2 flags in the namelist for this. First, shalwater_rough = 1/0 is on/off and can be set on individual domains. The second option, shalwater_depth, is used to define either a constant depth (shalwater_depth > 0) or, if bathymetry data is available and shalwater_depth <= 0, then the real bathymetry data will be used.
4. I am ready to share the bathymetry data, but don't know what the process is for adding that. I have it on Cheyenne and can point you to it if that helps.
5. Where would you like me to add this? In the PR text or somewhere in the code?

Thanks a lot for the help!

davegill:
@phawbeck @kkeene44 
Patrick and Kelly,
Would you two organize how to get the geogrid-tiled data for the bathymetry into the next static-data download. This will be for the v4.4 release (so spring time). But if the data exists, we can get a head start.
Thanks mucho

davegill:
@phawbeck 
You can add the DOI after the reference to the publication in the RELEASE NOTE.

davegill:
@phawbeck 
What is the meaning of "WARNING" for the label in the "real" test?

phawbeck:
> @phawbeck
> You can add the DOI after the reference to the publication in the RELEASE NOTE.

Got it - done!

phawbeck:
> @phawbeck
> What is the meaning of "WARNING" for the label in the "real" test?

This test is simply intended to show that if a user specifies shalwater_rough = 1 but they do not have bathymetry data and shalwater_depth is currently set to be <= 0, then they are warned that they will need to run with a positive value for shalwater_depth when running WRF. It's just to close the loop of all possible configurations.

kkeene44:
@davegill 
@phawbeck shared his bathymetry data with me, along with the necessary GEOGRID.TBL file. I ran a test case using the data and all works fine. I then turned off the shalwater_rough option and ran and see differences in the WATER_DEPTH variable. I've already added the bathymetry data to the WPS_GEOG static fields on Cheyenne, and will package them up to add to the static data web page when the v4.4 code is released.

kkeene44:
@phawbeck 
I would suggest adding the units to the description in the Registry and in README.namelist.

phawbeck:
> @phawbeck
> I would suggest adding the units to the description in the Registry and in README.namelist.

I checked to make sure all new variables had units specified in the Registry and added units for shalwater_depth in the README.namelist. If this isn't what you were referring to, let me know!

kkeene44:
@phawbeck Perfect, thanks!

Registry/Registry.EM_COMMON
```diff
@@ -880,6 +882,7 @@ state    real   XLAIDYN          ij     misc        1         -     -         "X
 # SKIN SST
 state    real   SSTSK            ij     misc        1         -     rhd=(interp_mask_field:lu_index,iswater)u=(copy_fcnm)   "SSTSK"              "SKIN SEA SURFACE TEMPERATURE" "K"
 state    real   lake_depth            ij     misc        1    -     i012rd=(interp_mask_water_field:lu_index,islake)   "lake_depth"              "lake depth" "m"
+state    real   water_depth      ij     misc        1         -     i0rhd  "water_depth"  "global water depth" "m"
```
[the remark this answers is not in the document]
phawbeck: Water depth is set to -2 m if not a water or lake cell. Within the call, itself, if the value is less than 10, then "depth" is set to 10 m. If the cell is a land cell, then shallow water roughness will not be called. So there _shouldn't_ be any issues with a value being negative within the shallow water roughness parameterization.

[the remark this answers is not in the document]
davegill: Thanks!

davegill:
@phawbeck @weiwangncar 
I tweaked the commit message to be consistent with the total files now modified. On my side, the last thing I see is that we need mods to share/module_check_a_mundo.F so that only the sfclay=1 tries to use these options. If should be a fatal error.

davegill:
@phawbeck @weiwangncar 
I have a commit to share/module_check_a_mundo.F, but no permission to write to this repo.

davegill:
@phawbeck 
> I have a commit to share/module_check_a_mundo.F, but no permission to write to this repo.

Patrick, 
You can either make davegill a collaborator or add in a mod to share/module_check_a_mundo.F so that the option is restricted to specific surface layer schemes. We are trying to move PRs into develop, and this only needs this last change.

davegill:
@weiwangncar @phawbeck 
Wei,
Go ahead and review this. I am going to approve. Once the code is in the repo, I'll add the short IF test for check_a_mundo

davegill (approved):

davegill:
@weiwangncar 
Wei,
Go ahead and review this (I already approved it). Once the code is in the repo, I'll add the short IF test for check_a_mundo as a separate PR since I have no permissions on this organization.

phawbeck:
> @phawbeck
> 
> > I have a commit to share/module_check_a_mundo.F, but no permission to write to this repo.
> 
> Patrick, You can either make davegill a collaborator or add in a mod to share/module_check_a_mundo.F so that the option is restricted to specific surface layer schemes. We are trying to move PRs into develop, and this only needs this last change.

@davegill - I just added you as a collaborator on the repo. (I assume it sent you a notification, but just in case it didn't...).

phys/module_sf_sfclayrev.F
```diff
@@ -1103,9 +1134,74 @@ SUBROUTINE sfclayrevinit
...
+   REAL,    DIMENSION( ims:ime, jms:jme )                   ::  lakemask  
+
+   ! Local 
+   LOGICAL :: overwrite_water_depth
+
+   overwrite_water_depth = .False.
+
+   IF ( bathymetry_flag .eq. 1 ) THEN
+      IF ( shalwater_depth .LE. 0.0 ) THEN
+         CALL wrf_message ( " Bathymetry dataset from GEBCO Compilation Group. Please acknowledge the following in presentations and publications: GEBCO Compilation Group (2021) GEBCO 2021 Grid (doi:10     .5285/c6612cbe-50b3-0cff-e053-6c86abc09f8f)." )
```
weiwangncar, line 1170: @phawbeck Should this be in the code or added to the release note? We don't have any other prints from the code to do this.

run/README.namelist
```diff
@@ -1068,6 +1068,12 @@ Namelist variables for lake module:
                                                   the lake depth data, but this switch is set to 1, the program will stop and tell one to go back to geogrid
                                                   program. 
                                                   = 0, do not use lake depth data.
+ shalwater_rough                     = 1,       ! Shallow water roughness scheme on/off.
+                                                  Bathymetry dataset from GEBCO Compilation Group. Please acknowledge the following in presentations and 
+                                                  publications: GEBCO Compilation Group (2021) GEBCO 2021 Grid (doi:10.5285/c6612cbe-50b3-0cff-e053-6c86abc09f8f)
+ shalwater_depth                     = 0.0,     ! Set constant depth [m] for shallow water roughness scheme if no bathymetry data availabile.
+                                                  The scheme is intended for depths between 10.0 and 100.0 m. Any depths outside of this range will be
+                                                  rounded to the nearest limit value.
```
weiwangncar, line 1076: @phawbeck Can you add the option values and their definition here (< 0, > 0, etc.)?

phys/module_sf_sfclayrev.F
weiwangncar: @phawbeck I see that this message is in the README file. I think adding it to the RELEASE NOTE should be sufficient.

phawbeck: I don't feel strongly one way or the other, so whatever you recommend! Thanks!

run/README.namelist
phawbeck: I'm not sure what else to include for this - shalwater_rough is either 0 or 1 (off/on), and shalwater_depth is a constant depth in meters that will be set for the entire water_depth field. So, if shalwater_depth is less than 10 m, then it will be set to 10 m. If it is larger than 100 m, then it will be set to 100 m. Does that make sense or should I re-word the description?

weiwangncar: @phawbeck What does negative value mean, for example?

phawbeck: Ah ok, I will add a statement for negative values (shalwater_depth must be positive).

phawbeck:
@davegill I'm not sure what that last message ("dismissed stale review") means... I tried to incorporate the changes to the README.namelist per @weiwangncar and had to pull in the new changes in order to push the new commit. That dismiss action seemed to happen because of it. Let me know if there's anything I can do to fix this!

davegill:
@phawbeck 
Patrick,
I'll see if I can find my mod. It's laying around SOMEWHERE!

That "dismissed stale review" refers to our settings that we have put on the WRF github repository. I favorably reviewed the code, but then a modification was made to the code. That just means that we need to have the code reviewed again. So github dismissed my previous (stale) review, because it no longer is relevant. You did nothing wrong at all, and github is just informing everyone that we still need two reviews. Folks who write compiler error messages and github informational messages are on a whole other level of being able to be completely accurate and completely uninformative.

davegill (dismissed):

davegill:
@weiwangncar @phawbeck 
Wei,
I approved this PR
1. I have added in the "acknowledge these folks" in the release notes.
2. I added in the check_a_mundo test for sfclay=1 when shalwater_rough=1

Registry/Registry.EM_COMMON
```diff
@@ -2566,6 +2569,8 @@ rconfig   integer sf_ocean_physics        namelist,physics      1            0
 rconfig   integer traj_opt                namelist,physics      1            0         h     "traj_opt"    "activate trajectory calculation  0=no, 1=on"   ""
 rconfig   logical dm_has_traj             namelist,physics      max_domains  .false.   rh    "has_traj"    "activate trajectory calculation per domain"   ""
 rconfig   integer tracercall              namelist,physics      1            0         h     "tracercall"  "activate tracer calculation  0=no, 1=on"   ""
+rconfig   integer shalwater_rough         namelist,physics      max_domains  0         rh    "shalwater_rough"     "shallow water roughness flag"      ""
+rconfig   real    shalwater_depth         namelist,physics      1            -1.0      rh    "shalwater_depth"     "water depth for shallow water scheme"      "m"
```
weiwangncar, line 2573: @phawbeck The default value set in the registry is -1. What does this signify?

phys/module_physics_init.F
```diff
@@ -755,14 +757,18 @@ SUBROUTINE phy_init ( id, config_flags, DT, restart, zfull, zhalf,     &
   real,    dimension( ims:ime,-nlevsnow+0:nlevsoil, jms:jme ),INTENT(inout) :: zi3d
   LOGICAL, DIMENSION( ims:ime, jms:jme ),intent(out)                      :: lake2d
 !  REAL,    DIMENSION( ims:ime, jms:jme ), INTENT(IN)                      ::  HT
-  REAL, OPTIONAL,    DIMENSION( ims:ime, jms:jme ), INTENT(IN)    ::  lake_depth
+  REAL, OPTIONAL,    DIMENSION( ims:ime, jms:jme ), INTENT(INOUT) ::  lake_depth
+  REAL, OPTIONAL,    DIMENSION( ims:ime, jms:jme ), INTENT(INOUT) ::  water_depth
```
weiwangncar, line 761: @phawbeck The variable water_depth does not need to declared as optional any more.

phys/module_physics_init.F
```diff
@@ -2938,13 +2948,17 @@ SUBROUTINE bl_init(STEPBL,BLDT,DT,RUBLTEN,RVBLTEN,RTHBLTEN,  &
   real,    dimension( ims:ime,-nlevsnow+0:nlevsoil, jms:jme ),INTENT(inout) :: zi3d
  
   logical,    dimension(ims:ime,jms:jme ),intent(out)                        :: lake2d
-  REAL, OPTIONAL,    DIMENSION( ims:ime, jms:jme ), INTENT(IN)    ::  lake_depth
+  REAL, OPTIONAL,    DIMENSION( ims:ime, jms:jme ), INTENT(INOUT) ::  lake_depth
+  REAL, OPTIONAL,    DIMENSION( ims:ime, jms:jme ), INTENT(INOUT) ::  water_depth
```
weiwangncar, line 2952: @phawbeck Variable water_depth shouldn't need to be an optional variable.

phys/module_sf_sfclayrev.F
```diff
@@ -190,6 +192,10 @@ SUBROUTINE SFCLAYREV(U3D,V3D,T3D,QV3D,P3D,dz8w,                    &
 
       INTEGER,  OPTIONAL,  INTENT(IN )   ::     ISFTCFLX, IZ0TLND
       INTEGER,  OPTIONAL,  INTENT(IN )   ::     SCM_FORCE_FLUX
+
+      INTEGER,  OPTIONAL,  INTENT(IN )   ::     shalwater_rough 
+      REAL,     OPTIONAL,  INTENT(IN )   ::     shalwater_depth 
+      REAL, OPTIONAL, DIMENSION( ims:ime, jms:jme ), INTENT(IN)  :: water_depth
```
weiwangncar, line 198: @phawbeck The three new variables here do not need to be optional variables.

phys/module_sf_sfclayrev.F
```diff
@@ -321,6 +329,9 @@ SUBROUTINE SFCLAYREV1D(J,UX,VX,T1D,QV1D,P1D,dz8w1d,                &
                                     
       REAL,     INTENT(IN   )               ::   CP,G,ROVCP,R,XLV,DX
 
+      INTEGER,  OPTIONAL,  INTENT(IN )   ::     shalwater_rough 
+      REAL,     OPTIONAL,  INTENT(IN )   ::     shalwater_depth 
+      REAL, OPTIONAL, DIMENSION( ims:ime ), INTENT(IN)  :: water_depth
```
weiwangncar, line 334: @phawbeck Same here.

phys/module_sf_sfclayrev.F
weiwangncar: @phawbeck Can you then remove this print?

phys/module_surface_driver.F
```diff
@@ -1412,6 +1413,9 @@ SUBROUTINE surface_driver(                                         &
                                                                                tkmg3d,         &
                                                                                tkdry3d,        &
                                                                                tksatu3d
+    INTEGER , OPTIONAL , INTENT(IN)                            :: shalwater_rough 
+    REAL,     OPTIONAL , INTENT(IN)                            :: shalwater_depth 
+    real, OPTIONAL , dimension(ims:ime,jms:jme ),intent(inout) :: water_depth
```
weiwangncar, line 1418: @phawbeck The new variables here do not need to be declared as optional variables.

phys/module_surface_driver.F
```diff
@@ -5912,6 +5921,9 @@ SUBROUTINE sfclayrev_seaice_wrapper(U3D,V3D,T3D,QV3D,P3D,dz8w,     &
                INTENT(INOUT)   ::                            ustm
 
      INTEGER,  OPTIONAL,  INTENT(IN )   ::     ISFTCFLX,IZ0TLND
+     INTEGER,  OPTIONAL,  INTENT(IN )   ::     shalwater_rough 
+     REAL,     OPTIONAL,  INTENT(IN )   ::     shalwater_depth 
+     REAL, OPTIONAL, DIMENSION( ims:ime, jms:jme ) , INTENT(IN ) :: water_depth
```
weiwangncar, line 5926: @phawbeck Here too. No optional needed.

weiwangncar:
@phawbeck @davegill I'm ok with not worrying about declaring the three new variables optional in this PR. But I'd like to know what the default value of -1 mean for variable shalwater_depth.

Registry/Registry.EM_COMMON
phawbeck: Hi @weiwangncar - This would come into play in the following case:
1. User sets shalwater_rough = 1
2. User does not have bathymetry data
3. User did not set shalwater_depth in the namelist
This would cause the value for shalwater_depth to default to -1, and a negative value would cause the model to produce an error. This prevents it from being activated accidentally / erroneously as setting the default value to something like 40 m might be problematic in some situations (e.g., simulations over deep water areas).

That said, I am open to changing it if this is an issue!

phawbeck:
> 

@weiwangncar - I'm going to take out the optional statements and recompile to verify it ran smoothly. I'll update the PR shortly! See above comment about the default shalwater_depth value.

phawbeck:
@weiwangncar - I have removed the optional declarations where you noted and recompiled successfully.

phys/module_sf_sfclayrev.F
weiwangncar: @phawbeck Can you remove this message, as we discussed?

phys/module_sf_sfclayrev.F
```diff
@@ -1103,9 +1134,74 @@ SUBROUTINE sfclayrevinit
...
+            END DO
+         END IF
+      ELSE
+         overwrite_water_depth = .True.
+      END IF
+   ELSE
+      IF ( shalwater_depth .GT. 0.0 ) THEN
+         overwrite_water_depth = .True.
+      ELSE
+         CALL wrf_error_fatal('No bathymetry data detected and shalwater_depth not greater than 0.0. Re-run WPS to get bathymetry data or set shalwater_depth > 0.0')
```
weiwangncar, line 1187: @phawbeck This is probably ok, but in the future namelist cross-checks probably should be placed in share/module_check_a_mundo.F. 
@davegill What do you think?

phys/module_sf_sfclayrev.F
phawbeck: Done!

Registry/Registry.EM_COMMON
weiwangncar: @phawbeck This is probably ok. Usually the default value put in the registry file is a valid option/value, representing expected default behavior. In this case, it may be ok to force a user to define the depth so that the user knows what is used.

weiwangncar:
@phawbeck Thanks for removing optional variable declarations. One last comment I have is what you think about changing the namelist option shalwater_rough to shalwater_z0?

weiwangncar:
@phawbeck Can you let us know if running this option with 1 and multiple processors would produce identical results? You can use MPI/dmpar compiled code, run a case with one processor and then multiple. You can then use external/io_netcdf/diffwrf utility to compare results coming out the two runs. We would like to see identical results. Also, we would also like to see if a restart can produce identical results when compared to a continuous run. Thanks.

phawbeck:
@weiwangncar I will have to check with Pedro and Jimy about changing the name to `shalwater_z0` and get back to you. It might not be until the new year.

As for the additional runs, I can give this a shot. But, again, this will probably have to wait until the new year.

weiwangncar:
@phawbeck Thanks. Waiting until new year wouldn't be problem, especially for those bit-for-bit and restart tests. Those are important validations for the code.

phawbeck:
@weiwangncar ok - no problem. I just got an automated email about the WRF regression tests that were completed on this PR and it looks like there were maybe tests that compared bit-to-bit(?) and they all passed. Should I still try to run my own?

The email shows:
```
    Number of Simulations  : 159           155        0
    Number of Comparisons  : 96           93        0

    Failed Simulations are: 
    None
    None
```

weiwangncar:
@phawbeck I believe the regression tests do not test this branch with your option turned on.

phawbeck:
@weiwangncar understood - I'll set something up and run for simulations with this turned on. Do I need to test each setup (i.e., w/ bathymetry and with a constant depth)?

weiwangncar:
@phawbeck If constant depth option can test the same physical code, it may be ok to test that option. But if we learnt anything from the past, the options that we did not test tend to have problems we didn't foresee at the time.

davegill:
@phawbeck @weiwangncar @dudhia 
Patrick,
These test that we are requested can be very SHORT - perhaps a dozen time steps. The test case does not need to be large, perhaps 70x70 to 100x100 is adequate. For checking the data, I would recommend that you output each time period in a separate file.
1. Build the code with no optimization (in WRF, that is with the `configure -d` option).
2. Build the code with MPI, serial, and openmp. At the end of the simulation (again, really short), are the results identical. Since this is not a performance test, it would be OK to only test GNU (since it builds relatively fast, so those builds would be 32, 33, and 34). The `WRF/external/io_netcdf diffwrf <wrf_file1> <wrf_file2>` command is able to give you a TRUE/FALSE on identical results. If you just get a printout of column headers, then the files are identical.
3. For one of the tests above (which ever is most convenient), drop a restart file half way through the short simulation. Re-run that simulation using the restart file. Check that the restart results vs the original results are also bit-wise identical.

weiwangncar:
@phawbeck Would you like to make a PR for WPS to add the bathymetry data to GEOGRID.TBL? Or can you share that section with us so that we can add it to the table?

phawbeck:
@davegill Thanks a lot for the explanation! I will try to get these done soon, but they might have to be pushed until after AMS.
@weiwangncar I worked with @kkeene44 on getting bathymetry data into WPS a little while back, but I'm not sure what the status of that is at the moment.

davegill:
@phawbeck 
Patrick,
Earliest for a develop branch freeze is 19-ish Jan, and that would be probably hard to do for a number of people.

phawbeck:
@davegill Ok - I will try to get them done by then. Thanks for the heads up.

weiwangncar:
@phawbeck I thought you might already have something since you used that data from geogrid. If you don't we can create something.

phawbeck:
@weiwangncar Sorry - maybe I misunderstood... I worked with @kkeene44 to get the bathymetry data that I use into WPS, so Kelly might know what the status of that is. Should I open a separate PR for that? Or is there something else that you're looking for me to do? Thanks!

weiwangncar:
@phawbeck We've got the data. No PR is needed for that. But to use that data, a section in WPS/geogrid/GEOGRID.TBL is needed which tells geogrid program where the data is and how the data is interpolated. That is the section we are interested.

phawbeck:
@weiwangncar ahh got it! See attached! (I had to add ".txt" to the end for GitHub to allow it to upload)
[GEOGRID.TBL.txt](https://github.com/wrf-model/WRF/files/7818436/GEOGRID.TBL.txt)

weiwangncar:
@phawbeck This is it! Thanks. We will make WPS PR to add this.

phawbeck:
@davegill I just ran a restart from an old simulation with the shallow water roughness scheme and it failed because the bathymetry_flag is only assigned to the input files. I added it to restarts and outputs, recompiled, and ran the same test and it passed. I will work on the other tests in the coming days.

davegill:
@phawbeck 
Patrick,
Very nice indeed.

phawbeck:
@davegill I ran the compile tests both with and without the "-d" option for GNU (32, 33, and 34). Comparing those ran with "-d", the return from diffwrf showed only column headers. The same result was shown for compiling without "-d" between each option. 

The restart tests, however, are showing the files are different. I ran the test again turning off the shallow water roughness scheme and even that is showing that the files are different. Maybe this is something specific to my setup and having some physics options on that shouldn't be.

davegill:
@phawbeck @weiwangncar 
Patrick,
Can you pass along your namelist to me (gill@ucar.edu).

davegill:
@phawbeck @weiwangncar 
Patrick,
Take a look on cheyenne in the /glade/scratch/gill/WRF_shallow_water/test/em_real directory. There is are two namelists: nml1 and nml2. These set `shalwater_rough = 0,`. Would you re-run this test with `shalwater_rough = 1,`.

```
set O = ORIG
set R = RESTART
set F = wrfout_d01_2019-07-30_00:10:15
set DIFF = ../../external/io_netcdf/diffwrf
mkdir $O
mkdir $R
```
Then the commands to run are:
```
rm $O/*
cp nml1 namelist.input
./wrf.exe > & print.out
mv print.out wrfo* $O

rm $R/*
cp nml2 namelist.input
./wrf.exe > & print.out
mv print.out wrfo* $R
```
The diffwrf command should just give the column headers
```
$DIFF $O/$F $R/$F

 Just plot  F
Diffing ORIG/wrfout_d01_2019-07-30_00:10:15 RESTART/wrfout_d01_2019-07-30_00:10:15
 Next Time 2019-07-30_00:10:15
     Field   Ndifs    Dims       RMS (1)            RMS (2)     DIGITS    RMSE     pntwise max
```

phawbeck:
@davegill - I just finished running the test you asked for (shalwater_rough = 1) here: 
`/glade/scratch/hawbecke/WRF/MMC/sandbox/shalwater_rough_tests/GillTest/em_real`

The output from diffwrf was just the headers, as expected:
```
cheyenne2:em_real➜ ./diffwrf ORIG/wrfout_d01_2019-07-30_00\:10\:15 RESTART/wrfout_d01_2019-07-30_00\:10\:15 
 Just plot  F
Diffing ORIG/wrfout_d01_2019-07-30_00:10:15 RESTART/wrfout_d01_2019-07-30_00:10:15
 Next Time 2019-07-30_00:10:15
     Field   Ndifs    Dims       RMS (1)            RMS (2)     DIGITS    RMSE     pntwise max
```

davegill:
@weiwangncar @dudhia 
Folks,
1. The PR provides demonstration that it behaves as it should.
2. There is no negative impact on the existing code - all the regtests pass.
3. The developer conducted the parallel tests and the restart tests.

This is ready for a review.

weiwangncar:
I added '3. The code passed bit-for-bit check and restart test.' in the PR.

weiwangncar (dismissed):
This code has passed bit-for-bit and restart tests.

phawbeck:
> @phawbeck Thanks for removing optional variable declarations. One last comment I have is what you think about changing the namelist option shalwater_rough to shalwater_z0?

@weiwangncar @davegill - I forgot about this comment from a while back. I got the go-ahead to change the name from `shalwater_rough` to `shalwater_z0` from the collaborators. Should I go ahead and make this change?

weiwangncar:
@phawbeck Apparently I forgot about the variable name change too. Since you got an 'OK' from your collaborators, could you make the change, please? Thanks!

phawbeck:
I've updated the name to `shalwater_z0` everywhere. Thanks, @weiwangncar, for the suggestion!

davegill:
@weiwangncar 
Wei,
With the name modification, a re-review is required. :)

davegill:
@phawbeck 
Patrick,
Would you please update the PR commit message to reflect the name change? Thanks

weiwangncar (approved):

weiwangncar:
@phawbeck I've updated the PR message to reflect the name change. Please check though.

phawbeck:
@weiwangncar Thanks a lot! Looks good to me.

davegill (approved):

PR #1552, opened 2021-09-09
Parallel netcdf-4 IO option

MicroTed:
TYPE: new feature

KEYWORDS: parallel netcdf4

SOURCE: Ted Mansell (NOAA/National Severe Storms Lab)

DESCRIPTION OF CHANGES:
This PR adds an I/O option to use the new netcdf capability in 4.7.4 (and later versions) to write both in parallel 
and with variable compression for the same file. Versions of pnetcdf permitted parallel I/O, but did not permit 
standard HDF5 compression. A new I/O library directory, `external/io_netcdfpar`, has been added that is 
modeled on the similar `external/io_netcdf` directory. Users should see the file doc/README.netcdf4par for 
details on configuring and running.

In a nutshell, steps to use this feature:
1. This capability to use HDF5 compression and to have parallel I/O (similar to pnetcdf) requires NetCDF v4.7.4 
or later.
2. The NetCDF library must be built with MPI.
3. Once the NetCDF library / module is chosen and the NETCDF env variable is correctly pointing to the right 
location, then prior to the `configure` step, the user sets a new environment variable. For example, in csh or 
bash `setenv NETCDFPAR $NETCDF` or `export NETCDFPAR=$NETCDF`, respectively.
4. Set the io_form to 13 to use the parallel compressed netcdf option (usually just for the model output or restarts).
5. This I/O option requires activating the NOCOLONS switch (which is why only model output or restarts are 
recommended, since metgrid files have colons imbedded in the file names).

LIST OF MODIFIED FILES: 
M       Makefile
M       arch/Config.pl
M       arch/md_calls.inc
M       arch/postamble
M       arch/preamble
M       configure
A       doc/README.netcdf4par
M       external/Makefile
A       external/io_netcdfpar/Makefile
A       external/io_netcdfpar/diffwrf.F90
A       external/io_netcdfpar/ext_ncdpar_get_dom_ti.code
A       external/io_netcdfpar/ext_ncdpar_get_var_td.code
A       external/io_netcdfpar/ext_ncdpar_get_var_ti.code
A       external/io_netcdfpar/ext_ncdpar_put_dom_ti.code
A       external/io_netcdfpar/ext_ncdpar_put_var_td.code
A       external/io_netcdfpar/ext_ncdpar_put_var_ti.code
A       external/io_netcdfpar/field_routines.F90
A       external/io_netcdfpar/module_wrfsi_static.F90
A       external/io_netcdfpar/transpose.code
A       external/io_netcdfpar/wrf_io.F90
M       frame/md_calls.m4
M       frame/module_io.F
M       share/mediation_integrate.F
M       share/module_io_domain.F
M       share/output_wrf.F
M       share/wrf_ext_write_field.F

TESTS CONDUCTED: 
1. Tested on lustre file system (cray) with domain sizes up to about 700x700. Will leave the existing chunking as
as the initial implementation.
2. Parallel I/O with compression successfully runs on the NCAR cheyenne system, a GPFS file system. The 
following were set up for modules during the build and within the job script:
```
module purge
module load intel
module load ncarcompilers
module load mpt
module load netcdf-mpi
module load ncarenv
```
3. The tests on cheyenne worked with both GNU/10.1.0 and Intel/19.1.1
4. Jenkins testing is OK.
5. Additional information is now output at the end of the `configure` step regarding I/O options:
```
NetCDF version: 4.8.1
Enabled NetCDF-4/HDF-5: yes
NetCDF built with PnetCDF: no
Enabled NetCDF parallel: yes
Using parallel NetCDF via NETCDFPAR option
```

RELEASE NOTE: Added the ability to write compressed netcdf4 files in parallel via NetCDF 4.7.4 (and later). The performance is slower than pnetcdf, but can be notably faster than regular netcdf4 on parallel file systems. As expected, the compression provides files significantly smaller than pnetcdf generates.


davegill:
@MicroTed 
Ted,
This is both great and a huge amount of work! Here are some preliminary questions. This is scheduled for the release branch, so spring 2022.

1. Would you please add your affiliation to this PR commit message?
2. Move the new README file into the top-level "doc" directory.
3. What are the consequences of NOT having a new enough version of netcdf?
4. Are there ways to disable the parallel writing?
5. Is there an accessible "chunk" variable that could be used for tuning?

MicroTed:
@davegill 
Thanks!
> 1. Would you please add your affiliation to this PR commit message?
  -- done
> 2. Move the new README file into the top-level "doc" directory.
 -- done
> 3. What are the consequences of NOT having a new enough version of netcdf?
 I did not add any check for netcdf version. I suppose could be done with 'nc-config --version' and some string testing?
> 4. Are there ways to disable the parallel writing?
 -- Yes, just switch to io_form =2. The regular serial netcdf interface works fine with the parallel-compiled libraries.
> 5. Is there an accessible "chunk" variable that could be used for tuning?
 -- Hmm, this is a good thought. Like a namelist variable? The testing I've done suggests it is more important to have uniform tile sizes so that they don't overlap chunks. But it would definitely be worth experimenting more with it. I replaced the chunk calculation from the io_netcdf/wrf_io.F90 with a simpler setup that I've used in the COMMAS cloud model, which only does evenly-divisible tiles for domain decomposition.

davegill:
@MicroTed 
Ted, 

Your mods do not break the standard IO output options. I can get data output with io_form = 2. But, I am having troubles getting this code to work with the new capability. Here is the specific error message:
```
open_hist_w : error opening wrfout_d01_2021-11-17_06_00_00 for writing. ***
```

1. I am using a recent enough version of netcdf:
```
> nc-config --version
netCDF 4.8.0
```
2. The NETCDFPAR env variable points to the correct location:
```
> echo $NETCDF
/glade/u/apps/ch/opt/netcdf/4.8.0/gnu/11.1.0/
> echo $NETCDFPAR
/glade/u/apps/ch/opt/netcdf/4.8.0/gnu/11.1.0/
```
3. The namelist is set up to output the new format:
```
 &time_control
 nocolons                            = .true.
 io_form_history                     = 13
 /
```
4. The code is built for MPI. I have tried various processor counts, from 1 through 36. The domain is relatively small, so there should not be any troubles in that regard.

Any ideas or suggestions?

MicroTed:
@davegill 
Such a helpful error message, isn't it? I have only tested with netcdf 4.7.4 and Intel fortran. I expect you have it set up correctly, and it should work on any file system (serial disk just has no speedup). Nothing comes to mind, though I suppose it is possible that there is a bug affecting gfortran but not ifort? I could give 4.8.0 and gfortran a try myself, but I'm a bit strapped for time at the moment -- I'll see if I can test that.

davegill:
@MicroTed 
Ted,
I have tried with Intel/21 and GNU/11, both with netcdf/4.8.0

MicroTed:
That's not a good sign that it's failing for you.... Here's a namelist for the em_quarter_ss test case that I used for simple tests. (I had to add '.txt' to get it to upload) If that  works (or doesn't), we can go from there. Those two namelist item should be all that is needed, though.

[namelist.input.txt](https://github.com/wrf-model/WRF/files/7557231/namelist.input.txt)

MicroTed:
@davegill 
Out of curiosity, which version of hdf5? I used 1.10.7

> I have tried with Intel/21 and GNU/11, both with netcdf/4.8.0

davegill:
@MicroTed 
Ted,
Our "module" command handles this. I know that netcdf was built with hdf5, since nc-config --all lists:
```
 --has-hdf5      -> yes
```

I am guessing HDF5 v1.10.7, as that is the only option offered - but I am having troubles tracking this down.

MicroTed:
@davegill 
I found the problem with the metadata, which was that I missed porting the update to frame/md_calls.m4. I pushed those changes (also to the backup arch/md_calls.inc).

davegill:
@MicroTed @weiwangncar @kkeene44 
Ted,
Thanks again for this contribution!

Wei and Kelly,
This is ready for review. I have worked with Ted to iron out the kinks on my end, and have added in the required diffwrf utility. It is a niche capability, but it is fully functional.

davegill (approved):

weiwangncar:
@davegill Have you retested code with this latest change?

davegill:
@weiwangncar 
> Have you retested code with this latest change?

Wei,
This is tested, it works, and it is ready to go. For quite a while, the only trouble was (unbeknownst to me) getting the environment INSIDE of the qsub'ed job to be identical to the environment used during the build.

weiwangncar (approved):

PR #1563, opened 2021-09-27
Revert print and when to turn on damping to pre 4.3 code

weiwangncar:
TYPE: bug fix

KEYWORDS: w_damp, cfl print criterion, IEVA

SOURCE: internal

DESCRIPTION OF CHANGES:

IEVA = Implicit Explicit Vertical Advection

Problem:
Code in w_damp subroutine was changed in 4.3 to turn on w_damping at CFL = 1.2, and print CFL when it exceeded 1.2. 
Prior to v4.3, w_damping was turned on when CFL = 1.0, and printed the message CFL message when the threshold 
value of 2.0 was exceeded. This change results many more CFL prints. We want to revert to the original behavior.

Solution:
This PR fixes the CFL warning issues by changing the threshold for notification back to what it was before 4.3. 

Importantly, this modification does not change the behavior that sets a higher CFL value to turn on w_damping when
IEVA is used. 

LIST OF MODIFIED FILES: 
M     dyn_em/module_big_step_utilities_em.F

TESTS CONDUCTED: 
1. Do mods fix problem? Yes.
2. Jenkins tests are all PASS.

RELEASE NOTE: Code in w_damp subroutine was changed in 4.3 to turn on w_damping at CFL = 1.2, and print CFL when it exceeded 1.2. This change resulted in many more CFL prints. This PR reverts both changes to what it was before v4.3, where w_damping starts at CFL = 1.0, and CFL prints appear when CFL exceeds 2.0. When IEVA is on, w_damping starts at CFL = w_crit_cfl set in the namelist, and CFL prints only appear for those exceeding 2.0.

### DIFF
--- a/dyn_em/module_big_step_utilities_em.F
+++ b/dyn_em/module_big_step_utilities_em.F
@@ -2550,14 +2550,14 @@ SUBROUTINE w_damp( rw_tend, max_vert_cfl,max_horiz_cfl, &
    integer :: total
    REAL :: msfuxt , msfxffl
    
-   REAL    :: w_damp_on     = 1.0
+   REAL    :: w_damp_on
    REAL    :: w_crit_cfl    = 2.0
    REAL    :: w_flag_cfl    = 1.2
-   LOGICAL :: wflags_differ = .false.
    LOGICAL :: print_flag    = .true.
    SAVE    :: print_flag
    INTEGER :: some1       ! Now have two catagories of CFL information, hence some1 & some2
    INTEGER :: some2        
+   INTEGER :: ieva
 
 !<DESCRIPTION>
 !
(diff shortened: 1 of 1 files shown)


dyn_em/module_big_step_utilities_em.F
```diff
@@ -2722,7 +2698,7 @@ SUBROUTINE w_damp( rw_tend, max_vert_cfl,max_horiz_cfl, &
      CALL get_current_time_string( time_str )
      CALL get_current_grid_name( grid_str )
      WRITE(temp,*) some1,                                            &
-            ' points exceeded w_critical_cfl in domain '//TRIM(grid_str)//' at time '//TRIM(time_str)//' hours'
+            ' points exceeded v_cfl = 2 in domain '//TRIM(grid_str)//' at time '//TRIM(time_str)//' hours'
```
davegill, line 2701: @weiwangncar 
Wei,
Should this print line use the value of w_crit_cfl instead of 2?

davegill: understood

davegill (approved):

dyn_em/module_big_step_utilities_em.F
```diff
@@ -2687,7 +2672,7 @@ SUBROUTINE w_damp( rw_tend, max_vert_cfl,max_horiz_cfl, &
 ! This internal write is costly on newer Xeon processors because it breaks
 ! vectorization.  (J. Michalakes for L. Meadows at Intel, 12/13/2012)
 
-         IF (vert_cfl .gt. w_crit_cfl) THEN
```
dudhia, line 2690: Maybe we can use w_crit_cfl when ieva is on and 2 when it is off. This could be done where w_beta is used above with another local variable like cflprint. Line in print would also need modifying.

weiwangncar: @dudhia That is assuming someone would set w_crit_cfl other than 2?

dudhia: Mainly to keep the ieva print as it was. Perhaps Lou wants that threshold shown when ieva is on.

weiwangncar: @dudhia Lou recommended using w_crit_cfl = 2 when turning on ieva, so 2 seems to be good criterion for printing.

dudhia:
Is that going to be the default because part of the problem was that it
wasn't 2.

dyn_em/module_big_step_utilities_em.F
weiwangncar: @dudhia w_crit_cfl can be set in the namelist, and the recommended value is 2 if ieva is on. This value happens to be the value we turn on CFL prints in the past. Another way to fix this partially is to change the registry value of w_crit_cfl from 1.2 to 1. This will turn on w_damp at this value, but print criterion still needs to be changed.

dudhia (approved):
Can you also state in the release note whether w_damping behavior is also changed back by using w_beta again?

weiwangncar:
@dudhia Just did that.

PR #1564, opened 2021-09-28
Increase total time periods available to process for climate run

### LINKED ISSUE #1383: times limit on the wrfbdy files
Currently if more than 9999 times are included in the wrfbdy file (for example, by running 20 year regional simulations), the wrfbdy files will have **** in the timestamps and wrf crashes (or hangs). The same seems to happen for ungrib.exe (it hangs if more than 9999 files are being processed).

I run multiple climatic simulations (20+ years) and I have to break the runs in 5 year segments (where I have to run real.exe to make new wrfbdy files and the next runs starts with the last restart of the previous run) due to this limitation.

Is there any plan to fix this in WRF and WPS?  It's really a pain...

davegill:
TYPE: bug fix
    
KEYWORDS: maxtimes, 10000, climate runs, LBC
    
SOURCE: internal
    
DESCRIPTION OF CHANGES:
Problem:
When processing a LARGE number of time periods in a single lateral boundary file, the WRF
model fails. The failure is always just after time period 9999 is processed. This points to
a likely culprit of 10000 being a maximum number of time periods. Sure enough, there is a
variable name `MaxTimes` in a couple of the netcdf IO directories that is set to 60000.
    
Solution:
Bump this fixed number up to a larger fixed number that is sufficient to handle proposed climate 
simulations at a typical data frequency.
    
ISSUE:
Fixes #1383
    
LIST OF MODIFIED FILES:
modified:   io_netcdf/wrf_io.F90
modified:   io_pnetcdf/wrf_io.F90
    
TESTS CONDUCTED:
1. The user who originally saw the problem indicates that this fixes the problem. Additionally,
the fix works for not only the LBC files, but for the FDDA and SST update files.
2. We are anticipating saying that All jenkins tests are a PASS.
    
RELEASE NOTE: For climate-length WRF simulations, the restriction of 10000 time periods in a single file has been increased to 60000 (a value that allows 3-hourly data for 20 years).


davegill:
@dmoreno11
If you have time, would you please verify that this fixes the simulations that you were trying to conduct. Also, if you would supply your name and affiliation, we'd like to credit you for reporting this bug.

dmoreno11:
> @dmoreno11 If you have time, would you please verify that this fixes the simulations that you were trying to conduct. Also, if you would supply your name and affiliation, we'd like to credit you for reporting this bug.

Hi Dave, thanks for updating me on this. I'll conduct a test and see if the bug is fixed. 

My name is David Carvalho and my affiliation is CESAM-Aveiro University (Portugal)

weiwangncar (approved):

dmoreno11:
I think this issue is also present in WPS (ungrib.exe and metgrid.exe). I think it would be a good idea to fix this bug in ungrib and metgrid also.

davegill:
@kkeene44 @smileMchen 
Ming and Kelly,
Would one of you review this WRF PR? It is pretty simple - it removes a hard-coded restriction on the maximum number of time periods that the model is able to correctly process. It does not use the value for space allocation, just a maximum number of loops.

PR #1576, opened 2021-11-05
configure script missing quotes in two echo statements

davegill:
TYPE: bug fix

KEYWORDS: configure, shell, echo

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
When building with NetCDF that does not support compression, the warning statements that
are printed out are misaligned, due to missing trailing quotes at the end of some `echo` statements in the
`configure` script.

Solution:
Add those quotes back in.

LIST OF MODIFIED FILES:
modified:   configure

TESTS CONDUCTED: 
1. The printout is now aesthetically appealing. Initially we had a mortifying display of the word `echo`, and the 
accompanying alignment disarray that would be associated with banal shell scripting encroaching upon a user's otherwise elevated build experience:
```
Please make sure NETCDF version is 4.1.3 or later and was built with         
--enable-netcdf4                                                             

OR set NETCDF_classic variable                                               
   bash/ksh : export NETCDF_classic=1
      echo  csh : setenv NETCDF_classic 1
```
After much soul searching, an appropriate solution was uncovered, resulting in reduced messaging dissonance:
```
Please make sure NETCDF version is 4.1.3 or later and was built with         
--enable-netcdf4                                                             

OR set NETCDF_classic variable                                               
   bash/ksh : export NETCDF_classic=1                                        
        csh : setenv NETCDF_classic 1     
```
2. Jenkins is all PASS.

### DIFF
--- a/configure
+++ b/configure
@@ -1044,8 +1044,8 @@ if [ -n "$NETCDF4" ] ; then
       echo "--enable-netcdf4                                                             "
       echo
       echo "OR set NETCDF_classic variable                                               "
-      echo "   bash/ksh : export NETCDF_classic=1
-      echo "        csh : setenv NETCDF_classic 1
+      echo "   bash/ksh : export NETCDF_classic=1                                        "
+      echo "        csh : setenv NETCDF_classic 1                                        "
       echo 
       echo "Then re-run this configure script                                            "
       echo


davegill:
Jenkins is OK
```
Please find result of the WRF regression test cases in the attachment. This build is for Commit ID: 64b7a931c94481211c4a5919534951cd6036d254, requested by: davegill for PR: https://github.com/wrf-model/WRF/pull/1576. For any query please send e-mail to David Gill.

    Failed Simulations are: 
    None
    None
```

kkeene44 (approved):

davegill:
@smileMchen 
Ming,
This should also be a simple one to review. I forgot to put in the trailing quote characters. Only the output from two `echo` statements was incorrect.

smileMchen (approved):

Issue #1514, opened 2021-05-10
MYNN in parent, LES in a nest: sometimes the sim does not run.


davegill:
**Describe the bug**
This is an old issue we are having. When we try to use MYNN in one domain and LES in a nest sometimes the sim does not run. And the output provides no clue about what the problem could be. You pointed us a few years ago to the solution, copy the MYNN package definition in the registry and change to a nopbl package if the pbl is set to 0 in the namelist (LES). 

The problem of this fix is that it is particular for MYNN. So it never made it into the official WRF, and it is consuming the time of many users trying to run MYNN with LES.

My idea for a generic fix was to put the fix above for MYNN (the most used PBL) and then in check a mundo put an check to stop any sim with a PBL with a package + LES and print the solution. This will stop consuming user's time and points to the fix.

I briefly talked to Jimy about it and he indicated that you may know a generic fix.

How do you think we should move forward with this to avoid this issue?

Thanks in advance,
Pedro.

**To Reproduce**
1. Use compiler and version '....'
2. Use namelist options '....'
3. Output is '....'

**Expected behavior**

**Screenshots**

**Attachments**
If applicable, attach supporting files: namelist.input, rsl.* files, etc.

**Additional context**
Add any other context about the problem here, such as:
   - This used to work with a previous version.
   - The documentation is different from the exhibited behavior.

mefrediani:
Dave, there's a directory with the run files here: /glade/scratch/frediani/WRF-Fire/runs/cameron_peak_2020/no-fire/WRF/2dom
(Please note that the wrfinput and wrfbdy files are symlinks to subfolders)
Let me know if you have any questions. 

Thanks
Maria

dudhia:
Just to update on what this fix is.
When we use a PBL scheme on any domain, any advected scalars in its package list need to be applied to all domains. This is so that LES nests don't allocate less memory. What is not clear yet is if all state variables in the package list also have to be allocated. For Franciano, the quick fix that seemed to work was to add the qke_adv scalar to the microphysics package.

PR #1578, opened 2021-11-05
Restrict which CG PBLs may use FG LES PBL

davegill:
TYPE: bug fix

KEYWORDS: PBL, LES

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
Only the YSU PBL scheme has no packaged variables. If a user wants to have any other PBL scheme on the CG,
and an LES PBL on the FG, the model will have memory corruption problems. This error is hard to detect.

Solution:
1. The variables el_pbl and tke_pbl are added to the list of packaged variables for the LES PBL. While this
scheme does not need them, this does not add any memory, _and_ this permits both the MYJ and 
ShinHong PBL schemes to be used.
2. A controlled stop is now in place when any incompatible PBL namelist options are selected.

ISSUES:
Partially addresses #1514 

LIST OF MODIFIED FILES:

TESTS CONDUCTED:
1. When an incompatible CG PBL and FG LES are selected, the code successfully stops.
```
--- ERROR: LES PBL on fine grid does not work with CG PBL option  5
--- ERROR: Only YSU(1), MYJPBL(2), and ShinHong(11) are acceptable options for LES PBL on the fine grid
--- ERROR: Fix bl_pbl_physics in namelist.input.
--- ERROR: Alternatively, remove all of the packaged variables from the CG PBL selection
FATAL CALLED FROM FILE:  <stdin>  LINE:    2534
```
2. Jenkins is all PASS.

RELEASE NOTE: A test was introduced to stop the incorrect combination of particular PBL schemes on coarser grids with the LES PBL option selected on a finer grid. Previously, this problem caused an inconsistent number of variables on the CG and FG, which caused segmentation faults when trying to do feedback or advection of unavailable fields.

### DIFF
--- a/Registry/Registry.EM_COMMON
+++ b/Registry/Registry.EM_COMMON
@@ -3086,6 +3086,7 @@ package   drip           sf_surf_irr_alloc==2        -             state:irrigat
 package   sprinkler      sf_surf_irr_alloc==3        -             state:irrigation,irr_rand_field
 
 
+package   lesscheme      bl_pbl_physics==0           -             -
 package   ysuscheme      bl_pbl_physics==1           -             -
 package   myjpblscheme   bl_pbl_physics==2           -             state:tke_pbl,el_pbl
 package   gfsscheme      bl_pbl_physics==3           -             -
--- a/Registry/Registry.EM_COMMON
+++ b/Registry/Registry.EM_COMMON
@@ -3086,7 +3086,7 @@ package   drip           sf_surf_irr_alloc==2        -             state:irrigat
 package   sprinkler      sf_surf_irr_alloc==3        -             state:irrigation,irr_rand_field
 
 
-package   lesscheme      bl_pbl_physics==0           -             -
+package   lesscheme      bl_pbl_physics==0           -             state:tke_pbl,el_pbl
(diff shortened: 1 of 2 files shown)


davegill:
Jenkins is OK
```
Please find result of the WRF regression test cases in the attachment. This build is for Commit ID: 7b6f4290e9e8eb9b3d9ae17a58a2b4406d68de2a, requested by: davegill for PR: https://github.com/wrf-model/WRF/pull/1578. For any query please send e-mail to David Gill.

    Failed Simulations are: 
    None
    None
```

kkeene44 (dismissed):

dudhia:
This is a placeholder stop until a better fix is found. Issue #1514 has some workarounds and can be mentioned in the release note. The stop message could also recommend YSU PBL as a safe option with no packaged arrays. I am not sure if MYJ is safe as it has packaged arrays.

dudhia:
adding el_pbl and tke_pbl to the les package would allow MYJ and shinhong to work with LES in nests (Shin-Hong has these as diagnostics).

davegill:
Choosing #1596 "PBL + LES restrictions" as a temporary solution.

PR #1580, opened 2021-11-05
smooth_cg_topo=T requires soil elevation data

davegill:
TYPE: bug fix

KEYWORDS: ERA5, smooth_cg_topo, elevation data

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
If a user does not include all of the correct ERA5 invariant files, then there could be
no soil elevation field coming from the metgrid program. Without that field, the flag
`flag_soilhgt` is not set. Then the real program will not generate any topo differences along
the lateral boundary (when requested with smooth_cg_topo = T). This is a quiet failure,
and difficult to track down.

Solution:
If the user asks to smooth the lateral boundary topo on d01, AND the user has no input
soil elevation data from the first-guess model, then this is now a fatal error.

LIST OF MODIFIED FILES:
modified:   dyn_em/module_initialize_real.F

TESTS CONDUCTED:
1. The original intent of the option still works - smoothed topo on d01 along the outer few rows and columns.
The first figure shows the change in topography. The second figure below shows which grid cells are
impacted. Basically, these two figures show that topography smoothing still is available.
Screen Shot 2021-11-05 at 11 27 03 AM

Screen Shot 2021-11-05 at 11 45 57 AM

2. By artificially setting the metgrid flags in a debugging mode within the real program, we can successfully 
emulate missing soil elevation data by setting flag_soilhgt = 0. Whether or not we artificially set the value,
there is now a controlled stop to the real program.
```
FATAL CALLED FROM FILE:  <stdin>  LINE:     766
 --- ERROR: NML option smooth_cg_topo=T, but no first guess soil elevation data
```
3. All jenkins tests are PASS.

RELEASE NOTE: A fatal error in the real program is now issued when a user requests lateral boundary topography smoothing, but lacks the first-guess soil elevation data to do the weighted averaging of topography. Previously, this discrepancy was ignored, and the d01 high-resolution elevation data was quietly left undisturbed.

### DIFF
--- a/dyn_em/module_initialize_real.F
+++ b/dyn_em/module_initialize_real.F
@@ -672,7 +672,8 @@ integer::oops1,oops2
               ( internal_time_loop .EQ. 1 ) .AND. &
               ( grid%id .EQ. 1 ) .AND. &
               ( flag_soilhgt .NE. 1) ) THEN
-            CALL wrf_message    (' --- ERROR: NML option smooth_cg_topo=T, but found no soil elevation data in metgrid files')
+            CALL wrf_message    (' --- ERROR: NML option smooth_cg_topo=T')
+            CALL wrf_message    ('            But found no soil elevation / terrain / topography data in metgrid files)
             CALL wrf_message    ('            The field SOILHGT is required when smoothing the CG topography on d01')
             CALL wrf_error_fatal('            If using ERA5 data, possibly need to add more time invariant fields')
          END IF
--- a/dyn_em/module_initialize_real.F
+++ b/dyn_em/module_initialize_real.F
@@ -672,7 +672,8 @@ integer::oops1,oops2
               ( internal_time_loop .EQ. 1 ) .AND. &
               ( grid%id .EQ. 1 ) .AND. &
               ( flag_soilhgt .NE. 1) ) THEN
-            CALL wrf_error_fatal(' --- ERROR: NML option smooth_cg_topo=T, but found no soil elevation data in metgrid files')
+            CALL wrf_message    (' --- ERROR: NML option smooth_cg_topo=T, but found no soil elevation data in metgrid files')
+            CALL wrf_error_fatal('            If using ERA5 data, requires including additional time invariant data files')
          END IF
 
          IF ( ( config_flags%smooth_cg_topo ) .AND. &
--- a/dyn_em/module_initialize_real.F
+++ b/dyn_em/module_initialize_real.F
@@ -672,7 +672,7 @@ integer::oops1,oops2
               ( internal_time_loop .EQ. 1 ) .AND. &
               ( grid%id .EQ. 1 ) .AND. &
               ( flag_soilhgt .NE. 1) ) THEN
-            CALL wrf_error_fatal(' --- ERROR: NML option smooth_cg_topo=T, but no first guess soil elevation data')
+            CALL wrf_error_fatal(' --- ERROR: NML option smooth_cg_topo=T, but found no soil elevation data in metgrid files')
          END IF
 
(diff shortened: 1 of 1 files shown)


davegill:
jenkins is OK
```
Please find result of the WRF regression test cases in the attachment. This build is for Commit ID: fdb4bd3dd5056f9ebca85138afb7e66ba6e71c40, requested by: davegill for PR: https://github.com/wrf-model/WRF/pull/1580. For any query please send e-mail to David Gill.

    Failed Simulations are: 
    None
    None
```

davegill:
@dudhia 
Jimy,
This mod would have fixed Andrew's ERA5 troubles.

dudhia:
So for single domain runs, the elevation is blended to the incoming topo in
the boundary zone with this option?

davegill:
@dudhia 
Jimy,
Correct.
For a real.exe run with any number of domains, the option is to blend the first-guess topo to the outer-most rows and column of d01 with the high-res topo (from geogrid)

davegill:
@kkeene44 
Kelly,
No good deed goes unpunished. Here is the PR related to the topo smoothing.
1. Any suggestions for the user message
2. A review would be appreciated

dyn_em/module_initialize_real.F
```diff
@@ -665,6 +665,16 @@ SUBROUTINE init_domain_rk ( grid &
...
          !  in the model betwixt the CG and FG domains.
 
+         !  An important point is to inform the user if their request cannot
+         !  be satisfied. Do not skip over this quietly.
+
+         IF ( ( config_flags%smooth_cg_topo ) .AND. &
+              ( internal_time_loop .EQ. 1 ) .AND. &
+              ( grid%id .EQ. 1 ) .AND. &
+              ( flag_soilhgt .NE. 1) ) THEN
+            CALL wrf_error_fatal(' --- ERROR: NML option smooth_cg_topo=T, but found no soil elevation data in metgrid files')
```
kkeene44, line 675: @davegill 
Could you add something like "If using ERA5 input, must use include all invariant files."

weiwangncar:
@davegill Here is a suggestion: How about 'but found no (incoming?) terrain field from driving data in metgrid files (field name SOILHGT)'? I'm trying to use words that may 'click' with users...

weiwangncar:
@kkeene44 Is there a place on our web or UG that we can add a note about invariant fields for dataset like ERA5?

kkeene44:
@weiwangncar Yes, I can add something at least in the UG. I already have an FAQ about it on the forum, and our website points to the FAQs, but not everyone looks there.

dyn_em/module_initialize_real.F
```diff
@@ -665,6 +665,18 @@ SUBROUTINE init_domain_rk ( grid &
...
          !  in the model betwixt the CG and FG domains.
 
+         !  An important point is to inform the user if their request cannot
+         !  be satisfied. Do not skip over this quietly.
+
+         IF ( ( config_flags%smooth_cg_topo ) .AND. &
+              ( internal_time_loop .EQ. 1 ) .AND. &
+              ( grid%id .EQ. 1 ) .AND. &
+              ( flag_soilhgt .NE. 1) ) THEN
+            CALL wrf_message    (' --- ERROR: NML option smooth_cg_topo=T, but found no soil elevation data in metgrid files')
```
weiwangncar, line 675: I'd say 'no terrain elevation' data maybe better.. I know the field is called SOILHGT, but it is actually terrain.

smileMchen (approved):
I am OK with this PR.

davegill:
@kkeene44 
Kelly,
I have addressed (I think) your suggestions for the the user facing messages. Would you please review this PR.

kkeene44 (approved):

PR #1592, opened 2021-12-07
Bug fix for aerosol optics Mie extrapolation WRFchem

cenlinhe:
TYPE: bug fix

KEYWORDS: aerosol optics, Mie calculation, interpolation, extrapolation

SOURCE: Cenlin He and Rajesh Kumar (NCAR)

DESCRIPTION OF CHANGES:
Problem:
When the aerosol refractive indices are outside the lookup table ranges (which are very rare occasions), the 
extrapolation in the parameterized Mie calculation ('mieaer' subroutine) will create unrealistic values for the 
extrapolated aerosol optical parameters. These unphysical fluctuations cause unrealistically high AOD 
values (10^10 and greater) and the model crashes.

Solution:
Force the out-of-range aerosol refractive index to be the min or max boundary values of the lookup table.

LIST OF MODIFIED FILES:
chem/module_optical_averaging.F

TESTS CONDUCTED: 
1. This solution fixes the issue and has been tested in NCAR Cheyenne HPC by Cenlin He (NCAR).
2. Jenkins testing is all PASS.

RELEASE NOTE: Minor bugfix for aerosol optics Mie extrapolation cases in chemistry. When the aerosol refractive indices are outside the lookup table ranges, the extrapolation in the parameterized Mie calculation will create unrealistic values for the extrapolated aerosol optical parameters. These unphysical fluctuations cause unrealistically high AOD values and the model crashes. The solution is to force the out-of-range aerosol refractive index to be the minimum or maximum boundary values of the lookup table.


jordanschnell:
@cenlinhe - There is an error check above your lines that with hard coded values for the range. What do you think about adding a warning in your check and/or harmonizing it with the 1e-3 - 10 range check?

cenlinhe:
> @cenlinhe - There is an error check above your lines that with hard coded values for the range. What do you think about adding a warning in your check and/or harmonizing it with the 1e-3 - 10 range check?

@jordanschnell That is actually the inconsistent place in the current chemistry. The 1e-3 - 10 range check does not prevent the extrapolation bug, because the minimum bound of the lookup table is close to 1.0. If the refractive index is much smaller than 1.0, the unstable extrapolation will occur, leading to unrealistic aerosol optics.

jordanschnell:
> > @cenlinhe - There is an error check above your lines that with hard coded values for the range. What do you think about adding a warning in your check and/or harmonizing it with the 1e-3 - 10 range check?
> 
> @jordanschnell That is actually the inconsistent place in the current chemistry. The 1e-3 - 10 range check does not prevent the extrapolation bug, because the minimum bound of the lookup table is close to 1.0. If the refractive index is much smaller than 1.0, the unstable extrapolation will occur, leading to unrealistic aerosol optics.

@cenlinhe I see, I think putting a debug print statement for when they are reset to the bounds would be nice but not necessary

cenlinhe:
@jordanschnell Thanks for the suggestion. I have added a debug printout statement to warn that the values are reset to bounds. Please see the updated commit and file.

davegill:
@jordanschnell 
Let us know if you are OK with these mods, with a review. We are trying to move the develop branch along.

jordanschnell (approved):
Looks good to me. Thanks for adding the warning/print.

davegill (approved):

PR #1596, opened 2021-12-10
PBL + LES restrictions

### LINKED ISSUE #1514: MYNN in parent, LES in a nest: sometimes the sim does not run.

davegill:
TYPE: bug fix

KEYWORDS: PBL, LES

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
A few PBL schemes include scalar components that are advected. These fields are
required on all domains. With an LES option on a finer domain, that grid does not
have the additional scalar component, which causes a segfault when trying to
access that field.

Solution:
Restrict what PBL schemes are eligible for use with the LES (no PBL) option.

ISSUE:
Fixes #1514

LIST OF MODIFIED FILES:

TESTS CONDUCTED:
1. When an incompatible CG PBL and FG LES are selected, the code successfully stops.
```
 &physics
 bl_pbl_physics                      =  5,     0,

--- ERROR: LES PBL on fine grid does not work with CG PBL option  5
           Choose a CG PBL option without any scalar components
           Fix bl_pbl_physics in namelist.input
           Alternatively, remove all of the packaged variables from the CG PBL selection
FATAL CALLED FROM FILE:  <stdin>  LINE:    2523
```
```
 &physics
 bl_pbl_physics                      =  6,     0,

--- ERROR: LES PBL on fine grid does not work with CG PBL option  6
           Choose a CG PBL option without any scalar components
           Fix bl_pbl_physics in namelist.input
           Alternatively, remove all of the packaged variables from the CG PBL selection
FATAL CALLED FROM FILE:  <stdin>  LINE:    2523
```
```
 &physics
 bl_pbl_physics                      = 16,     0,

--- ERROR: LES PBL on fine grid does not work with CG PBL option  16
           Choose a CG PBL option without any scalar components
           Fix bl_pbl_physics in namelist.input
           Alternatively, remove all of the packaged variables from the CG PBL selection
FATAL CALLED FROM FILE:  <stdin>  LINE:    2523
```
2. Hopefully we say All jenkins tests are a PASS.

RELEASE NOTE: A test was introduced to stop the incorrect combination of particular PBL schemes on coarser grids with the LES PBL option selected on a finer grid. Previously, this problem caused an inconsistent number of variables on the CG and FG, which caused segmentation faults when trying to do feedback or advection of unavailable fields.

### DIFF
--- a/Registry/Registry.EM_COMMON
+++ b/Registry/Registry.EM_COMMON
@@ -3086,6 +3086,8 @@ package   drip           sf_surf_irr_alloc==2        -             state:irrigat
 package   sprinkler      sf_surf_irr_alloc==3        -             state:irrigation,irr_rand_field
 
 
+package   lesscheme      bl_pbl_physics==0           -             -
+package   nopblscheme    bl_pbl_physics==0           -             -
 package   ysuscheme      bl_pbl_physics==1           -             -
 package   myjpblscheme   bl_pbl_physics==2           -             state:tke_pbl,el_pbl
 package   gfsscheme      bl_pbl_physics==3           -             -
--- a/Registry/Registry.EM_COMMON
+++ b/Registry/Registry.EM_COMMON
@@ -3087,7 +3087,6 @@ package   sprinkler      sf_surf_irr_alloc==3        -             state:irrigat
 
 
 package   lesscheme      bl_pbl_physics==0           -             -
-package   nopblscheme    bl_pbl_physics==0           -             -
 package   ysuscheme      bl_pbl_physics==1           -             -
 package   myjpblscheme   bl_pbl_physics==2           -             state:tke_pbl,el_pbl
 package   gfsscheme      bl_pbl_physics==3           -             -
--- a/share/module_check_a_mundo.F
+++ b/share/module_check_a_mundo.F
@@ -406,6 +406,34 @@
          END IF
       ENDDO
 
+
+!-----------------------------------------------------------------------
+! Check that LES PBL is only paired with acceptable other PBL options.
+! Currently, problems occur with any CG PBL option that has a packaged 
+! scalar (as opposed to state) component: MYNN2, MYNN3, EEPS.
+!-----------------------------------------------------------------------
+      DO i = 1, model_config_rec % max_dom
+         IF ( .NOT. model_config_rec % grid_allowed(i) ) CYCLE
+         IF ( model_config_rec % bl_pbl_physics(i) .EQ. LESscheme ) THEN
+            exists = .TRUE.
+         END IF
+      END DO
+      IF ( ( exists ) .AND. &
+           ( ( model_config_rec % bl_pbl_physics(1) .EQ. MYNNPBLSCHEME2 ) .OR. &
+             ( model_config_rec % bl_pbl_physics(1) .EQ. MYNNPBLSCHEME3 ) .OR. &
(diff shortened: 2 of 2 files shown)


davegill:
@weiwangncar @dudhia 
This PR relies assumes that nothing bad happens when we have different fields on different domains. That is not always a great assumption.

davegill:
@weiwangncar @dudhia 
Folks,
This simple mod to check_a_mundo makes sure that only PBL schemes that are NOT (MYNN or EEPS) on the coarser grids are permitted to be used with no PBL on a finer domains. This is ready for review for v4.3.2 inclusion.

weiwangncar (dismissed):

davegill:
@dudhia
Jimy,
What do you think about this PR? Can we code it :shipit:

Registry/Registry.EM_COMMON
```diff
@@ -3086,6 +3086,8 @@ package   drip           sf_surf_irr_alloc==2        -             state:irrigat
 package   sprinkler      sf_surf_irr_alloc==3        -             state:irrigation,irr_rand_field
 
 
+package   lesscheme      bl_pbl_physics==0           -             -
+package   nopblscheme    bl_pbl_physics==0           -             -
```
dudhia, line 3090: you have two names for this?

davegill: @dudhia 
Jimy,
I can remove one. Wei asked about if we should call it nopbl or les. We could call it both. What is your preference?

dudhia:
We can leave it as it was (lesscheme?)

Registry/Registry.EM_COMMON
davegill: OK, LESscheme it is!

weiwangncar:
@davegill @dudhia I'm ok with the decision. But I just wanted to say that sometimes a user could turn off PBL for a test, but it doesn't mean LES is used.

davegill:
@weiwangncar @dudhia 
Wei and Jimy,
I put in some small tweaks (getting rid of the name NOPBLSCHEME, edits to the error message, etc). Because of that, you both need to review again! 
Almost there ...

weiwangncar (approved):

dudhia (approved):

twjuliano:
@davegill Just a heads up... I am running some single domain simulations using MYNN and am unintentionally catching this error. I was able to fix the issue by adding `exists = .FALSE.` before looping through domains. Thanks.

PR #1604, opened 2021-12-17
WRFDA memory usage reduction for thin_conv=true

jamiebresch:
TYPE: enhancement

KEYWORDS: WRFDA, memory, thin_conv

SOURCE: Jamie Bresch (NCAR)

DESCRIPTION OF CHANGES:
Problem:
thin_conv is true by default. When ob_format=1 (bufr for conventional ob types), thinning_grid_conv(num_ob_indexes) and its sub data arrays are allocated for all ob types as defined in da_control.f90 (obs_names(num_ob_indexes)) even though not all ob types are used.

Solution:
Do not call make3grids (which allocates several data arrays based on the setting of thin_mesh_conv) for ob types that are not used in the assimilation and not coded for thinning.

Note that gpsref horizontal thinning is not implemented for ob_format=1 (but is implemented for ob_format=2), so its thinning grid does not need to be allocated in da_setup_obs_structures_bufr.inc. That is the reason the specific check for gpsref appears in da_setup_obs_structures_bufr.inc but not da_setup_obs_structures_ascii.inc.

Caveat: The same fix logic is done for thin_conv_ascii (default is .false. and only matters when ob_format=2). But I have never tested thin_conv_ascii=.true.

LIST OF MODIFIED FILES:
M       var/da/da_setup_structures/da_setup_obs_structures.inc
M       var/da/da_setup_structures/da_setup_obs_structures_ascii.inc
M       var/da/da_setup_structures/da_setup_obs_structures_bufr.inc

TESTS CONDUCTED:
1. fixed code has been used extensively by Craig Schwartz.
2. WRFDA regression test?

RELEASE NOTE: WRFDA has been fixed to reduce the memory usage for ob_format=1 (i.e. bufr for conventional obs) and thin_conv=true.

### DIFF
--- a/var/da/da_setup_structures/da_setup_obs_structures.inc
+++ b/var/da/da_setup_structures/da_setup_obs_structures.inc
@@ -169,6 +169,38 @@ subroutine da_setup_obs_structures( grid, ob, iv, j_cost)
       rtminit_sensor   = pseudo_rad_senid
    end if
 
+   obs_use(1:num_ob_indexes) = .false.
+   if ( use_airepobs ) obs_use(airep) = .true.
+   if ( use_bogusobs ) obs_use(bogus) = .true.
+   if ( use_buoyobs ) obs_use(buoy) = .true.
+   if ( use_geoamvobs ) obs_use(geoamv) = .true.
+   if ( use_gpsephobs ) obs_use(gpseph) = .true.
+   if ( use_gpsrefobs ) obs_use(gpsref) = .true.
+   if ( use_gpsztdobs .or. use_gpspwobs ) obs_use(gpspw) = .true.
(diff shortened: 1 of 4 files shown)


liujake:
This looks good. I see thin_conv (and thin_conv_ascii) is a scaler logic variable. Is it possible to change it as an array so that we can turn on/off thinning for individual obs even they are read in?

jamiebresch:
> This looks good. I see thin_conv (and thin_conv_ascii) is a scaler logic variable. Is it possible to change it as an array so that we can turn on/off thinning for individual obs even they are read in?

I can look into the possibility of potential future improvement.

liujake (approved):

davegill (approved):

PR #1611, opened 2021-12-21
Run-time option of GHG concentration from various climate RCPs

smileMchen:
TYPE: improvement

KEYWORDS: radiation, green house gas

SOURCE: internal

DESCRIPTION OF CHANGES: This PR improves the way to handle green house gases in radiation schemes (CAM, RRTMG, RRTMGF, RRTM). Previously green house gases (GHGs) are either specified to constant values or read from historical gas concentartion  datasets (CAMtr). Note that the latter only works when the WRF codes are compiled with the macro flag `-DCLWRFGHG`. This PR modifies radiation schemes so that input of GHG concentration from CAMtr becomes  a run-time option. Meanwhile, the old approach of constant-specification remains unchanged for backward consistency. 
In addition, the GHG concentration is interpolated on a daily basis instead of at each radiation time step. 

LIST OF MODIFIED FILES:
M   Registry/Registry.EM_COMMON
M   dyn_em/module_first_rk_step_part1.F
M   phys/module_ra_cam.F
M   phys/module_ra_cam_support.F
M   phys/module_ra_clWRF_support.F 
M   phys/module_ra_rrtm.F
M   phys/module_ra_rrtmg_lw.F
M   phys/module_ra_rrtmg_lwf.F
## maybe not this one M   phys/module_ra_rrtmg_lwk.F
M   phys/module_ra_rrtmg_sw.F
M   phys/module_ra_rrtmg_swf.F
M   phys/module_radiation_driver.F
M   phys/module_radiation_driver.F

TESTS CONDUCTED: 
(1) Two single cases using CAM and RRTMG are conducted  to make sure the run-time option works as expected. 
(2) pending for Jenkins test

RELEASE NOTE: green house gas concentration input from historical datasets (CAMtr) is a run-time option in the WRFV4.4. The old approach of constant-specification remains unchanged for backward consistency. Radiation schemes of CAM, RRTMG, RRTMGF and RRTM are affected.


davegill:
@smileMchen @weiwangncar 
Folks,
Here is the error that we are getting:
```
time gfortran -o module_ra_farms.o -c  -w -ffree-form -ffree-line-length-none -fconvert=big-endian -frecord-marker=4  -g -O0   -I../dyn_em  -I/wrf/WRF/external/esmf_time_f90  -I/wrf/WRF/main -I/wrf/WRF/external/io_netcdf -I/wrf/WRF/external/io_int -I/wrf/WRF/frame -I/wrf/WRF/share -I/wrf/WRF/phys -I/wrf/WRF/wrftladj -I/wrf/WRF/chem -I/wrf/WRF/inc -I/wrf/libs/netcdf/include    module_ra_farms.f90
module_ra_rrtmg_lwk.f90:9005:7:

 9005 |    use rrlw_kg06
      |       1
Fatal Error: Cannot open module file 'rrlw_kg06.mod' for reading at (1): No such file or directory
compilation terminated.
```

This is a build error in a file that Ming did not change! I need to take a closer look at this. Ming, let me poke around a little bit here.

davegill:
@smileMchen 
Take a look at file module_ra_rrtmg_lwf.F. There are redundant declarations for cfc11 and cc12
```
  15494 !ccc To add time-varying trace gases (CO2, N2O and CH4). Read the conc.  from file
  15495 ! then interpolate to date of run.
  15496       REAL(8), SAVE                                :: clco2, cln2o, clch4, clcfc11, clcfc12
  15497       REAL(8)                                      :: cfc11, cfc12
  15498 
  15499 ! Set trace gas volume mixing ratios, 2005 values, IPCC (2007)
  15500 ! carbon dioxide (379 ppmv) - this is being replaced by an annual function in v4.2
  15501     real :: co2
  15502 !   data co2 / 379.e-6 /
  15503 ! methane (1774 ppbv)
  15504     real :: ch4
  15505     data ch4 / 1774.e-9 /
  15506 ! nitrous oxide (319 ppbv)
  15507     real :: n2o
  15508     data n2o / 319.e-9 /
  15509 ! cfc-11 (251 ppt)
  15510     real :: cfc11
  15511     data cfc11 / 0.251e-9 /
  15512 ! cfc-12 (538 ppt)
  15513     real :: cfc12
  15514     data cfc12 / 0.538e-9 /
```

Same file, `ghg_iput` should probably be `ghg_input`
```
  15325 !ccc Added for time-varying trace gases.
  15326    INTEGER, INTENT(IN    ) ::                             ghg_iput
```

Same file, `ghg_inout` should be `ghg_input`:
```
  15647  IF(ghg_inout.eq.1) THEN
  15648    CALL read_CAMgases(yr,julian,"RRTMG",clco2,cln2o,clch4,clcfc11,clcfc12)
  15649  ENDIF
```

smileMchen:
@davegill 
Dave,
many thanks ! I have corrected the errors you listed and is recouping the codes.

davegill:
@smileMchen @weiwangncar 
MIng,
I'll watch this also. I _think_ that it will still fail after your mods, for that rrlw_kg06 error. That I will need to probably put a fix in the physics dependencies. This rrlw_kg06 has been latent for a while, and crops up occasionally.

smileMchen:
@davegill 
Dave,
I have corrected the codes following your list. WRF is compiled successfully. Please see /glade/scratch/chenming/WRFR (I saved the compiling output in log). By the way, I use GNU.
Strangely, I actually compiled the code this afternoon before commit to repository. I use GNU and got all the exe files. I then run CAM and RRTNG with ghg_input =1 (/glade/scratch/chenming/RT_MET/clcam and clrrtmg). --- in order to understand  what is going on, I will put the errors back to the code and recompile it. I will keep you updated. 
Thanks again.

smileMchen:
@davegill 
Dave,
I intentionally put the error back to phys/module_ra_rrtmg_lwf.F, and recompile the code (/glade/scratch/chenming/WRFR_Err).The code is compiled successfully because at the beginning of the code module_ra_rrtmg_lwf.F, we have the following codes:
```
#if( BUILD_RRTMG_FAST != 1)
      MODULE module_ra_rrtmg_lwf
      CONTAINS
      SUBROUTINE RRTMG_LWRAD_FAST
         REAL :: dummy
         dummy = 1
      END SUBROUTINE RRTMG_LWRAD_FAST
      END MODULE module_ra_rrtmg_lwf
#else
!MODULE module_ra_rrtmg_lwf
```
Since BUILD_RRTMG_FAST != 1, we actually didn't really compile this code.

smileMchen:
@davegill 
Dave,
I didn't have the 'rrlw_kg06 error'. What compiler did you use? I use GNU.

davegill:
@smileMchen @weiwangncar 
I am getting a bit-for-bit comparison diff for the OpenMP comparison tests.

I get a bunch of these prints with serial and MPI:
```
WRF NUMBER OF TILES =   1
d01 2000-01-24_12:00:00  CLWRF reading...: istatus:           0  idata:           1  year:        1765  co2:    278.05200000000002       n2o:    272.95999999999998       ch4:   721.89400000000001
d01 2000-01-24_12:00:00  CLWRF reading...: istatus:           0  idata:           2  year:        1766  co2:    278.10599999999999       n2o:    272.97100000000000       ch4:   722.25599999999997
d01 2000-01-24_12:00:00  CLWRF reading...: istatus:           0  idata:           3  year:        1767  co2:    278.22000000000003       n2o:    272.99400000000003       ch4:   722.97100000000000
```

But for OpenMP, I do not get the prints. Maybe that means these are not used for OpenMP?

davegill:
Here are the jenkins results after the build is OK:
```
Please find result of the WRF regression test cases in the attachment. This build is for Commit ID: 611da94b8ff58256301838009107e7f65d29ed4f, requested by: smileMchen for PR: https://github.com/scala-computing/WRF/pull/1611. For any query please send e-mail to David Gill.

    Number of Simulations  : 159           155         0
    Number of Comparisons  : 96           93         26

    Failed Simulations are: 
    None
    output_15:SUCCESS_RUN_WRF_d01_em_real_32_em_realB_10 vs SUCCESS_RUN_WRF_d01_em_real_33_em_realB_10 status = 1
output_15:SUCCESS_RUN_WRF_d01_em_real_32_em_realB_11 vs SUCCESS_RUN_WRF_d01_em_real_33_em_realB_11 status = 1
output_16:SUCCESS_RUN_WRF_d01_em_real_32_em_realD_38 vs SUCCESS_RUN_WRF_d01_em_real_33_em_realD_38 status = 1
output_16:SUCCESS_RUN_WRF_d01_em_real_32_em_realD_48 vs SUCCESS_RUN_WRF_d01_em_real_33_em_realD_48 status = 1
output_16:SUCCESS_RUN_WRF_d01_em_real_32_em_realD_49 vs SUCCESS_RUN_WRF_d01_em_real_33_em_realD_49 status = 1
output_17:SUCCESS_RUN_WRF_d01_em_real_32_em_realE_52 vs SUCCESS_RUN_WRF_d01_em_real_33_em_realE_52 status = 1
output_17:SUCCESS_RUN_WRF_d01_em_real_32_em_realE_52DF vs SUCCESS_RUN_WRF_d01_em_real_33_em_realE_52DF status = 1
output_18:SUCCESS_RUN_WRF_d01_em_real_32_em_realF_65DF vs SUCCESS_RUN_WRF_d01_em_real_33_em_realF_65DF status = 1
output_21:SUCCESS_RUN_WRF_d01_em_real_32_em_realI_03FD vs SUCCESS_RUN_WRF_d01_em_real_33_em_realI_03FD status = 1
output_21:SUCCESS_RUN_WRF_d01_em_real_32_em_realI_06 vs SUCCESS_RUN_WRF_d01_em_real_33_em_realI_06 status = 1
output_22:SUCCESS_RUN_WRF_d01_em_real_32_em_realJ_50 vs SUCCESS_RUN_WRF_d01_em_real_33_em_realJ_50 status = 1
output_22:SUCCESS_RUN_WRF_d01_em_real_32_em_realJ_51 vs SUCCESS_RUN_WRF_d01_em_real_33_em_realJ_51 status = 1
output_3:SUCCESS_RUN_WRF_d01_em_real_32_em_realA_03 vs SUCCESS_RUN_WRF_d01_em_real_33_em_realA_03 status = 1
output_3:SUCCESS_RUN_WRF_d01_em_real_32_em_realA_03DF vs SUCCESS_RUN_WRF_d01_em_real_33_em_realA_03DF status = 1
output_4:SUCCESS_RUN_WRF_d01_em_real_32_em_realC_20 vs SUCCESS_RUN_WRF_d01_em_real_33_em_realC_20 status = 1
output_5:SUCCESS_RUN_WRF_d01_em_real_32_em_realK_52FD vs SUCCESS_RUN_WRF_d01_em_real_33_em_realK_52FD status = 1
output_5:SUCCESS_RUN_WRF_d01_em_real_32_em_realK_60 vs SUCCESS_RUN_WRF_d01_em_real_33_em_realK_60 status = 1
output_5:SUCCESS_RUN_WRF_d01_em_real_32_em_realK_60NE vs SUCCESS_RUN_WRF_d01_em_real_33_em_realK_60NE status = 1
output_6:SUCCESS_RUN_WRF_d01_em_real_32_em_realL_66FD vs SUCCESS_RUN_WRF_d01_em_real_33_em_realL_66FD status = 1
output_6:SUCCESS_RUN_WRF_d01_em_real_32_em_realL_71 vs SUCCESS_RUN_WRF_d01_em_real_33_em_realL_71 status = 1
output_6:SUCCESS_RUN_WRF_d01_em_real_32_em_realL_78 vs SUCCESS_RUN_WRF_d01_em_real_33_em_realL_78 status = 1
output_6:SUCCESS_RUN_WRF_d01_em_real_32_em_realL_79 vs SUCCESS_RUN_WRF_d01_em_real_33_em_realL_79 status = 1
output_7:SUCCESS_RUN_WRF_d01_em_real_32_em_real_3dtke vs SUCCESS_RUN_WRF_d01_em_real_33_em_real_3dtke status = 1
output_7:SUCCESS_RUN_WRF_d01_em_real_32_em_real_conus vs SUCCESS_RUN_WRF_d01_em_real_33_em_real_conus status = 1
output_7:SUCCESS_RUN_WRF_d01_em_real_32_em_real_rap vs SUCCESS_RUN_WRF_d01_em_real_33_em_real_rap status = 1
output_7:SUCCESS_RUN_WRF_d01_em_real_32_em_real_tropical vs SUCCESS_RUN_WRF_d01_em_real_33_em_real_tropical status = 1
```

phys/module_ra_rrtmg_lw.F
```diff
@@ -11894,18 +11893,25 @@ SUBROUTINE RRTMG_LWRAD(                                        &
 
 !ccc Read time-varying trace gases concentrations and interpolate them to run date.
 !
-#ifdef CLWRFGHG
-
-   CALL read_CAMgases(yr,julian,"RRTMG",co2,n2o,ch4,cfc11,cfc12)
-
+IF(xtime.eq.0 .or. (mod(xtime,1440.).eq.0)) THEN
+ IF(ghg_input.eq.1) then
```
weiwangncar, line 11897: @smileMchen Maybe the above two IFs should be swapped? Chech ghg_input first, before compute mod(xtime,1440.).

phys/module_ra_rrtmg_lw.F
```diff
@@ -11894,18 +11893,25 @@ SUBROUTINE RRTMG_LWRAD(                                        &
...
+   CALL read_CAMgases(yr,julian,"RRTMG",clco2,cln2o,clch4,clcfc11,clcfc12)
+ ENDIF
+ENDIF
+
+ IF(ghg_input.eq.1) then
```
weiwangncar, line 11902: @smileMchen If you swap those above, this block can be part of above IF black. I'd also suggest that since these are not computed once per day, we could have the values printed every time it interpolates.

phys/module_ra_rrtmg_sw.F
```diff
@@ -10356,10 +10355,20 @@ SUBROUTINE RRTMG_SWRAD(                                        &
 
 !-----CALCULATE SHORT WAVE RADIATION
 !                                                              
-#ifdef CLWRFGHG
+IF(xtime.eq.0 .or. (mod(xtime,1440.).eq.0)) THEN
+ IF(ghg_input.eq.1) then
```
weiwangncar, line 10359: @smileMchen Same change suggested above could be applied here. But no need to print these values twice. LW is called first, so let it print.

weiwangncar:
@davegill Would code like 'IF(xtime.eq.0 .or. (mod(xtime,1440.).eq.0)) THEN' inside radiation subroutine break OpenMP?

phys/module_ra_rrtmg_lw.F
smileMchen: @weiwangncar 
Wei,
I don't feel like to swap the two IFs. This is because logically, we want to activate ghg_input =1  (1) only at the beginning of the model run, and (2) only on a daily basis.

phys/module_ra_rrtmg_lw.F
smileMchen: @weiwangncar 
Wei,
They are computed once per day. This is why I add the extra IF block.

davegill:
@weiwangncar 
> Would code like 'IF(xtime.eq.0 .or. (mod(xtime,1440.).eq.0)) THEN' inside radiation subroutine break OpenMP?

I don't see how _this_ could be a problem, but clearly, there is an OpenMP issue.

weiwangncar:
@davegill Maybe the variables need to be initialized.

davegill:
@weiwangncar @smileMchen 
Folks,
I am looking at the problem now.

davegill:
@smileMchen @weiwangncar 
Folks,
It is definitely an OpenMP issue.  
I did a single domain (30 km), one time step simulation.

Here are the different fields:
```
U,V,W,PH,T,THM,MU,P,Q2,T2,TH2,QVAPOR,QCLOUD,QRAIN,TSLB,TSK,RAINC,RAINNC,SNOWNC,GLW,ACLWUPT,ACLWUPTC,ACLWUPB,ACLWUPBC,ACLWDNB,ACLWDNBC,LWUPT,LWUPTC,LWUPB,LWUPBC,LWDNB,LWDNBC,OLR,HFX,QFX,LH,ACHFX,ACLHF
```
Notice this is basically radiation + everything that radiation would change in a time step. 

Here is a difference plot of the output after the first time step (where difference is openmp - serial). Anywhere that the color is blue means NO DIFFS.

Screen Shot 2021-12-21 at 2 44 13 PM

I used 4 OpenMP threads. The figure shows that the top rectangle (one of the four openmp threads) has identical results.

On cheyenne, I used GNU, configure -d, and used options 32 (serial) and 33 (openmp).

davegill:
@smileMchen @weiwangncar 
Folks,
One of the ways that this particular difference shows up is if there are initializations that DO NOT use the tile sizes. So, assume that A and B are memory-sized 2d or 3d arrays. If we say
```
A=0.0
B=0.0
```
Which ever OpenMP thread hits that command last, well that thread is the winner. All of the other threads have fields (such as A and B) that are now overwritten.

smileMchen:
I wonder whether the double and single decision will affect the OpenMP run? Below is an example in phys/module_ra_rrtmg_lwf.F:

The original code is:
```
#ifdef CLWRFGHG
! CLWRF-UC June.09
      REAL(8)                                      :: co2, n2o, ch4, cfc11, cfc12
#else
! Set trace gas volume mixing ratios, 2005 values, IPCC (2007)
! carbon dioxide (379 ppmv) - this is being replaced by an annual function in v4.2
    real :: co2
!   data co2 / 379.e-6 /
! methane (1774 ppbv)
    real :: ch4
    data ch4 / 1774.e-9 /
    ...
    ...
#ifdef CLWRFGHG

   CALL read_CAMgases(yr,julian,"RRTMG",co2,n2o,ch4,cfc11,cfc12)

```

The modified code is:
 ```
      REAL(8), SAVE                                :: clco2, cln2o, clch4, clcfc11, clcfc12
! Set trace gas volume mixing ratios, 2005 values, IPCC (2007)
! carbon dioxide (379 ppmv) - this is being replaced by an annual function in v4.2
    real :: co2
!   data co2 / 379.e-6 /
! methane (1774 ppbv)
    real :: ch4
    data ch4 / 1774.e-9 /
   ....
   ....
    IF(ghg_input.eq.1) THEN
   CALL read_CAMgases(yr,julian,"RRTMG",clco2,cln2o,clch4,clcfc11,clcfc12)
   ...
      co2=clco2
      n2o=cln2o
      ch4=clch4
      cfc11=clcfc11
      cfc12=clcfc12
```

In original code, with the macro "CLWRFGHG", co2, ch4, etc are real(8)
In thee modified code, these variables are real(4).

Could this be a reason for the OpenMP issue?

**Comment by smileMchen** [comment]
I just double check the codes before and after the changes. Other than the single and double precision issue, I cannot find any other reason why OpenMP and serial is not bit-for-bit.
Will continue to look at this issue.

**Comment by smileMchen** [comment]
Those GHG concentrations are not initialized when ghg_input =1 (or WLWRFGHG is activated in the original codes). They gain values by  reading from CAMtr datasets, and the reading is conducted at every radiation time step.

Dave mentioned that "if there are initializations that DO NOT use the tile sizes", we may have trouble with OpenMP. Does the current way we read CAMtr dataset use tile sizes?

**Comment by davegill** [comment]
@smileMchen @weiwangncar 
> Does the current way we read CAMtr dataset use tile sizes?

Can you point me to that part of the code?

**Comment by weiwangncar** [comment]
@davegill This would be around line 11898 in module_ra_rrtmg_lw.F. But the issue is with code that is not using the data from the text file, is that right?

**Comment by smileMchen** [comment]
@davegill @weiwangncar 
I manually compiled the code in serial and smpar mode, using gnu compiler. Then run a test case with ghg_input = 1. It is found that results are bit identical when using cam radiation.  However, the results are different when using RRTMG radiation. I will continue to look at this issue and keep you updated.

**Comment by smileMchen** [comment]
@davegill @weiwangncar 

Dave and Wei,
I have modified the codes and run a test case in serial and openMP modes. The codes are saved at /glade/scratch/chenming/WRFR/. If you run git status and git diff, you can find the changes I made after my last PR. 

 Below is the plots that show difference between OpenMP runs and serial run.For the openMP run, I  use 2 and 4 OpenMP threads.  The issue still exists. I am not sure yet what could be the possible reason. I will continue to work on this issue. At the same time, please let me know if you have any suggestions. 
test2
test4

**Comment by smileMchen** [comment]
This is the difference of LWDNB between serial and openMP run. OpenMP is run with 4 threads. Now we only see difference along the blue lines (j=, which is the place where the domain is divided.

**Comment by smileMchen** [comment]
This is the plot of LWDNB difference between serial and openMP runs. OpenMP is run with 4 threads. Note we only see differences along the blue lines, which correspond to j=1, 83, 164, along which the domain is divided. Also, the difference only appears for i=1, 16. Results at all other grids are identical.   

test1

**Comment by davegill** [comment]
@smileMchen @weiwangncar 
Ming,
I understand that this is development. Would you be able to push the commits to github that show the code diffs that changed the results from "half the domain is different" to "only the blue lines are different".

**Comment by smileMchen** [comment]
@davegill 
Dave,
I add many "print" to various places in the codes. I will clean up these 'print' and then submit the code to git. I will let you know after I am done. 
Ming

**Comment by smileMchen** [comment]
@davegill 
Dave,
I just pushed the new changes to the branch GHG. The issue is only related to LW. I have the test case saved at 
/glade/scratch/chenming/RT_MET/clrrtmg, where you can find 
wrfout_d01_2017-02-23_00:15:00.serial,  wrfout_d01_2017-02-23_00:15:00.dmpar  and their difference  diff.nc. 
This is a rrtmg case. 
Looking at LWDNB, again I can see difference along the dividing line between subregions.  Note the difference only occurs at two points along the western boundary (i=1, 2). 
test4

**Comment by smileMchen** [comment]
@davegill 
Dave,
In my previous test with a larger domain, the difference starts from the west boundary and over the grids i=1, 16. For the test I am doing now over a smaller domain, the difference starts from the west boundary and only over grids i=1,2. 
How to interpret this phenomenon?

File: phys/module_ra_rrtmg_lw.F [hunk 4]
```diff
@@ -11740,15 +11742,14 @@ SUBROUTINE RRTMG_LWRAD(                                        &
 
 !ccc To add time-varying trace gases (CO2, N2O and CH4). Read the conc.  from file
 ! then interpolate to date of run.
-#ifdef CLWRFGHG
 ! CLWRF-UC June.09
-      REAL(8)                                      :: co2, n2o, ch4, cfc11, cfc12
-#else
+      REAL(8), SAVE             :: clco2, cln2o, clch4, clcfc11, clcfc12
```
weiwangncar, line 11746: @smileMchen @davegill Compared to the original GHG code, maybe SAVE should not be used here?

smileMchen: @weiwangncar 
Wei,
I have deleted all the 'save' statement in the codes. I believe this could at least  be one reason for the bit-identical failure. I am compiling the code manually now. Once it is done, I will push to GitHub.

davegill:
@smileMchen @weiwangncar 

Screen Shot 2022-01-04 at 7 06 01 PM

The mods that I put in for d3c5e6d work: all of the Jenkins tests are passing by turning off the GHG input option. Without this mod, all of the OpenMP b4b tests are failing.  

The subsequent mods break the PR.

jamiebresch:
Here is my unsolicited layperson thinking.

co2vmr, n2ovmr, ch4vmr, f11vmr, and f12vmr need to be declared in a module and then `use` by each radiation modules.
In the module that declares those vmr, there exist a `init_ghg` subroutine that set default values currently used in each radiation scheme. Also in `init_ghg`, if (ghg_input==1) then call read_CAMgases.
`init_ghg` is called once in the beginning of the model integration to initialize those vmr.
During model integration, call read_CAMgases to update vmr when desired.

In the current implementation, when vmr is a local variable, read_CAMgases has to be called every time for ghg_input=1.

davegill:
This is replaced by "Run-time option for climate GHG for radiation" #1625

PR #1642, opened 2022-01-16
Bug fix for radar_rf_opt=2

liujake:
Bug fix for direct assimilation of radar reflectivity (radar_rf_opt=2)

TYPE: bug fix

KEYWORDS: reflectivity operator, quality control 

SOURCE: Zhiquan (Jake) Liu (NCAR/MMM)

DESCRIPTION OF CHANGES:
Problem:
The namelist parameter "qthres", which is used for hydrometeor base state of TL/AD for no-rain background, should not be used in reflectivity forward operator. QC for radar_rf_opt=2 should not be confused with that of retrieval radar DA.

Solution:
1. Remove the use of "qthres" in forward operator.
3. Simplify QC to keep obs with both observed and background reflectivity >=rfmin (a namelist parameter).

LIST OF MODIFIED FILES:
M       var/da/da_radar/da_get_innov_vector_radar.inc
M       var/da/da_radar/da_radzicevar.inc
M       var/da/da_radar/da_radzicevar_upper_f.inc
M       var/da/da_radar/da_transform_xtoy_radar.inc
M       var/da/da_radar/da_transform_xtoy_radar_adj.inc
M       var/da/da_setup_structures/da_setup_obs_structures_radar.inc

TESTS CONDUCTED: 
1. WFRDA Regtest passed.
2. Qualified yes. The parallel netcdf4 mods introduced an error in DA that was not caught.

RELEASE NOTE: Bug fix for direct assimilation of radar reflectivity.

### DIFF
--- a/var/da/da_radar/da_get_innov_vector_radar.inc
+++ b/var/da/da_radar/da_get_innov_vector_radar.inc
@@ -96,7 +96,7 @@ subroutine da_get_innov_vector_radar (it, grid, ob, iv)
    qng=0
    rn0_r=8e6
    rn0_s=3e6
-   rn0_g=4e5
+   rn0_g=4e6
    rhos=100.0
    rhog=400.0
    !------------------------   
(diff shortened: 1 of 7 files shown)


junmeiban (approved):

thomasuhoh:
According to the publication of Jung et al. (2008),  the parameter rn0_g should be _4.e4_ instead of _4.e6_.

PR #1662, opened 2022-01-28
BF firebrand array index must be integer

davegill:
TYPE: bug fix

KEYWORDS: firebrand, Cray, HPE, cat

SOURCE: Patricia Balle (HPE), internal

DESCRIPTION OF CHANGES:
Problem:
The line
```
get_fuel_cat = ksb(cat)
```
should have an integer as the index to the `ksb` array.

Solution:
At the top of the function is the line
```
icat = INT(cat)
```
which is supposed to be used as the argument.

LIST OF MODIFIED FILES:
modified:   phys/module_firebrand_spotting.F

TESTS CONDUCTED:
1. The fix allows the HPE CCE compilation to proceed.
2. All jenkins tests are PASS.

### DIFF
--- a/phys/module_firebrand_spotting.F
+++ b/phys/module_firebrand_spotting.F
@@ -570,7 +570,7 @@ END FUNCTION order_val
 
         ENDIF
 
-        get_fuel_cat = ksb(cat)
+        get_fuel_cat = ksb(icat)
 
     END FUNCTION get_fuel_cat
 !=============================================================


davegill:
@mefrediani @pkb99 @weiwangncar @dudhia 
Tricia,
Thanks for finding and reporting this bug.

Wei and Jimy,
Would you review please. This is a pretty simple one character change.

Also, Tricia may have a new Cray stanza to contribute, so that might be coming down the pike before the release.

phys/module_firebrand_spotting.F
```diff
@@ -570,7 +570,7 @@ ELEMENTAL FUNCTION get_fuel_cat(crosswalk, cat)
 
         ENDIF
 
-        get_fuel_cat = ksb(cat)
+        get_fuel_cat = ksb(icat)
```
davegill, line 573: Look back at line 538, that is where `icat` is defined.

mefrediani (approved):
Thanks for catching this!

davegill:
@weiwangncar @kkeene44 @smileMchen 
Folks,
This is a really simple PR to review. The problem was a using a non-integer as an index. The integer value is available locally. The developer, Maria, said the the change is OK.

weiwangncar (approved):

kkeene44 (approved):

PR #1685, opened 2022-02-19
Changes to derive snow height from snow water equivalent

thomasuhoh:
Changing the conversion from snow water equivalent to physical snow depth using ECMWF data

TYPE: enhancement 

KEYWORDS: snow water equivalent, snow height, real.exe

DESCRIPTION OF CHANGES:
Problem: When calculating physical snow height from ECMWF snow water equivalent, very high values of 50 m occur over glaciers (e.g, over Iceland or the Alps).

Solution:
Changed the conversion from snow water equivalent to snow height following Hill et al., (2019) (https://tc.copernicus.org/articles/13/1767/2019/; eq. 2) when only snow water equivalent is present. This reduces the maximum snow height to approx. 24 m and the other values by about a 40 %.

LIST OF MODIFIED FILES: 

TESTS CONDUCTED: 

RELEASE NOTE: Changed the conversion from snow water equivalent to physical snow depth in case only snow water equivalent is available.

### DIFF
--- a/dyn_em/module_initialize_real.F
+++ b/dyn_em/module_initialize_real.F
@@ -538,9 +538,9 @@ integer::oops1,oops2
       ELSE IF ( ( flag_snow .EQ. 1 ) .AND. ( flag_snowh .EQ. 0 ) ) THEN
          DO j=jts,MIN(jde-1,jte)
             DO i=its,MIN(ide-1,ite)
-!              ( kg/m^2 -> m)  & ( liquid to snow depth, 5:1 ratio )
                IF ( skip_middle_points_t ( ids , ide , jds , jde , i , j , em_width , hold_ups ) ) CYCLE
-               grid%snowh(i,j) = grid%snow(i,j) / 1000. * 5.
+               ! Hill et al 2019, Cryosphere         
+               grid%snowh(i,j) = (10.**(((log10(grid%snow(i,j))-log10(0.146))/1.102)))/1000.
             END DO
          END DO
 


weiwangncar:
@thomasuhoh Do you know if this method also works for GFS data?

weiwangncar:
Jenkins tests have passed:

    Failed Simulations are: 
    None
    None

jimbresch:
This change should not be made in real. The snow/snowh conversions are handled in WPS, specifically in ungrib/src/rrpr.F lines 882 to 908. Different models use different ratios and only ungrib knows the source model information. Ungrib supplies a snowh to real that assumes a 200:1 ratio which is what WRF (and NOAH) expects. The ECMWF calculation can be changed in rrpr.F if necessary. If ECMWF changed the way they compute snowh, then we will need the date of the change so we can handle the pre- and post-change input files correctly.

thomasuhoh:
Jim, 
Thank you for pointing this out. Seems that I overlooked this. ECMWF also provides snow density (grib code 33 in their table) so that it's necessary to add this field to Vtable.ECMWF and Vtable.ECMWF_sigma as "SNOW_DEN" to obtain a correct calculation for SNOWH.
I tried this rather quickly, and it seems that "SNOWH" needs to be included in the Vtable.ECMWF in addition as a derived field.

dudhia:
I would add that the liquid water equivalent is the key variable for the
LSM input.
It probably immediately computes its own SNOWH in the first call to it, so
you might
see it quickly change regardless of input value.
Jimy

thomasuhoh:
If I see it correctly, NOAH+NOAHMP uses both liquid water equivalent and SNOWH as input as the logical "FNDSNOWH" is set to true in module_physics_init.F

dudhia:
@cenlinhe this has been in our queue for 2 years. Can we remove it, or is it something to consider? It is only making a change to real.exe to the snowh calculation. I am not sure if it is a better calculation than the one there already.

cenlinhe:
I would prefer this is done in WPS, since this only applies for ECMWF data and should not be added as a general solution in real.exe for all datasets. So probably we could remove it.

dudhia:
Closed noting comment from @cenlinhe

PR #1689, opened 2022-02-23
Urban model: fix undefined WDR and direction logic

dudhia:
TYPE: bug fix

KEYWORDS: urban model WDR undefined before use
SOURCE: internal 
DESCRIPTION OF CHANGES:
Problem:
WDR was used before being defined in urban model. This is a wind direction used in computing the urban roughness length for local surface flux calculations.
Another problem was that logic for some directions was wrong in selecting the correct one of 4 frontal area directions.

Solution:
Uncomment WDR calculation (assumes grid is roughly y direction northwards).
Correct direction if-tests.

ISSUE: None
LIST OF MODIFIED FILES: 
M       phys/module_sf_urban.F

TESTS CONDUCTED: 
1. Compile only - recommend test for effect

RELEASE NOTE: minor fix for urban wind directions when using detailed urban morphology maps

### DIFF
--- a/phys/module_sf_urban.F
+++ b/phys/module_sf_urban.F
@@ -686,7 +686,7 @@ use module_wrf_error
    SIGMA_ZED = stdh_urb
 
  !Calculate Wind Direction and Assign Appropriae lf_urb
-   !WDR = (180.0/PI)*ATAN2(U10,V10)
+   WDR = (180.0/PI)*ATAN2(U10,V10)
    
    IF(WDR.ge.0.0.and.WDR.lt.22.5)THEN
      lambda_f = lf_urb(1)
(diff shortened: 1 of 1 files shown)


phys/module_sf_urban.F
```diff
@@ -686,7 +686,7 @@ SUBROUTINE urban(LSOLAR,                                           & ! L
    SIGMA_ZED = stdh_urb
 
  !Calculate Wind Direction and Assign Appropriae lf_urb
-   !WDR = (180.0/PI)*ATAN2(U10,V10)
+   WDR = (180.0/PI)*ATAN2(U10,V10)
    
    IF(WDR.ge.0.0.and.WDR.lt.22.5)THEN
      lambda_f = lf_urb(1)
```
weiwangncar, line 692: @dudhia Is lf_urb() defined coming into this subroutine?

weiwangncar (approved):

kkeene44 (approved):

PR #1567, opened 2021-10-14
Removing legacy code and cleaning up logic for real processing of aerosol for mp_physics=28

twjuliano:
TYPE: enhancement

KEYWORDS: real, Thompson aerosol-aware, microphysics

SOURCE: Timothy W. Juliano and Pedro A. Jimenez (NCAR/RAL)

DESCRIPTION OF CHANGES:
Problem:
Several modifications have been made to the Thompson Aerosol-Aware scheme (mp_physics=28) in an effort to clean the logic in dyn_em/module_initialize_real and enable more flexibility when pre-processing first guess aerosol fields (e.g., GEOS-5). These modifications are made in preparation for upcoming enhancements to the Thompson AA scheme related to the addition of black carbon aerosol (anthropogenic and biomass burning) and organic carbon biomass burning aerosol, as well as time-varying surface emissions.

Solution:
1. Processing of aerosol in real.exe contained legacy code. Specifically, the monthly qnwfa and qnifa variables (qn?fa_jan, qn?fa_feb, etc.) have been removed from Registry files and dyn_em/module_initialize_real.

2. Modifications have been made to the METGRID.TBL file to simplify the subsequent logic in the real program's dyn_em/module_initialize_real. Specifically, we: 
   * replace FLAG_QNWFA and FLAG_QNIFA with FLAG_QNWFA_CL and FLAG_QNIFA_CL, respectively, in the W_WIF_DEC and I_WIF_DEC entries, where “CL” stands for climatology; 
   * add FLAG_QNWFA and FLAG_QNIFA to the QNWFA and QNIFA entries, respectively, for the first guess aerosol arrays;
   * add entry P_WIF to give the user the option to interpolate the first guess aerosol using the native pressure levels of the host model;
   * use num_wif_levels for any z dimension associated with the aerosol field entries (climatology or first guess);
   * add surface aerosol emission arrays QNWFA2D and QNIFA2D to be processed.

The modifications to METGRID.TBL are as follows: 
```
========================================
name=W_WIF_DEC
z_dim_name=num_wif_levels
interp_option=four_pt+average_4pt
flag_in_output=FLAG_QNWFA_CL
========================================
name=I_WIF_DEC
z_dim_name=num_wif_levels
interp_option=four_pt+average_4pt
flag_in_output=FLAG_QNIFA_CL
========================================
name=QNWFA
z_dim_name=num_wif_levels
interp_option=four_pt+average_4pt
flag_in_output=FLAG_QNWFA
========================================
name=QNIFA
[...]

PR #1722, opened 2022-04-20
Reduce memory increase from PRs 1567 and 1616

weiwangncar:
Reduce memory used by the model from two previous PRs

TYPE: bug fix

KEYWORDS: qnbca, wif_input_opt, package

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
PR1567 changed default wif_input_opt from 0 to 1, which increased memory use by about 12%. PR1616 added new variable qnbca, but it is not packaged.

Solution:
Change default wif_input_opt option back to 0, and add qnbca to package for mp_physics=28 and its dfi counterpart. Testing the latest release 4.4 code shows a memory reduction (when mp_physics=28 is not used) is down by 11%.

LIST OF MODIFIED FILES: 
M    Registry/registry.new3d_wif

TESTS CONDUCTED: 
1. Do mods fix problem? The fix reduces memory use in the model.

RELEASE NOTE:

### DIFF
--- a/Registry/Registry.EM_COMMON
+++ b/Registry/Registry.EM_COMMON
@@ -2988,7 +2988,7 @@ package   nssl_1momlfo    mp_physics==21               -             moist:qv,qc
 package   nssl_2momg      mp_physics==22               -             moist:qv,qc,qr,qi,qs,qg;scalar:qndrop,qnr,qni,qns,qng,qvolg;state:re_cloud,re_ice,re_snow
 package   wsm7scheme      mp_physics==24               -             moist:qv,qc,qr,qi,qs,qg,qh;state:re_cloud,re_ice,re_snow
 package   wdm7scheme      mp_physics==26               -             moist:qv,qc,qr,qi,qs,qg,qh;scalar:qnn,qnc,qnr;state:re_cloud,re_ice,re_snow
-package   thompsonaero    mp_physics==28               -             moist:qv,qc,qr,qi,qs,qg;scalar:qni,qnr,qnc,qnwfa,qnifa;state:re_cloud,re_ice,re_snow,qnwfa2d,qnifa2d,taod5503d,taod5502d
+package   thompsonaero    mp_physics==28               -             moist:qv,qc,qr,qi,qs,qg;scalar:qni,qnr,qnc,qnwfa,qnifa,qnbca;state:re_cloud,re_ice,re_snow,qnwfa2d,qnifa2d,taod5503d,taod5502d
 package   p3_1category    mp_physics==50               -             moist:qv,qc,qr,qi;scalar:qni,qnr,qir,qib;state:re_cloud,re_ice,vmi3d,rhopo3d,di3d,refl_10cm,th_old,qv_old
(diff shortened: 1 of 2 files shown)


dudhia:
Need to update PR message if this is the final version.

weiwangncar:
The Jenkins tests have passed (with adding wif_input_opt = 1 to two of the test namelists):

    Failed Simulations are: 
    None
    None

weiwangncar:
@dudhia @kkeene44 @twjuliano This PR is ready for review.

twjuliano:
@weiwangncar Thanks for looking into this memory issue. The only concern I have is that we added the BC aerosols to be included in mp=28 only when wif_input_opt=2 (we kept wif_input_opt=1 the same so that a user can run a simulation without BC). Thus, I added qnbca (but not dfi_qnbca) in the `use_wif_input_bc` package in the new3d_wif registry. Just want to make sure that it is OK to have qnbca included in the general Thompson package even when wif_input_opt=1 (i.e., BC not included).

dudhia:
Was #1567 after the last release or will this change the behavior of a release?

dudhia (approved):

weiwangncar:
@dudhia PR1567 is for v4.4. Is that what you're asking?

dudhia:
Was it after the last minor release?

weiwangncar:
@twjuliano I don't think this matters to your concern. Without this PR, qnbca array is available ALL the time - that is why it increases the memory used by the model whether you are using mp_physics = 28 or not. With this change, the array is made available only to this microphysics option. Does it make sense? It wouldn't hurt if you could run a test to verify that.

dudhia:
I notice that the default is still wif_input_option = 1 in the registry,
and also the description there needs updating to include the 2 option.
My view is that it is OK to leave this in the Thompson package even if not
used.

weiwangncar:
@dudhia Yes. It has been on the develop branch all this time, until the cut-over to release-v4.4.

weiwangncar:
@dudhia That will change with this PR - this PR sets wif_input_opt = 0.

dudhia:
OK, need to make sure that if a user forgets to set that to 1 or 2, there is a reasonable message from real and probably a stop for option 28.

weiwangncar:
@dudhia The default value for wif_input_opt is 0 in previous releases.

twjuliano:
> @twjuliano I don't think this matters to your concern. Without this PR, qnbca array is available ALL the time - that is why it increases the memory used by the model whether you are using mp_physics = 28 or not. With this change, the array is made available only to this microphysics option. Does it make sense? It wouldn't hurt if you could run a test to verify that.

@weiwangncar I see, that makes sense, thanks. I can run the test later today if that isn't too late.

weiwangncar:
@twjuliano No, it's not too late. I will wait for your test results.

twjuliano:
@weiwangncar I just ran a test: this works as intended, thanks!

weiwangncar:
@twjuliano Thanks for running the test!
@kkeene44 Can you review this and if it is ok, approve it?

kkeene44 (approved):

PR #1575, opened 2021-11-02
WRFDA for racm_soa_vbs_da (chem_opt=108) in the WRF-Chem model

syha:
TYPE: new features

KEYWORDS: aerosol data assimilation, RACM, MADE-VBS aerosols, WRF-Chem

SOURCE: Soyoung Ha (MMM/NCAR)

DESCRIPTION OF CHANGES: 
WRFDA is newly expanded for chem_cv_options=108 for surface chemical data assimilation (PM2.5, PM10, CO, SO2, NO2, O3) using the RACM/MADE-VBS chemistry option in WRF-Chem.

LIST OF MODIFIED FILES: 
M       Registry/registry.var_chem
M       var/da/da_chem_sfc/da_ao_stats_chem_sfc.inc
M       var/da/da_chem_sfc/da_calculate_grady_chem_sfc.inc
M       var/da/da_chem_sfc/da_chem_sfc.f90
M       var/da/da_chem_sfc/da_get_innov_vector_chem_sfc.inc
M       var/da/da_chem_sfc/da_residual_chem_sfc.inc
M       var/da/da_chem_sfc/da_transform_xtoy_chem_sfc.inc
M       var/da/da_chem_sfc/da_transform_xtoy_chem_sfc_adj.inc
M       var/da/da_define_structures/da_allocate_observations_chem_sfc.inc
M       var/da/da_define_structures/da_deallocate_background_errors.inc
M       var/da/da_define_structures/da_define_structures.f90
M       var/da/da_main/da_update_firstguess.inc
M       var/da/da_obs_io/da_final_write_obs_gas_sfc.inc
M       var/da/da_obs_io/da_read_obs_chem_sfc.inc
M       var/da/da_obs_io/da_read_omb_tmp.inc
M       var/da/da_obs_io/da_scan_obs_chem_sfc.inc
M       var/da/da_obs_io/da_write_obs_chem_sfc.inc
M       var/da/da_setup_structures/da_get_vertical_truncation.inc
M       var/da/da_setup_structures/da_setup_background_errors.inc
M       var/da/da_setup_structures/da_setup_be_regional.inc
M       var/da/da_setup_structures/da_setup_firstguess_wrf.inc
M       var/da/da_setup_structures/da_setup_obs_structures.inc
M       var/da/da_setup_structures/da_setup_obs_structures_chem_sfc.inc
M       var/da/da_statistics/da_analysis_stats.inc
M       var/da/da_statistics/da_print_qcstat.inc
[...]

PR #1724, opened 2022-04-24
Fix quality control issue of chemical DA

liujake:
TYPE: bug fix

KEYWORDS: Quality control, chemical DA

SOURCE: Jake Liu (NCAR/MMM)

DESCRIPTION OF CHANGES:
Problem:
PR #1575 introduced a new chemical DA option for the MADE-VBS aerosol scheme (chem_cv_options=108).
Ideally this should not change the observation part of the chemical DA code. However, current implementation also modified observation part along with the new option (e.g., obs file format, obs units, quality control).  The New QC check for rejecting obs with omb>maxomb leads to too many obs rejected for the obs file format for other chemical DA options originally implemented in v4.3. 

Solution:
Restrict the new maxomb QC check only for chem_cv_options=108. Also limit a print statement only for chem_cv_options=108. This does not recover exactly the v43 QC for other options, but it is close.

LIST OF MODIFIED FILES:
M       var/da/da_chem_sfc/da_get_innov_vector_chem_sfc.inc
M       var/da/da_obs_io/da_read_obs_chem_sfc.inc

TESTS CONDUCTED: 
The statistic_chem file now has the similar number of good obs for chem_cv_options=20 (case provided in the user's guide) with the fixed code. 

RELEASE NOTE: None.

### DIFF
--- a/var/da/da_chem_sfc/da_get_innov_vector_chem_sfc.inc
+++ b/var/da/da_chem_sfc/da_get_innov_vector_chem_sfc.inc
@@ -183,11 +183,13 @@ subroutine da_get_innov_vector_chem_sfc( it, num_qcstat_conv, grid, ob, iv)
 
             ! Quality Control
             ! ----------------
+           if (chem_cv_options >= 108) then
             if (abs(iv % chemic_surf(n) % chem(ichem) % inv) .ge. maxomb(ipm)) then
                 iv % chemic_surf(n) % chem(ichem) % inv   = missing_r
                 iv % chemic_surf(n) % chem(ichem) % qc    = missing
                 iv % chemic_surf(n) % chem(ichem) % error = missing_r
             end if
+           end if
 
          end if    ! good obs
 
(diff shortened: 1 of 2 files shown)


weiwangncar:
The Jenkins tests have passed:

    Failed Simulations are: 
    None
    None

syha:
For clarification -
1) the observation part of the chemical DA code was highly customized for specific dataset, which did not work for my data from different resources, so I had to make changes. At some point, it would be nice to provide a preprocessing tool to generalize the input chem data format. For now, we need to go with the if-statement for different chem_opt because the test data for each option comes with different format and units.
2) DA algorithms do need QCs based on omb to minimize the shock to the model. This is a reliability issue, and has nothing to do with chemical options. So we should not limit the QCs to a specific chem_opt, as suggested.
Instead, I propose to relax maxomb values to (/300, 500, 2, 2, 2, 2/) in da_get_innov_vector_chem_sfc.inc. (Feel free to make them larger, if you want.) 
As we are about to release the new version, we cannot make any major changes now, but for the next release, we can certainly put them in the namelist to let users adjust the values as they want. But we should keep this capability regardless of chem_opt.
In a nutshell, I suggest you revert all the changes you made here.

liujake:
@syha will make another PR soon for this. closing this one.

PR #1742, opened 2022-06-16
configure: check nc-config and nf-config

wkliao:
Fix the use of nc-config and nf-config.

TYPE: bug fix

KEYWORDS: nc-config, nf-config

SOURCE: Wei-keng Liao

DESCRIPTION OF CHANGES:
Problem: Sometimes $NETCDF/bin/nc-config  is used and others nc-config. The same for nf-config. Error messages below appear when the folder of `nc-config` and `nf-config` is not in the user's env PATH.
```
configure: line 163: nc-config: command not found
configure: line 174: nc-config: command not found
expr: syntax error: unexpected argument ‘100’
expr: syntax error: unexpected argument ‘10’
expr: syntax error: missing argument after ‘+’
expr: syntax error: missing argument after ‘+’
configure: line 182: [: -lt: unary operator expected
...
NetCDF version: 
configure: line 917: nc-config: command not found
Enabled NetCDF-4/HDF-5: 
configure: line 918: nc-config: command not found
```

Solution:
* Use 2 variables nc_config and nf_config
* Requires both NetCDF C and Fortran libraries installed in the same location set in env variable NETCDF.
* Add a check for whether NetCDF is built with parallel I/O enabled by calling `nc-config --has-parallel4`
* Improve error messages

LIST OF MODIFIED FILES: configure

TESTS CONDUCTED: 
* NETCDF env variable is set, but its bin folder does not appear in env PATH

### DIFF
--- a/configure
+++ b/configure
@@ -220,29 +220,28 @@ export NETCDF
 export NETCDF4
 export USENETCDFPAR
 
+# Check NetCDF C libaray files
+if [ -f "$NETCDF/lib/libnetcdf.a" ] || [ -f "$NETCDF/lib/libnetcdf.so" ] || [ -f "$NETCDF/lib/libnetcdf.dll.a" ] ; then
+    USENETCDF="-lnetcdf"
+else
+    USENETCDF=""
+fi
+export USENETCDF=$USENETCDF
+
+# Check NetCDF Fortran library files, assuming both C and Fortran libraries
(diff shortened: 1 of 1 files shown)


dudhia (commented):
looks substantial, someone needs to check for approval

wkliao:
See #1923

PR #1755, opened 2022-08-09
Bug fix for ozone input option o3input = 0

weiwangncar:
Bug fix for o3input=0

TYPE: bug fix

KEYWORDS: RRTMG radiation, o3input=0

SOURCE: reported by Mauro Morichetti of Institute of Atmospheric Sciences and Climate (ISAC), ITALY, fixed internally

DESCRIPTION OF CHANGES:
Problem:
The ozone input option 0 (o3input=0) has been broken since 3.5.1 after a different ozone input option was introduced in 3.5. The code checks the presence of the ozone array O33D, and if it is not present, then the ozone is computed using o3input option 0. But the array is always present, so that code never reaches the other branch. Note also that the default o3input option was changed to 2 since 3.7.

Solution:
This PR fixes the option by checking the ozone input option rather than the presence of the array O33D. As there is an interest to see the output from this option, code is added to pass out the data, which is based on 1-d ozone profile. 

LIST OF MODIFIED FILES: 
M       phys/module_ra_rrtmg_lw.F
M       phys/module_ra_rrtmg_lwf.F
M       phys/module_ra_rrtmg_swf.F

TESTS CONDUCTED: 
1. Do mods fix problem? Tested and it fixed the bug.
2. The Jenkins tests have passed.

RELEASE NOTE: Fixed a bug associated with ozone input option o3input = 0. When o3input is set to 0, ozone is effectively not used in the RRTMG radiation option before this fix.

### DIFF
--- a/phys/module_ra_rrtmg_lw.F
+++ b/phys/module_ra_rrtmg_lw.F
@@ -12412,7 +12412,6 @@ CONTAINS
             endif
          enddo
         else
-         print *, 'are we here in lw rad'
          do k = kts, nlayers
             o3vmr(ncol,k) = o3mmr(k) * amdo
             o31d(k) = o3vmr(ncol,k)
--- a/phys/module_ra_rrtmg_lw.F
+++ b/phys/module_ra_rrtmg_lw.F
@@ -11701,9 +11701,8 @@ CONTAINS
 
 !  Ozone
    REAL, DIMENSION( ims:ime, kms:kme, jms:jme )                 , &
-         OPTIONAL                                               , &
-         INTENT(IN   ) :: O33D
-   INTEGER, OPTIONAL, INTENT(IN ) :: o3input
+         INTENT(INOUT) :: O33D
+   INTEGER, INTENT(IN ) :: o3input
 
       real, parameter :: thresh=1.e-9
       real slope
@@ -12011,7 +12010,7 @@ CONTAINS
             QV1D(K)=max(0.,QV1D(K))
          ENDDO
 
-         IF (PRESENT(O33D)) THEN
+         IF (o3input.eq.2) THEN
             DO K=kts,kte
                O31D(K)=O33D(I,K,J)
             ENDDO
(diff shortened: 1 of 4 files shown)


kkeene44 (approved):

dudhia:
I think you are only outputting it from the longwave, so if someone just chose only the shortwave, they would not have it. Is this OK?

weiwangncar:
@dudhia Note that O3RAD isn't a default output variable. One needs to add it to output stream. I had the following thoughts: 1. We do encourage use the RRTMG pairs; 2. When one outputs ozone from shortwave scheme, you won't see any values unless it is during the day, so it could add confusion.

weiwangncar:
Jenkins tests have passed:

    Failed Simulations are: 
    None
    None

dudhia (approved):
Sounds fine.

sael9740:
@dudhia, @weiwangncar -- The changes in https://github.com/wrf-model/WRF/commit/e46673998931fa54b5c6d0f49a71228836130b79 are still buggy for `o3input=0`.

https://github.com/wrf-model/WRF/blob/6233639c599119e76fca17dba9ea211af53a0ba9/phys/module_ra_rrtmg_lw.F#L11729-L11741

https://github.com/wrf-model/WRF/blob/6233639c599119e76fca17dba9ea211af53a0ba9/phys/module_ra_rrtmg_lw.F#L12402-L12426

Since `nlayers > kte`, line 12417 causes a segfault. Based on the usage of `o31d`, do I assume correctly that this assignment on line 12417 is intended to occur only for `k<=kte`?

weiwangncar:
@sael9740 You're probably right. I will take another look. Thanks for spotting it.

weiwangncar:
@sael9740 Can you take a look at PR#[1803](https://github.com/wrf-model/WRF/pull/1803) and see if it is ok. Thanks.

sael9740:
> @sael9740 Can you take a look at PR#[1803](https://github.com/wrf-model/WRF/pull/1803) and see if it is ok. Thanks.

Looks good to me!

PR #1773, opened 2022-09-27
Matching wrf results aarch64 x86 with gcc10.2

juntangc:
This update is to resolve different simulation results on aarch64 and x86 for the same dataset (caused by round up error in the last position for some transcendental functions) 

TYPE: bug fix

KEYWORDS: round up error, transcendental, compiler optimization

SOURCE: Jun Tang, Amazon

DESCRIPTION OF CHANGES:
Problem:
After investigation, different simulation results on aarch64 and x86 for the same dataset with gcc 10.2 are caused by round up error in the last position for some transcendental functions; and several optimization flags (FMA, inverse square root, ...).  

Solution:
Use double precision for the computation of transcendental function at selected locations can make sure exact same output models are produced on aarch64 and x86 (this patch only covers a few PBL and cumulus scheme). Also disabling some risky optimization can guarantee same output as lower optimization level.  

M       phys/module_cu_tiedtke.F
M       phys/module_sf_myjsfc.F

TESTS CONDUCTED: 
1. The mods fixes the correctness problem between aarch64 and x86 for two WRF models (conus2.5km and conus12km).  The output models are matched, bit by bit, on the two specified platform with the patch.

2. The regression tests have passed - as indicated the change should not affect its results.

RELEASE NOTE: Fix numerical divergence on x86 and arm64. For best performance for WRF on arm64 please use armclang. (https://github.com/juntangc/notes/blob/main/release-note.pdf)

### DIFF
--- a/arch/configure.defaults
+++ b/arch/configure.defaults
@@ -793,7 +793,7 @@ CFLAGS_LOCAL    =       -w -O3 -c  # -DRSL0_ONLY
 LDFLAGS_LOCAL   =       
 CPLUSPLUSLIB    =       
 ESMF_LDFLAG     =       $(CPLUSPLUSLIB)
-FCOPTIM         =       -O2 -ftree-vectorize -funroll-loops
+FCOPTIM         =       -Ofast -march=native -mtune=native -funroll-loops -fno-expensive-optimizations -fno-reciprocal-math -fsigned-zeros -fno-unsafe-math-optimizations
 FCREDUCEDOPT	=       $(FCOPTIM)
 FCNOOPT		=       -O0
 FCDEBUG         =       # -g $(FCNOOPT) # -ggdb -fbacktrace -fcheck=bounds,do,mem,pointer -ffpe-trap=invalid,zero,overflow
@@ -2067,6 +2067,49 @@ CC_TOOLS        =      $(SCC)
 NETCDFPAR_BUILD	=      CONFIGURE_NETCDFPAR_BUILD
 
 #insert new stanza here
+
+###########################################################
+#ARCH    Linux aarch64, GCC compiler OpenMPI # serial smpar dmpar dm+sm
+#
+DESCRIPTION     =      GCC ($SFC/$SCC): Aarch64
+DMPARALLEL      =
+OMPCPP          =      -fopenmp
+OMP             =      -fopenmp
+OMPCC           =      -fopenmp
+SFC             =      gfortran
+SCC             =      gcc
+CCOMP           =      gcc
+DM_FC           =      mpif90
+DM_CC           =      mpicc -DMPI2_SUPPORT
+FC              =      CONFIGURE_FC
+CC              =      CONFIGURE_CC
+LD              =      $(FC)
+RWORDSIZE       =      CONFIGURE_RWORDSIZE
+PROMOTION       =
+ARCH_LOCAL      =
+CFLAGS_LOCAL    =      -w -O3 -c
+LDFLAGS_LOCAL   =      -fopenmp
+FCOPTIM         =      -Ofast -mcpu=native -funroll-loops -fno-expensive-optimizations -fno-reciprocal-math -fsigned-zeros -fno-unsafe-math-optimizations
+# for graviton 2 use the folowing flag
+#FCOPTIM         =      -Ofast -march=armv8.2-a+fp16+rcpc+dotprod -funroll-loops -fno-expensive-optimizations -fno-reciprocal-math -fsigned-zeros -fno-unsafe-math-optimizations
+FCREDUCEDOPT    =      $(FCOPTIM)
+FCNOOPT         =      -O0 -fopenmp -frecursive
+FCDEBUG         =      -g $(FCNOOPT)
+FORMAT_FIXED    =      -ffixed-form -ffixed-line-length-0 -fallow-argument-mismatch -fallow-invalid-boz
+FORMAT_FREE     =      -ffree-form -ffree-line-length-0 -fallow-argument-mismatch -fallow-invalid-boz
+FCSUFFIX        =
+BYTESWAPIO      =      -fconvert=big-endian -frecord-marker=4
+FCBASEOPTS      =      -w $(FORMAT_FREE) $(BYTESWAPIO)
+MODULE_SRCH_FLAG=      -I$(WRF_SRC_ROOT_DIR)/main
+TRADFLAG        =      -traditional-cpp
+CPP             =      /lib/cpp CONFIGURE_CPPFLAGS
+AR              =      ar
+ARFLAGS         =      ru
+M4              =      m4 -B 14000
+RANLIB          =      ranlib
+RLFLAGS         =
+CC_TOOLS        =      $(SCC)
+
 
 ###########################################################
 #ARCH    Fujitsu FX10/FX100 Linux x86_64 SPARC64IXfx/SPARC64Xlfx, mpifrtpx and mpifccpx compilers #serial smpar dmpar dm+sm
--- a/arch/configure.defaults
+++ b/arch/configure.defaults
@@ -788,7 +788,9 @@ CC              =       CONFIGURE_CC
 LD              =       $(FC)
 RWORDSIZE       =       CONFIGURE_RWORDSIZE
 PROMOTION       =       #-fdefault-real-8
-ARCH_LOCAL      =       -DNONSTANDARD_SYSTEM_SUBR  CONFIGURE_D_CTSM  # -DAARCH64_X86_CORRECTNESS_FIX
+ARCH_LOCAL      =       -DNONSTANDARD_SYSTEM_SUBR  CONFIGURE_D_CTSM
+# to validate WRF correctness on aarch64 against x86_64; compile with x86 this alternate ARCH_LOCAL option; able to achieve identical output model on ubuntu 18.04 with gcc 10.2 (on graviton and intel processors)
+#ARCH_LOCAL      =       -DNONSTANDARD_SYSTEM_SUBR  CONFIGURE_D_CTSM -DAARCH64_X86_CORRECTNESS_FIX
 CFLAGS_LOCAL    =       -w -O3 -c  # -DRSL0_ONLY
 LDFLAGS_LOCAL   =       
 CPLUSPLUSLIB    =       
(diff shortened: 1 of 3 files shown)


weiwangncar:
@ChunxiZhang-NOAA Can you take a look at the code change in cu_tiedtke, and see if it is ok. Thanks.

weiwangncar:
@dudhia Could you take a look at the change in MYJ surface layer code and see if it is ok. Thanks.

ChunxiZhang-NOAA:
> @ChunxiZhang-NOAA Can you take a look at the code change in cu_tiedtke, and see if it is ok. Thanks.

@weiwangncar The code change for cu_tiedtke looks fine.

dudhia:
I am not sure that this added complexity in the code is needed. It seems to
only address a library function difference.
Perhaps they can add it as commented code in case users need it for
comparisons between compilers.
Is it clearly wrong without this change or just different?

juntangc:
@dudhia,
In my opinion, WRF output models are different but still correct without this patch.  

Not every user has the level of experience when comes to judging how much differences are acceptable. They probably don't mind the differences until when it comes to important decisions, like adopting a new processor for the workload. They put more weight on correctness than performance and cost. We do have customers with the following questions:  
- why the results are different as WRF are doing floating-point compute, Graviton and Intel should get the same results.
- Is there any tuning that can make Graviton get the same result as Intel? if not, why?

The difference are in the transcendental functions between x86 and aarch64.  Both are justified because the difference is the unit in the last place (1ulp). Luckily, out of all transcendental instances used in WRF, only a few matters to the output model. They are used for the calculation of sensitive parameters (RUBLTEN, RVBLTEN, RTHBLTEN, RUCUTEN, and RVCUTEN...) where small difference are magnified by multiplied with total mu of a vertical column at one surface location (https://github.com/wrf-model/WRF/blob/f34b11dbb89c002c5c0dca1195aab35daeed7349/dyn_em/module_em.F#L2213). As the result, up to 3 m/s difference are observed in wind speed components between output models. 

Baseline - 
```
Differences of the output from two WRF model simulations for a few important fields
Variable Name                Minimum      Maximum      Average      Std Dev        Skew
=========================================================================================
U                            -3.09856      1.85174  2.46794e-06   0.00515369     -26.9818
V                            -1.14332      1.48483 -2.44727e-06   0.00411124      -6.1785
W                          -0.0647928    0.0795061 -4.23795e-07  0.000582573    -0.464928
T                           -0.767151     0.687927 -3.14528e-07    0.0023242      21.2634
PH                            -6.7738      11.8072 -0.000252944    0.0670978      4.89246
QVAPOR                    -0.00105231   0.00109666  5.24334e-10  3.33247e-06       11.923
TSLB                          -0.1297    0.0665894 -1.88114e-06  0.000664625     -55.7122
MU                           -10.9693      7.76303  2.82832e-05     0.106964     -11.4369
TSK                          -4.28049      1.37746  -0.00013701    0.0333615     -17.6595
RAINC                      -0.0346563    0.0416356  7.38085e-07  0.000415339      9.28308
RAINNC                      -0.374385    0.0686252 -1.45955e-05   0.00187228      -108.44
```
Updated (please note that skew are not defined for two identical volumes)-
```
Variable Name                Minimum      Maximum      Average      Std Dev        Skew
=========================================================================================
U                                   0            0            0            0          nan
V                                   0            0            0            0          nan
W                                   0            0            0            0          nan
T                                   0            0            0            0          nan
PH                                  0            0            0            0          nan
QVAPOR                              0            0            0            0          nan
TSLB                                0            0            0            0          nan
MU                                  0            0            0            0          nan
TSK                                 0            0            0            0          nan
RAINC                               0            0            0            0          nan
RAINNC                              0            0            0            0          nan
```

dudhia:
An alternative is to apply these differences as compile-time #ifdef sections via -D.

juntangc:
That's a great solution.  I will update the PR.

Regards,

Jun

phys/module_cu_tiedtke.F
```diff
@@ -2936,6 +2936,9 @@ subroutine cuentr_new                              &
             ztmzk = -(pgeoh(jl,ikh)-pgeoh(jl,ikt))*zrg
             arg = 3.1415*(zzmzk/ztmzk)*0.5
             zorgde = tan(arg)*3.1415*0.5/ztmzk
+#ifdef aarch64_x86_correctness_fix
+            zorgde = real(tan(dble(arg))*3.1415*0.5/dble(ztmzk))
+#endif
```
mgduda, line 2941: With the preprocessing directives here and in `module_sf_myjsfc.F` as written now, it looks like some calculations will be done twice when `aarch64_x86_correctness_fix` is defined; is that intended?

dudhia: Yes, #else could be used, but if this is just for testing and not full runs, the efficiency won't matter.

juntangc:
Thanks for the suggestions, 
I have removed redundant computation.  

Regards,

phys/module_cu_tiedtke.F
juntangc: We do intend this to be used for future runs on arm64 machines for customers who have concerns over correctness. I have updated the PR as suggested.  Thanks,

juntangc:
@mgduda, @dudhia, 

Please let me know if anything else needed from my side to get this merged.  Thanks,

mgduda:
> @mgduda, @dudhia,
> 
> Please let me know if anything else needed from my side to get this merged. Thanks,

Just to check my own understanding, was there any intent to add a definition of `aarch64_x86_correctness_fix` by default when compiling with the new configuration stanza? If not, should we document somewhere for users that `-Daarch64_x86_correctness_fix` can be added, e.g., in `ARCH_LOCAL` if it's necessary to match the results from x86-64 systems with the GNU compilers?

dudhia:
I agree with Michael. The stanza makes no mention of what string needs to be added and a commented out string could be added in ARCH_LOCAL since this is not the default.

dudhia:
It is now the default in optimized but not debugging code. Check if that is what you want.

juntangc:
@mgduda, @dudhia,

Thanks for the suggestions, I have add 

> It is now the default in optimized but not debugging code. Check if that is what you want.

Thanks for the recommendation.  If I put it under ARCH_LOCAL, the option will be taken by both optimized and debug build, right? I intend to have this enabled by both x86_64 and aarch64 when the comparison is needed. I will make another revision.  

Regards,

Jun

weiwangncar:
This PR has passed reg tests:

    Failed Simulations are: 
    None
    None

juntangc:
@mgduda, @dudhia,

Do you see anything remaining or any data that I can provide to get this PR merged?

Regards,

Jun

arch/configure.defaults
```diff
@@ -788,7 +788,7 @@ CC              =       CONFIGURE_CC
 LD              =       $(FC)
 RWORDSIZE       =       CONFIGURE_RWORDSIZE
 PROMOTION       =       #-fdefault-real-8
-ARCH_LOCAL      =       -DNONSTANDARD_SYSTEM_SUBR  CONFIGURE_D_CTSM
+ARCH_LOCAL      =       -DNONSTANDARD_SYSTEM_SUBR  CONFIGURE_D_CTSM  # -DAARCH64_X86_CORRECTNESS_FIX
```
mgduda, line 791: It's not at all obvious or documented, but I think adding comments to the end of the `ARCH_LOCAL` lines in configuration stanzas can be problematic. The `ARCH_LOCAL` line is appended by the `Config.pl` script to include NetCDF4 options; see [Config.pl L.801 - 812](https://github.com/wrf-model/WRF/blob/v4.4.1/arch/Config.pl#L801-L812) in the v4.4.1 release.

The result of adding `# -DAARCH64_X86_CORRECTNESS_FIX` to the end of `ARCH_LOCAL` is that anything from `$(NETCDF4_IO_OPTS)` is ignored as a comment. Here's the resulting line in my `configure.wrf` file:
```
ARCH_LOCAL      =       -DNONSTANDARD_SYSTEM_SUBR  -DWRF_USE_CLM  # -DAARCH64_X86_CORRECTNESS_FIX $(NETCDF4_IO_OPTS)
```
One possible solution might be to just provide an alternative `ARCH_LOCAL` definition as a comment:
```
ARCH_LOCAL      =       -DNONSTANDARD_SYSTEM_SUBR  CONFIGURE_D_CTSM
# ARCH_LOCAL      =       -DNONSTANDARD_SYSTEM_SUBR  CONFIGURE_D_CTSM -DAARCH64_X86_CORRECTNESS_FIX
```
Longer-term, I think the best solution would be to fix the logic in the `Config.pl` script, and I'd be willing to help with this, but it might be a little while before I could do so.

mgduda (changes requested):
I think we need to fix the issue I've noted with the `ARCH_LOCAL` flags.

arch/configure.defaults
juntangc: Michael,

I am not aware of the Config.pl.  Thanks for pointing this out.  The second solution works for me.  It would be nice that you can help with the logic in Config.pl; but it is not required at the moment.

Regards,

Jun

juntangc:
> I think we need to fix the issue I've noted with the `ARCH_LOCAL` flags.

I have made a revision for this.  Please review.  thanks,

DDxPlague:
@mgduda with the current revisions @juntangc made, is there something blocking this from being merged?

juntangc:
@mgduda, I have updated.  Thanks for the recommendation,

> > @mgduda with the current revisions @juntangc made, is there something blocking this from being merged?
> 
> I've just left an updated review. The addition of a single space should be all that's needed.

weiwangncar:
The reg tests have passed:

    Failed Simulations are: 
    None
    None

DDxPlague:
Who else do we need to review this so we can go ahead and merge?

mgduda:
@dudhia Would you be able to add a review for this PR?

dudhia:
I see that one stanza has it as the default ARCH_LOCAL and the other has it
commented out. Is this the intent?

juntangc:
Yes. This way the result with the default setting remains the same.  And it is easy enough to adjust the setting and repeat this experiment for those who have concerns over correctness. 
Regards,
Jun

weiwangncar:
@juntangc Would you like to include any message for the code release? The message for release note is extracted from the 'RELEASE NOTE' section of this PR.

juntangc:
@weiwangncar, Thanks for the reminder, 
I have put a release note and short instruction on this patch.

This type of numerical divergence are likely to show up again when using a different math library, version of math library, or compiler (Arm, Intel, ...). It will be helpful if a criteria for acceptable differences can be established and published for arm64 and other architectures, similar to x86 (https://www2.mmm.ucar.edu/wrf/users/benchmark/v42/benchdata_v422.html).

PR #1775, opened 2022-10-03
LFC, CAPE, and CIN Calculation Improvement

zhixiaozhang:
TYPE: Bug fix

KEYWORDS: LFC, CAPE, CIN Calculations

SOURCE: "Zhixiao Zhang (zhixiao.zhang@physics.ox.ac.uk, University of Oxford and University of Utah)

DESCRIPTION OF CHANGES:
Problem:
1) LFC identification doesn't function well when the vertical profile has multiple temperature inversion layers.
2) CAPE and CIN are slightly different from the definitions in AMS glossaries.
3) LCL pressure level index is redundant since the LCL pressure is already recorded.

Solution:
1) Reset the LFC identification flag when there are inversion layers above the current LFC level, ensuring the highest LFC is the final level of free convection.
2) To be consistent with AMS glossaries, CAPE and CIN are defined as the integrated LFC-EL (equilibrium level) positive buoyancy and LCL-LFC negative buoyancy, respectively. This is to replace the previous CAPE and CIN calculations that directly integrate positive and negative buoyancy from LCL to EL as CAPE and CIN.
3) Remove redundant searching for the pressure level index of LCL.

RELEASE NOTE: Refine CAPE and CIN calculations to be consistent with AMS definitions. Improve LFC identification for adapting to multi-inversion layers profiles.


weiwangncar:
Jenkins test has passed:

    Failed Simulations are: 
    None
    None

weiwangncar (approved):
Diagnostics only.

weiwangncar:
@dudhia Since this is categorized as 'enhancement', are you ok with merging it into 4.4.2 release?

zhixiaozhang:
@dudhia @weiwangncar Thanks for your follow-up. This is in between bug-fix and enhancement. I could relabel it as a bug fix if it is more appropriate.

weiwangncar:
@zhixiaozhang I'm ok with this being labeled as 'bug fix'.

zhixiaozhang:
> @zhixiaozhang I'm ok with this being labeled as 'bug fix'.

Thanks, I have updated the label.

sael9740:
@weiwangncar - This PR introduced a bug.

`plfc` is initialized to a dymmy value here:

https://github.com/wrf-model/WRF/blob/3dd1546ba425b5e77d682dd92253738416b3cb0a/phys/module_diag_functions.F#L518-L526

Once the `lfclev` is found then `plfc` is then assigned it final value:
https://github.com/wrf-model/WRF/blob/3dd1546ba425b5e77d682dd92253738416b3cb0a/phys/module_diag_functions.F#L808-L813

Notice that `CAPE` is only updated when `p(k) <= plfc`:

https://github.com/wrf-model/WRF/blob/3dd1546ba425b5e77d682dd92253738416b3cb0a/phys/module_diag_functions.F#L774-L777

This makes sense based on the PR description but the problem is that the `CAPE` assignment happens before `plfc` is assigned. Therefore the condition on line 774 uses the dummy value for `plfc` and it always ends up being false. The overall effect of this is that `cape` is never updated and the WRF outputs for `AFWA_CAPE`, `AFWA_CAPE_MU`, etc. are always zero now.

All we need to do is move the `plfc = p(lfclev)` assignment to somewhere before it is used for the `cin`/`cape` assignment loop at line 821.

sael9740:
Do the regression tests not include something for checking AFWA diagnostics output fields?

weiwangncar:
@zhixiaozhang Can you take a look at the comments by @sael9740 and also Issue #[1903](https://github.com/wrf-model/WRF/issues/1903)? Thanks!

weiwangncar:
@sael9740 Unfortunately the regression tests do not test diagnostics.

PR #1801, opened 2022-12-29
Updates and fixes to module_firebrand_spotting

mefrediani:
TYPE: bug fix and namelist options

KEYWORDS: short grass factor, release location, release height, track ember (for research only). 

SOURCE: Maria Frediani (NCAR) 

DESCRIPTION OF CHANGES:

The updates include the following bugfix:
* change the spotting threat factor for short grass from 0 to 1 (i.e. now spotting can occur on short grass) 
* change release location to be based on instant ROS rather than ROS between release intervals (before embers were being released too far behind the firefront) 
* aligned min land height with release height - the option to set landing height on the namelist was allowing embers to be released below land height, causing them to land immediately. Now the min release height in 1m + min land height

The updates include the following new options on the namelist:

* option to print ember properties in rsl files (user must set trackember = .true., default is false. In the future, this must be improved: data should be written on a distinct file). This option is intended for research and debugging purposes only, not for regular use. 
* include a new variable fs_p_src to store ember tile source - this is used to determine the ember origin when it travels between tiles. It should only be used when the number of embers is low. This option is only relevant when trackember = .true.

LIST OF MODIFIED FILES: 
M       Registry/registry.fire
M       dyn_em/solve_em.F
M       dyn_em/start_em.F
M       phys/module_firebrand_spotting.F
M       phys/module_firebrand_spotting_mpi.F

TESTS CONDUCTED: 
1. Simulations using gnu + openmpi, in both debug and standard configuration 
2. Jenkins tests have passed.

RELEASE NOTE: The firebrand spotting parameterization was updated to (1) allow likelihood of spotting on short grass and (2) fix the location and height of ember release.


mefrediani:
@weiwangncar @dudhia The tests are not completing. Do you have more information about why?

dudhia:
We have been having problems with our tests. @weiwangncar can check them and get back to you. It is not necessarily a failure.

weiwangncar:
The Jenkins test results:

    Failed Simulations are: 
    None
    None

dudhia (approved):

weiwangncar (approved):

PR #1830, opened 2023-03-22
Test for RPC instead of checking hardcoded paths

vanderwb:
Change to configure to test for RPC package instead of checking hardcoded paths, so that non-standard RPC installs can still be found.

TYPE: enhancement

KEYWORDS: configure, build, rpc, make, test

SOURCE: Brian Vanderwende (NCAR)

DESCRIPTION OF CHANGES:
Problem:
If an RPC library is installed in a non-standard prefix with a package manager like Spack, WRF will not be able to find it.

Solution:
Instead of checking hardcoded locations for `rpc/types.h`, compile a test program to check for one of the variants. Since symbolic linking hacks are occasionally done to make tirpc friendlier, it's also good to check constants to confirm the version used.

LIST OF MODIFIED FILES (from develop):
M       Makefile
M       configure
A       tools/rpc_test.c

TESTS CONDUCTED: 
1. The success is determined by running `configure` on a system that uses Spack to install libtirpc.
2. The Jenkins tests have passed - no impact on other part of code.

RELEASE NOTE: Test for RPC headers to support non-standard locations.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1095,6 +1095,10 @@ fortran_2003_fflush_test:
 fortran_2008_gamma_test:
 	@cd tools ; /bin/rm -f fortran_2008_gamma_test.{exe,o} ; $(SFC) -o fortran_2008_gamma_test.exe fortran_2008_gamma_test.F ; cd ..
 
+# rule used by configure to test for RPC support
+rpc_test:
+	@cd tools ; /bin/rm -f rpc_test.exe ; $(SCC) -DUSE_TIRPC -o rpc_test.exe rpc_test.c ; $(SCC) -o rpc_test.exe rpc_test.c; cd ..
+
 toolsdir :
 	@ echo '--------------------------------------'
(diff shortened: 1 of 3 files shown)


weiwangncar:
@mgduda Can you review this PR and see if we should consider this for release-v4.5? Thanks!

weiwangncar:
The Jenkins tests have passed:

    Failed Simulations are: 
    None
    None

weiwangncar:
@mgduda If you could, please review this one as well for 4.5. Thanks...

mgduda (approved):
This represents an improvement over the hard-coded paths that are currently checked for `rpc/types.h`.

mgduda:
Apparently my review was taken to be on behalf of the wrf-model/infrastructure team, so I think we'll need a second person to review these changes.

weiwangncar (approved):

PR #1834, opened 2023-03-24
bugfix Mp38 allocatable array dimension

mefrediani:
Large memory requirement in Thompson microphysics mp=8 and mp=28

TYPE: bug fix

KEYWORDS: memory, thompson, microphysics

SOURCE: Maria Frediani (NCAR) 

DESCRIPTION OF CHANGES:
The new graupel (qg) table created for mp=38 required a larger memory to either compute or use. Because of that, when using options 8 or 28, the model stopped right away if there wasn't enough memory. 

These changes make the 5D arrays truly allocatable, meaning when Thompson option 8 or 28 are used, the middle dimension (NRHG) is 1, and when 38 is used that dimension becomes 9. This should reduce the memory required when option 8 and 28 are used, and the newer qg table written out by the new code won't be as huge as it is now, but more similar to what it was before option 38 was introduced. 

Also, the table was initially named qr_acr_qg_mp28V4, but this table is also used for option 8. For this reason, the name was changed to qr_acr_qg_V4.

LIST OF MODIFIED FILES: phys/module_mp_thompson.F

TESTS CONDUCTED: 
1. Bit for bit comparison using diffwrf to compare the results between the current develop branch and the new code using mp=38 and mp=28 
2. It passes the regression tests.


phys/module_mp_thompson.F
```diff
@@ -4275,11 +4275,11 @@ subroutine qr_acr_qg(NRHGtable)
 
         if (.not. is_hail_aware) then
 !. Distribute the same value across NRHG dimension
-         do n3 = 2, NRHG
+         do n3 = 2, dimNRHG
```
weiwangncar, line 4278: @mefrediani If dimNRHG = 1, would it work? Maybe you could check the value as a condition to do this loop?

weiwangncar: @mefrediani On second thought, perhaps this loop is no longer needed. What do you think?

mefrediani: Good catch, Wei! I can't believe I missed this!

dudhia: Yes, probably no longer needed. Looks good so far. Hopefully this will solve the problem.

weiwangncar:
@mefrediani Maybe we could take this opportunity to change the new table name for option 8 and 28 to be qr_acr_qg_V4.dat or something that does not imply option 28?

mefrediani:
@weiwangncar I had already changed the table name ;)

weiwangncar:
@mefrediani @dudhia I just tested with our obs-nudging case which used to stop right a way when 30 processors were used with one 36 processor node requested, it is now working! And the table size for option 8/28 has been reduced from 809,637,600 to 74,966,480 k-bytes! For reference, the qr_acr_qg table from prior WRF version is 89,959,776.

mefrediani:
That's great, @weiwangncar! Thanks for finding a simple solution to the issue. Your directions were very helpful!

weiwangncar:
@mefrediani You're welcome. I'm glad it is this simple. Do you know if the content in the new table for 8 and 28 has changed?

dudhia:
Sounds good. Loop could be removed completely instead of left commented out.

weiwangncar:
The regression test results:

    Failed Simulations are: 
    None
    None

mefrediani:
@weiwangncar effectively, the tables in WRFv4.4 should provide the same values for mp28 and mp8. However, with this PR, we now have removed the variable tmg_gacr, so the tables will be different (that's why the new table is smaller).

weiwangncar:
@mefrediani Thanks! I did realize the removal of the unused variable in the table. But it is good to know the content has not changed for options 8 and 28.

dudhia (dismissed):
loop is still there as commented out, but we can leave it

smileMchen:
@mefrediani 
Just rerun a nested case that failed previously. The case can run successfully with this fix. Many thanks !

mefrediani:
@smileMchen I'm glad it worked! Thanks to @weiwangncar and the WRF team who suggested a simple solution to the issue.

weiwangncar (approved):

dudhia (approved):

PR #1836, opened 2023-03-24
Fix Lightning Potential Index calculation failure

weiwangncar:
TYPE: bug fix

KEYWORDS: LPI, malloc error, NaNs

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
The model stops right away with malloc error (with gfortran) when using lightning_option = 3, the option to compute lightning potential index (the problem occurred only with a 2 domain test). The error is due to variable packaging for this option. Once this error is fixed, the model can run but produces NaNs in the diagnostics. Array for hail mixing ratio is removed from the call as well, as it is not used.

Solution:
Add additional variables for packaging, and limit hydrometeors to a number that is not smaller than 1.e-10, which eliminates the NaNs.

LIST OF MODIFIED FILES: 
M       phys/module_lightning_driver.F
M       phys/module_ltng_lpi.F

TESTS CONDUCTED: 
1. The model ran with 2 domain test, and do not produce NaNs.

RELEASE NOTE: Fixed a bug for lightning diagnostic option 3, an option for lightning potential index calculation. The bug prevented nested case to run, and may produce NaNs in the diagnostics.


weiwangncar:
The Jenkins test results:

    Failed Simulations are: 
    None
    None

dudhia:
When you say it produces NaNs in the diagnostics, is that still the case after the fix? It was not clear.

weiwangncar:
@dudhia Added more in 'Solution'.

smileMchen (approved):

PR #1251, opened 2020-07-10
Temporary fix for gfortran/10 compiler complaints

syalavarthi:
TYPE: bug fix

KEYWORDS: gfortran/10

SOURCE: Srikanth Yalavarthi (Marvell Semiconductor)

DESCRIPTION OF CHANGES: 
Problem:
Due to argument inconsistencies, the WRF code will not compile with gfortran/10. 

Here are some examples of incorrect Fortran argument usages:

```
Error: Rank mismatch between actual argument at (1) and actual argument at (2) (rank-1 and scalar)
wrf_io.f:7605:71:

 7605 |                                                    ,i1,i2,j1,j2,k1,k2 )
      |                                                                       1
```
```
Error: Type mismatch in argument ‘field’ at (1); passed INTEGER(4) to REAL(8)
wrf_io.f:7744:49:

 2538 |     stat = NF_GET_ATT_INT (DH%NCID,NF_GLOBAL,Element,Buffer)
      |                                                     2
......
 7744 |     stat = NF_GET_ATT_INT(NCID,VarID,'FieldType',FType)
      |                                                 1
```
```
Error: Rank mismatch between actual argument at (1) and actual argument at (2) (scalar and rank-3)
field_routines.f:158:52:

  110 |     stat = NF_GET_VARA_INT(NCID,VarID,VStart,VCount,Data)
      |                                                    2
......
  158 |     stat = NF_GET_VARA_INT(NCID,VarID,VStart,VCount,Buffer)
      |                                                    1
```
```
Error: Rank mismatch in argument ‘fileindex’ at (1) (scalar and rank-1)
io_grib1.f90:685:27:

  685 |   CALL gr1_fill_eta_levels(fileinfo(DataHandle)%fileindex(:), FileFd(DataHandle), &
      |                           1
```
```
Error: Type mismatch between actual argument at (1) and actual argument at (2) (INTEGER(4)/REAL(4)).
module_internal_header_util.f:1592:39:

 1592 |                            DataHandle, Data, Count, code )
      |                                       1
......
 1654 |                            DataHandle, Data, Count, code )
      |                                       2
```
```
Error: Rank mismatch between actual argument at (1) and actual argument at (2) (rank-1 and scalar)
module_io_int_read.f:129:39:

  129 |     call mpi_file_read_at(ifd, offset, tmp, 1, &
      |                                       1
......
  573 |     call mpi_file_read_at(ifd, offset, tmp, num, &
      |                                       2
```
```
[...]

PR #1848, opened 2023-04-21
Apply type-mismatch for gcc-10 to all gcc targets

stephenmsachs:
TYPE: bug-fix

KEYWORDS: GCC

SOURCE: Stephen Sachs (AWS)

DESCRIPTION OF CHANGES:
Problem:
GCC-10 onwards compilation fails on ARM

Solution:
In order to patch GCC arguments when using GCC-10+ a line `FCCOMPAT` is necessary in `configure.wrf`. This patch is an extension to https://github.com/wrf-model/WRF/pull/1251.

### DIFF
--- a/arch/configure.defaults
+++ b/arch/configure.defaults
@@ -119,8 +119,9 @@ FCDEBUG         =       # -g $(FCNOOPT)
 FORMAT_FIXED    =       -ffixed-form
 FORMAT_FREE     =       -ffree-form -ffree-line-length-huge
 FCSUFFIX        =
+FCCOMPAT        =
 BYTESWAPIO      =       -fendian=big
-FCBASEOPTS_NO_G =       -Wno=101,139,155,158 $(FORMAT_FREE) $(BYTESWAPIO)
+FCBASEOPTS_NO_G =       -Wno=101,139,155,158 $(FORMAT_FREE) $(BYTESWAPIO) $(FCCOMPAT)
 FCBASEOPTS      =       $(FCBASEOPTS_NO_G) $(FCDEBUG)
 MODULE_SRCH_FLAG =     -fmod=$(WRF_SRC_ROOT_DIR)/main
 TRADFLAG        =      CONFIGURE_TRADFLAG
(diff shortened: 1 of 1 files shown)


weiwangncar:
The regression test results:

    Failed Simulations are: 
    None
    None

stephenmsachs:
Re-worded the title to be more clear about the meaning. This fix makes sure the `sed` work here https://github.com/wrf-model/WRF/blob/21c72141142fc6c8d203d2bf79f1990e45a0aef8/configure#L721 is applied for ever gcc target.

was: "Apply pull 1251 to all GCC targets"

weiwangncar:
@mgduda @islas Can you review this PR? Thanks!

arch/configure.defaults
```diff
@@ -119,8 +119,9 @@ FCDEBUG         =       # -g $(FCNOOPT)
 FORMAT_FIXED    =       -ffixed-form
 FORMAT_FREE     =       -ffree-form -ffree-line-length-huge
 FCSUFFIX        =
+FCCOMPAT        =
 BYTESWAPIO      =       -fendian=big
-FCBASEOPTS_NO_G =       -Wno=101,139,155,158 $(FORMAT_FREE) $(BYTESWAPIO)
+FCBASEOPTS_NO_G =       -Wno=101,139,155,158 $(FORMAT_FREE) $(BYTESWAPIO) $(FCCOMPAT)
```
mgduda, line 124: This build target is for the `g95` compiler if I'm not mistaken. Since the `g95` compiler hasn't been actively developed in quite some time, I'd suggest just leaving this configuration stanza as-is.

arch/configure.defaults
```diff
@@ -984,8 +985,9 @@ FCDEBUG         =       # -g $(FCNOOPT)
...
```
mgduda, line 990: This stanza is also for the `g95` compiler.

mgduda (changes requested):

islas (approved):

weiwangncar (approved):

mgduda (approved):
:shipit:

PR #1812, opened 2023-01-21
BLD: Fix compile on Cygwin.

DWesl:
Allow compilation on Cygwin to complete

TYPE: bug fix

KEYWORDS: 5 to 10 words related to commit, separated by commas
compilation,cygwin,

SOURCE: 
Daniel Wesloh (Penn State)

DESCRIPTION OF CHANGES:
Problem:
Generally or specifically, what was wrong and needed to be addressed?

Compiling WRF failed on Cygwin due to lack of netCDF4

Solution:
What was down algorithmically and in the source code to address the problem?

Match assumptions about `USENETCDFPAR` (#1743 would also fix this) and pass flags to allow legacy Fortran constructs (disallowed by default with GCC 10)

ISSUE: For use when this PR closes an issue.
Fixes #1271

LIST OF MODIFIED FILES: list of changed files (use `git diff --name-status master` to get formatted list)
```
M       configure
M       doc/README.cygwin.md
M       doc/README.netcdf4par
```

TESTS CONDUCTED: 
1. Checked whether model compiles on Cygwin in DWesl/wrf#1
2. The Jenkins tests have passed.

RELEASE NOTE: Fix compilation on Cygwin.

PR #1858, opened 2023-05-08
Quiet compiler warnings for registry.exe

DWesl:
Quiet the compiler warnings in registry.exe

TYPE: bug fix

KEYWORDS: compiler warnings, snprintf, initial values, string length

SOURCE: Either "Daniel Wesloh (Pennsylvania State University Department of Meteorology and Atmospheric Science)"

DESCRIPTION OF CHANGES:
Problem:
Generally or specifically, what was wrong and needed to be addressed?
A few years ago, `registry.exe` was segfaulting while compiling WRF.  I enabled every compiler warning for that build process I could find, and tried to fix every one I saw.  The problem ended up being insufficient stack space (fix merged a few years ago), but fixing the warnings probably won't hurt.

Solution:
- make sure string buffers are large enough for the contents put in them
- don't use larger types than necessary (i.e. use `short` instead of `int` for numbers in the 300--16,000 range, so printf knows that five characters will be enough)
- Specify maximum lengths to copy using `snprintf` and friends
- Make sure all variables are initialized.

ISSUE: None

```
M       configure
M       doc/README.cygwin.md
M       inc/streams.h
M       tools/Makefile
M       tools/data.h
M       tools/fseek_test.c
M       tools/gen_allocs.c
M       tools/gen_args.c
M       tools/gen_config.c
M       tools/gen_defs.c
M       tools/gen_interp.c
M       tools/gen_irr_diag.c
M       tools/gen_scalar_indices.c
M       tools/gen_streams.c
M       tools/gen_wrf_io.c
M       tools/misc.c
M       tools/protos.h
M       tools/reg_parse.c
M       tools/registry.c
M       tools/registry.h
M       tools/set_dim_strs.c
M       tools/sym.c
M       tools/sym.h
M       tools/symtab_gen.c
M       tools/type.c
```

TESTS CONDUCTED: 
   - DWesl/WRF#3
   - WRF compile finishes with this patch
   - I have no idea how to run those

RELEASE NOTE: Eliminate some compiler warnings when building registry.exe

### DIFF
--- a/tools/gen_allocs.c
+++ b/tools/gen_allocs.c
@@ -80,13 +80,13 @@ int
 gen_alloc2 ( FILE * fp , char * structname , char * structname2 , node_t * node, int *j, int *iguy, int *fraction, int numguys, int frac, int sw ) /* 1 = allocate, 2 = just count */
 {
   node_t * p ;
-  int tag ;
+  int tag = 0 ;
   char post[NAMELEN], post_for_count[NAMELEN] ;
-  char fname[NAMELEN], dname[NAMELEN], dname_tmp[NAMELEN] ;
-  char x[NAMELEN] ;
-  char x2[NAMELEN], fname2[NAMELEN] ;
-  char dimname[3][NAMELEN] ;
-  char tchar ;
+  char fname[NAMELEN], dname[NAMELEN + 11], dname_tmp[NAMELEN] ;
+  char x[NAMELEN + 4] ;
+  char x2[NAMELEN + 2], fname2[2 * NAMELEN + 1] ;
+  char dimname[3][NAMELEN + 12] ;
+  char tchar = '\0';
   unsigned int *io_mask ;
   int nd ;
   int restart ;
@@ -120,7 +120,7 @@ gen_alloc2 ( FILE * fp , char * structname , char * structname2 , node_t * node,
 */
 if ( tag == 1 )
 {
-     char dname_symbol[128] ;
+     char dname_symbol[NAMELEN + 7] ;
      sym_nodeptr sym_node ;
 
      sprintf(dname_symbol, "DNAME_%s", dname_tmp ) ;
@@ -607,7 +607,7 @@ gen_dealloc2 ( FILE * fp , char * structname , node_t * node )
   int tag ;
   char post[NAMELEN] ;
   char fname[NAMELEN] ;
-  char x[NAMELEN] ;
+  char x[NAMELEN + 4] ;
 
   if ( node == NULL ) return(1) ;
 
--- a/tools/gen_interp.c
+++ b/tools/gen_interp.c
@@ -66,7 +66,7 @@ int contains_tok( char *s1, char *s2, char *delims )
   
 
  /* Had to increase size for SOA from 4*4096 to 4*7000 */
-char halo_define[4*7000], halo_use[NAMELEN], halo_id[NAMELEN], x[NAMELEN] ;
+char halo_define[4*7000], halo_use[NAMELEN] = {'\0'}, halo_id[NAMELEN], x[2 * NAMELEN + 1] ;
 
 /*KAL added this for vertical interpolation */
 /*DJW 131202 modified to create files required for vertical interpolation from parent to nest */
@@ -130,7 +130,7 @@ else if ( down_path[ipath] == FORCE_DOWN  ) { sprintf(halo_id,"HALO_FORCE_DOWN")
 else if ( down_path[ipath] == INTERP_UP   ) { sprintf(halo_id,"HALO_INTERP_UP") ; }
 else if ( down_path[ipath] == SMOOTH_UP   ) { sprintf(halo_id,"HALO_INTERP_SMOOTH") ; }
 sprintf(halo_define,"80:") ;
-sprintf(halo_use,"") ;
+ halo_use[0] = '\0' ;
       gen_nest_interp1 ( fp , Domain.fields, NULL, down_path[ipath], (down_path[ipath]==FORCE_DOWN)?2:2 ) ;
 {
   node_t * comm_struct ;
@@ -168,8 +168,8 @@ gen_nest_interp1 ( FILE * fp , node_t * node, char * fourdname, int down_path , 
   char nddim2[3][2][NAMELEN] ;
   char nmdim2[3][2][NAMELEN] ;
   char npdim2[3][2][NAMELEN] ;
-  char vname[NAMELEN], vname2[NAMELEN] ; 
-  char tag[NAMELEN], tag2[NAMELEN] ; 
+  char vname[3 * NAMELEN + 88], vname2[3 * NAMELEN + 88] ;
+  char tag[NAMELEN] = {'\0'}, tag2[NAMELEN] = {'\0'} ;
   char fcn_name[NAMELEN] ;
   char xstag[NAMELEN], ystag[NAMELEN] ;
   char dexes[NAMELEN] ;
@@ -200,7 +200,7 @@ gen_nest_interp1 ( FILE * fp , node_t * node, char * fourdname, int down_path , 
     if ( nest_mask & down_path )
     {
         if ( p->ntl > 1 ) { sprintf(tag,"_2") ; sprintf(tag2,"_%d", use_nest_time_level) ; }
-        else              { sprintf(tag,"")   ; sprintf(tag2,"")                         ; }
+        else              { tag[0] = '\0'; tag2[0] = '\0'; }
 
         /* construct variable name */
         if ( p->node_kind & FOURD ) {
@@ -359,7 +359,7 @@ fprintf(fp,"                  ngrid%%i_parent_start, ngrid%%j_parent_start,     
 fprintf(fp,"                  ngrid%%parent_grid_ratio, ngrid%%parent_grid_ratio                &\n") ;
    
         {
-           char tmpstr[NAMELEN], *p1 ;
+           char tmpstr[2 * NAMELEN + 8], *p1 ;
            node_t * nd, * pp  ;
            pp = NULL ;
            if ( p->node_kind & FOURD ) {
@@ -377,10 +377,10 @@ fprintf(fp,"                  ngrid%%parent_grid_ratio, ngrid%%parent_grid_ratio
                strcpy( tmpstr , pp->interpu_aux_fields ) ;
 	     } else if ( down_path & FORCE_DOWN ) {
                /* by default, add the boundary and boundary tendency fields to the arg list */
-               if ( ! p->node_kind & FOURD ) {
-                 sprintf( tmpstr , "%s_b,%s_bt,", pp->name, pp->name )  ;
+               if ( ! (p->node_kind & FOURD) ) {
+                 snprintf( tmpstr , 2 * NAMELEN + 8, "%s_b,%s_bt,", pp->name, pp->name )  ;
                } else {
-                 sprintf( tmpstr , "%s_b,%s_bt,", p->name, p->name )  ;
+                 snprintf( tmpstr , 2 * NAMELEN + 8, "%s_b,%s_bt,", p->name, p->name )  ;
                }
                strcat( tmpstr , pp->force_aux_fields ) ;
 	     } else if ( down_path & INTERP_DOWN ) {
@@ -507,7 +507,7 @@ gen_nest_interp2 ( FILE * fp , node_t * node, char * fourdname, int down_path , 
   char ddim[3][2][NAMELEN] ;
   char mdim[3][2][NAMELEN] ;
   char pdim[3][2][NAMELEN] ;
-  char vname[NAMELEN], vname2[NAMELEN] ; 
+  char vname[3 * NAMELEN + 88], vname2[3 * NAMELEN + 88] ;
   char tag[NAMELEN], tag2[NAMELEN] ; 
   char dexes[NAMELEN] ;
   char ndexes[NAMELEN] ;
@@ -546,11 +546,11 @@ gen_nest_interp2 ( FILE * fp , node_t * node, char * fourdname, int down_path , 
           set_dim_strs2 ( p , ddim , mdim , pdim , "", 1 ) ;
         } 
         if ( !strcmp ( ddim[0][1], "kde") ||
-	             ( ddim[1][1], "kde") ||
-		         ( ddim[2][1], "kde")) {	
+	     !strcmp ( ddim[1][1], "kde") ||
+	     !strcmp ( ddim[2][1], "kde")) {
     
             if ( p->ntl > 1 ) { sprintf(tag,"_2") ; sprintf(tag2,"_%d", use_nest_time_level) ; }
-            else              { sprintf(tag,"")   ; sprintf(tag2,"")                         ; }
+            else              { tag[0] = '\0'; tag2[0] = '\0'; }
 
             /* construct variable name */
             if ( p->node_kind & FOURD ) {
--- a/tools/gen_irr_diag.c
+++ b/tools/gen_irr_diag.c
@@ -5,11 +5,12 @@
 #include <unistd.h>
 #include <string.h>
 #include <strings.h>
+#include <ctype.h>
 
-int  nChmOpts = 0;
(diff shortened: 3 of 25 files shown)


weiwangncar:
@DWesl It looks like the compilation is failed for all of the tests. Our regression tests are run using gfortran compiler. Here is a particular error (also see the attached file):

make[2]: Entering directory '/wrf/WRF/tools'
gcc  -DIWORDSIZE=4 -DMAX_HISTORY=25 -DNMM_CORE=0  -c -g registry.c
registry.c:7:11: fatal error: io.h: No such file or directory
    7 | # include <io.h>
      |           ^~~~~~

[output_3.txt](https://github.com/wrf-model/WRF/files/11435209/output_3.txt)

weiwangncar:
OK, now our regression tests have passed:

    Failed Simulations are: 
    None
    None

DWesl:
> OK, now our regression tests have passed:

As have mine on Cygwin (with #1812)

DWesl:
Am I missing anything for this PR?

weiwangncar:
@islas Are any of the changes proposed here in conflict with anything you've been working on?

islas:
@weiwangncar Nope, no conflicts should arise from our respective changes in registry

islas:
I'd say this would be a nice-to-have if we can get it in, but not critical.

weiwangncar:
@islas Thanks. We can leave it on the develop branch.

tools/gen_allocs.c
```diff
@@ -80,13 +80,13 @@ int
...
   node_t * p ;
-  int tag ;
+  int tag = 0 ;
   char post[NAMELEN], post_for_count[NAMELEN] ;
-  char fname[NAMELEN], dname[NAMELEN], dname_tmp[NAMELEN] ;
-  char x[NAMELEN] ;
-  char x2[NAMELEN], fname2[NAMELEN] ;
-  char dimname[3][NAMELEN] ;
-  char tchar ;
+  char fname[NAMELEN], dname[NAMELEN + 11], dname_tmp[NAMELEN] ;
```
islas, line 85: I'm guessing a lot of these extra `NAMELEN` additions are due to adjusting for fields within the registry that have longer lengths than first anticipated. Would it be possible to make these all a common size, or if not for those that share a common reason for increased size to break it out into an aptly named `#define` macros?

DWesl: I think it's mostly concatenating variables of length `NAMELEN` or putting extra bits on the end, e.g. `sprintf(dname, "%s/external/io_netcdf", fname);`, though that specific example would require `char dname[NAMELEN + 19];`.  How much extra room the destination buffer requires depends on how much extra stuff gets tacked on, though I could define  `BUFFER_FOR_DESTINATION` to be the largest such requirement and use `NAMELEN + BUFFER_FOR_DESTINATION` for most destination buffers or `NAMELEN + 2 * BUFFER_FOR_DESTINATION` for others.

islas: That sounds like a good path forward. I'm mostly angling for avoiding cryptic magic numbers where it might not be obvious why that number is what it is, especially when compared to similarly named and initialized variables.

DWesl: Changed, and most warnings eliminated again.  There is one left, which I don't understand and can't get to go away:
```
gen_scalar_indices.c: In function ‘gen_scalar_indices1.isra’:
gen_scalar_indices.c:194:57: error: ‘%s’ directive output between 0 and 124999 bytes may exceed minimum required size of 4095 [-Werror=format-truncation=]
  194 |                   snprintf(fourd_bnd, NAMELEN_LONG + 3, "%s_b",assoc_4d) ;
      |                                                         ^      ~
gen_scalar_indices.c:194:19: note: ‘snprintf’ output between 3 and 125002 bytes into a destination of size 125003
  194 |                   snprintf(fourd_bnd, NAMELEN_LONG + 3, "%s_b",assoc_4d) ;
      |                   ^
```
but I think that will continue to be as fine as it was; it has enough space.

tools/gen_irr_diag.c
```diff
@@ -266,18 +271,18 @@ int irr_diag_scalar_indices( char *dirname )
...
-       sprintf( piece," ,%s_kpp",chm_scheme[i]);
+       snprintf( piece,135," ,%.128s_kpp",chm_scheme[i]);
      strcat( line,piece );
    }
    strcat( line," /)\n" );
    fprintf( fp_inc,line );
    fprintf( fp_inc," \n");
 
-   for( i = 0; i < nChmOpts,rxt_cnt[i] > 0; i++ ) {
+   for( i = 0; /*i < nChmOpts &&*/ rxt_cnt[i] > 0; i++ ) {
```
islas, line 283: It seems like this is relying on `rxt_cnt` to be default initialized to zero. I'm not too sure what it is used for, but given that all other loops use `nChmOpts` to specify number of indices into `rxt_cnt`, keeping this check seems prudent

tools/registry.c
```diff
@@ -157,7 +159,23 @@ main( int argc, char *argv[], char *env[] )
       sprintf( fname_wrk,"%s/Registry_irr_diag",dir ) ;
     }
 //  fprintf(stderr,"Registry tmp file = %s\n",fname_wrk);
-    sprintf(command,"/bin/cp %s %s\n",fname_in,fname_wrk);
+    /* we should be able to implement this using posix_spawn */
```
islas, line 162: Remove unless this is for future use/notes

tools/registry.c
```diff
@@ -169,6 +187,15 @@ main( int argc, char *argv[], char *env[] )
       exit(2) ;
     }
     if( !access( "Registry/registry.irr_diag",F_OK ) ) {
+      /*
```
islas, line 190: Same as above, remove if obsolete

tools/registry.c
```diff
@@ -279,6 +306,15 @@ main( int argc, char *argv[], char *env[] )
    sprintf(command,"del /F /Q %s\n",fname_tmp );
 #else
    if( do_irr_diag ) {
+      /*
```
islas, line 309: Same as above, remove if obsolete

tools/sym.c
```diff
@@ -150,6 +150,7 @@ sym_nodeptr x ;
 int
 sym_forget() 
 {
+  /* for (int i = 0; i < SIZEOF_SYMTAB; i++); free(symtab[i]); free(symtab); */
```
islas, line 153: Remove if obsolete

tools/gen_irr_diag.c
DWesl: The old version was ignoring the result of the `nChmOpts` comparison. Are you suggesting that check be included in the for-loop exit condition?

islas: Ah! You're correct, I forgot that this use of the `,` operator is nulling and not `&&`

I'd imagine that the original intent was to have it included via `&&` given the other loops. Maybe @weiwangncar or someone else with historic knowledge on this section of the Registry can chime in. Otherwise, as long as tests agree, I think that check _should_ be there

tools/registry.c
DWesl: It's notes for a potential future application. My usual development system doesn't have a working `posix_spawn`.  I could try a regular `spawn`, but then I'd have to test for whether the system has that function and I'm not sure how to do that.

I'd forgotten about this until you reminded me.

tools/gen_irr_diag.c
DWesl: Changed to the suggested form.  I'd still like to get an opinion from someone who actually knows this bit of the code.

tools/registry.c
DWesl: Removed to a different branch for consideration later.

tools/reg_parse.c
```diff
@@ -85,22 +85,23 @@
...
   char *p, *q ;
   char *tokens[MAXTOKENS], *toktmp[MAXTOKENS], newdims[NAMELEN_LONG], newdims4d[NAMELEN_LONG],newname[NAMELEN_LONG] ;
-  int i, ii, len_of_tok ;
+  unsigned int i, ii;
+  ssize_t len_of_tok = 0;
   char x, xstr[NAMELEN_LONG] ;
-  int is4d, wantstend, wantsbdy ;
-  int ifdef_stack_ptr = 0 ;
-  int ifdef_stack[100] ;
+  unsigned char is4d = 0, wantstend = 0, wantsbdy = 0 ;
```
DWesl, line 102: C99 `stdbool.h` looks to have another option for these.

tools/reg_parse.c
```diff
@@ -117,7 +118,7 @@ pre_parse( char * dir, FILE * infile, FILE * outfile )
     for ( p = inln ; ( *p == ' ' || *p == '	' ) && *p != '\0' ; p++ ) ;
     if ( !strncmp( p , "include", 7 ) &&  ! ( ifdef_stack_ptr >= 0 && ! ifdef_stack[ifdef_stack_ptr] ) ) {
       FILE *include_fp ;
-      char include_file_name[128] ;
+      char include_file_name[130] ;
```
DWesl, line 121:
```suggestion
      char include_file_name[NAMELEN] ;
```

tools/reg_parse.c
```diff
@@ -256,7 +275,7 @@ pre_parse( char * dir, FILE * infile, FILE * outfile )
 	        if ( !strcmp( tokens[F_USE] , tracers[i] ) ) found = 1 ; 
               }
 	      if ( found == 0 ) {
-	        sprintf(tracers[ntracers],tokens[F_USE]) ;
+	        snprintf(tracers[ntracers], 100, tokens[F_USE]) ;
```
DWesl, line 278: Definition on line 89; should this be another magic constant?

tools/registry.c
```diff
@@ -1,16 +1,17 @@
...
+#if defined(_WIN32) || defined(__CYGWIN__)
 # include <io.h>
+#endif
+#ifdef _WIN32
 # define rindex(X,Y) strrchr(X,Y)
 # define index(X,Y) strchr(X,Y)
-#else
+#endif
 # include <sys/time.h>
 # include <sys/resource.h>
```
DWesl, line 11: I suspect these two headers are still unavailable on Windows; the others might be.

tools/registry.h
```diff
@@ -1,6 +1,12 @@
 #ifndef REGISTRY_H
+#include <stdlib.h>
+#include <ctype.h>
+#include <sys/unistd.h>
```
DWesl, line 4: This might also be unavailable on Windows.

tools/registry.c
```diff
@@ -157,7 +159,8 @@ main( int argc, char *argv[], char *env[] )
       sprintf( fname_wrk,"%s/Registry_irr_diag",dir ) ;
     }
 //  fprintf(stderr,"Registry tmp file = %s\n",fname_wrk);
-    sprintf(command,"/bin/cp %s %s\n",fname_in,fname_wrk);
+    /* we should be able to implement this using posix_spawn */
+    sprintf(command,"/bin/cp \'%s\' \'%s\'\n",fname_in,fname_wrk);
```
DWesl, line 163: This will still break if the filename includes apostrophes, but I think that's unlikely enough to not try re-implementing [Python's `shlex.quote`](https://docs.python.org/3/library/shlex.html#shlex.quote).

DWesl (commented):
A few magic constants I didn't catch earlier, plus some possibly-premature optimizations.

weiwangncar:
@DWesl The compilation failed in our regression test. If you need to see the output, let me know.

weiwangncar:
The regression test results from the last commit:

    Failed Simulations are: 
    None
    None

weiwangncar:
@islas Is this ready for approval?

islas:
@weiwangncar I think @DWesl has a lot of self-marked revisions they are still going through. Is that correct?

tools/gen_interp.c
```diff
@@ -546,11 +546,11 @@ gen_nest_interp2 ( FILE * fp , node_t * node, char * fourdname, int down_path ,
...
         } 
         if ( !strcmp ( ddim[0][1], "kde") ||
-	             ( ddim[1][1], "kde") ||
-		         ( ddim[2][1], "kde")) {	
+	     !strcmp ( ddim[1][1], "kde") ||
+	     !strcmp ( ddim[2][1], "kde")) {
     
             if ( p->ntl > 1 ) { sprintf(tag,"_2") ; sprintf(tag2,"_%d", use_nest_time_level) ; }
-            else              { sprintf(tag,"")   ; sprintf(tag2,"")                         ; }
+            else              { tag[0] = '\0'; tag2[0] = '\0'; }
```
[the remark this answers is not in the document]
DWesl: Alternately, `strcpy`/`strncpy`

DWesl:
I completed all the simple work I noticed on a second passthrough.  There's a few places that need a choice or where I would appreciate feedback.

DWesl:
Something that should be noted before the final merge: I accidently added a Cygwin workflow in GitHub actions; I'm not sure if the maintainers want to keep that in or out of the main repository.

weiwangncar:
@DWesl Thanks for removing the Cygwin changes!
@islas Are we ok with this PR?

islas:
I still see some Cygwin changes to configure and the workflow. I'm unsure if that is something we will want to add, though having a test would be incredibly helpful. That content & discussion might be more suited to a Cygwin-specific PR

tools/gen_interp.c
```diff
@@ -130,7 +130,7 @@ else if ( down_path[ipath] == FORCE_DOWN  ) { sprintf(halo_id,"HALO_FORCE_DOWN")
 else if ( down_path[ipath] == INTERP_UP   ) { sprintf(halo_id,"HALO_INTERP_UP") ; }
 else if ( down_path[ipath] == SMOOTH_UP   ) { sprintf(halo_id,"HALO_INTERP_SMOOTH") ; }
 sprintf(halo_define,"80:") ;
-sprintf(halo_use,"") ;
+ halo_use[0] = '\0' ;
```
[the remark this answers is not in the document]
DWesl: Or `strcpy`, since this isn't using any of the formatting features.

DWesl:
I split off the remaining Cygwin-specific fixes to DWesl/WRF#5.  This should be just the bits that address warnings, with comments on the bits I thought might be controversial.

tools/gen_irr_diag.c
```diff
@@ -7,10 +7,13 @@
...
 #include <ctype.h>
 
-int  nChmOpts = 0;
-char rxt_tbl[5][1000][128];
-char chm_scheme[5][128];
-int  rxt_cnt[5];
+#define TABLE_ENTRY 128
+#define EXTRA_FOR_DEST_BUFFER 32
+
+short int  nChmOpts = 0;
```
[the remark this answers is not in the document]
DWesl: I think with `EXTRA_FOR_DEST_BUFFER` there's enough space, and `int` is what everyone will be keeping an eye on.
```suggestion
int  nChmOpts = 0;
```

weiwangncar:
@DWesl This PR has code conflicts with what is in the latest repository. Can you resolve these conflicts while preserving the latest change in the repository?

DWesl:
There's a number of places that change `!(p->node_kind & BIT_MASK)` to `(!p->node_kind) & BIT_MASK`, which is basically always going to return false.  Could you point me to the PR that made that change and the reasoning for it?  My inclination otherwise would be to revert that change or to make it `(~p->node_kind & BIT_MASK)`.

DWesl:
Apparently that's #1942, which also attempted to fix warnings, in this case a warning that `! p->node_kind & BIT_MASK` almost certainly wasn't doing what was wanted, but was instead applying a logical not to a bit sequence (which results in either 1 or 0) and then applying the bit mask (here 8).  Should I go through and change those to be either a bitwise-invert or apply the parentheses implied by the warning (so the bit-mask gets applied before the logical-not)?

DWesl:
Discussion resolved there, conflicts resolved here.

DWesl:
Should I try to squash the changes into fewer commits?

weiwangncar:
@islas Can you see if this PR can be accepted?

DWesl:
I need to check locally to see if there's been more warnings in the past year or two

PR #1876, opened 2023-06-06
Add 3-moment option to NSSL microphysics (plus other minor scheme updates)

MicroTed:
This upgrades the NSSL microphysics scheme to include a 3-moment option that predicts the reflectivity moment for rain, graupel, and hail

TYPE: enhancement

KEYWORDS: microphysics, NSSL scheme

SOURCE: Ted Mansell (NOAA/NSSL) (MicroTed)

DESCRIPTION OF CHANGES:

3-moment bulk microphysics provides more accurate size-sorting of fast-falling hydrometeors (rain, graupel, hail) by effectively predicting the shape parameter of the gamma function particle size distribution. Also various updates to increase run-time flexibility, for example, it is now possible to predict CCN concentration but have hail turned off. Some bug fixes and changes in default settings, as well. The previous mp_physics options (17,19,21,22) still work as expected, but can be replicated with mp_physics=18 plus new namelist options. A new README.NSSLmp gives an overview of the scheme and tips on usage and some details on an internal namelist for setting additional options within the module. The 5th-order WENO advection (module_advect_em.F) has a decreased 'epsilon' value (1.d-40) to accommodate the order of magnitude of reflectivity moments. This should not noticeably affect scalar advection but will cause round-off differences.  Another change fixes an NSSL-MP bug where sedimentation did not work for k > 128 (i.e., if number of model levels exceeds 128, sedimentation only worked for k <= 128).

LIST OF MODIFIED FILES: 

Registry/Registry.EM_COMMON
Registry/registry.var
Registry/wrfplus
chem/chemics_init.F
dyn_em/module_advect_em.F 
dyn_em/solve_em.F
dyn_em/start_em.F
phys/module_mp_nssl_2mom.F
phys/module_physics_init.F
phys/module_diagnostics_driver.F
phys/module_diag_nwp.F
wrftladj/module_microphysics_driver_ad.F
wrftladj/module_microphysics_driver_tl.F
run/README.namelist
share/module_check_a_mundo.F
doc/README.NSSLmp (new file)

TESTS CONDUCTED: 
It passed the regression tests.

RELEASE NOTE: Adds 3-moment option to NSSL microphysics scheme (for rain, graupel, and hail) (mp_physics=18). Changes to snow aggregation parameters reduce stratiform reflectivity bias. Also adds namelist options to enable/disable bulk CCN prediction and graupel/hail density prediction, and deprecates old mp_physics options (17, 19, 21, 22). Main default parameter changes are for graupel/hail fall speed options (icdx, icdxhl; changed from 3 to 6), and default maximum gr/hail droplet collection efficiencies (ehw0/ehlw0 changed from 0.5/0.75 to 0.9/0.9) See README.NSSLmp for details. Also fixes a problem for high resolution vertical domains with nz > 128, where hydrometeor sedimentation did not work at k > 128 (this bug affected most or all previous releases).


dudhia:
Should be targeted to develop.

MicroTed:
Whoops -- sorry, haven't done it in a while. I'll fix that.

weiwangncar:
@MicroTed One of the regression test failed, which is for DA build. I think it is because the new variables you added. Check current registry.var and see what you need to add in the file. If you would like to test DA compile before you change the code, check out this page, and you can find the DA compile instructions near the end of of file (or search for 'WRF DA').

MicroTed:
@weiwangncar Could it be because I moved the 'qh' variables to a separate package? There is a new flag (nssl_hail_on) that adds the hail variables.  Maybe I need to add that elsewhere or perhaps put that feature off. The CCPP version has a single physics option and flags to control optional variables, so I was trying to emulate that in WRF, too. (I don't know if there is a way to disable a variable that is in a selected package, so I made a separate package and added it automatically if needed.)

MicroTed:
OK, that's the next question: can check_a_mundo just check for specific values (that don't have text tokens) and 'reset' the options? That would work (basically what my code change there already does), but then the question is whether those values (17, 22) stay 'reserved' for mp_physics.

weiwangncar:
> @weiwangncar Could it be because I moved the 'qh' variables to a separate package? There is a new flag (nssl_hail_on) that adds the hail variables. Maybe I need to add that elsewhere or perhaps put that feature off. The CCPP version has a single physics option and flags to control optional variables, so I was trying to emulate that in WRF, too. (I don't know if there is a way to disable a variable that is in a selected package, so I made a separate package and added it automatically if needed.)

@MicroTed Check what registry.var currently has for NSSL schemes, and transfer your new change in Registry.EM_COMMON to registry.var.

dudhia:
Yes, I think it can check even unlisted options, but that's a guess. Maybe
it fails before that.

dudhia:
A note on the changed options could be added to the READE.namelist too,
such as old option 17 is now ...

MicroTed:
> A note on the changed options could be added to the READE.namelist too, such as old option 17 is now ...

Yup, that is done -- just not uploaded yet. I removed the old options from the registry, but left a comment that the value are still 'reserved' for the time being.

MicroTed:
In the wrfplus compile, I found a syntax error (leftover comma) in the the microphysics drivers. Easy fix there. 

But I am also getting a bunch of missing halo include files, e.g.:

./wrftladj/module_first_rk_step_part1_tl.F:209:#     include "HALO_EM_PHYS_A_TL.inc"

I suppose it is an automatic variant of HALO_EM_PHYS_A, but I'm not sure where these are generated. It may be a system issue (Darwin, intel chip), as I haven't compiled this before.

dudhia:
@MicroTed this halo should be in one of the other registry files, but I don't immediately see it.

MicroTed:
I am testing again on a linux system (ifort/icc), and the HALO_EM_PHYS_A_TL.inc and others are showing up there. 'compile wrfplus' I suspect there is something about the macos setup that is failing to create those. Anyway, wrfplus seems to compile fine -- at least with ifort/icc. The jenkins output doesn't show a particular compile error, just that a dependency is missing.

The other failure in the output was OpenMP with real8 (seg fault in solve_em). Apparently real8 serial and real4 OpenMP are fine. I'll work on that next.

MicroTed:
I haven't been able to reproduce a failure with -r8 OpenMP, so I went ahead and pushed the syntax error fix in the WRF DA code (which doesn't even use the NSSL_2MOM token, so it could just be removed from the 'use' statement, along with most of the others).

The failure in the output_11 (OpenMP) is right off the bat with no model output. When I recently had a similar error, it was a problem with arrays in the registry, and it showed up in MPI.  Guess we'll see if the next RT passes or not.

MicroTed:
OK, I've been able to get a failure with OpenMP and -r8 using gcc-12 on a linux system. OMP_NUM_THREADS > 2 generates NaNs. It is pretty weird and I don't understand how it is failing.  It works if I completely turn off the the extra 3-moment code, which isn't being used in the test (set for 2-moment). Partially turning off the 3-moment code does some really odd things (like passed dimensions changing value in a subroutine). I've successfully tested gcc-13 on apple M1, so I wonder if it is a compiler quirk, but there must be something.... I'll keep looking.

MicroTed:
Solutions!
OpenMP-r8:
After a number of false leads, the trick was to reduce the size of temporary 1-D work arrays.

WRF-4dvar:
The failure here was in building da_wrfvar.exe, which was getting errors like "relocation truncated to fit: R_X86_64_32S" The fix appears to be to compile both wrfplus and 4dvar with '-mcmodel=medium'. It would seem that the added number of small work arrays pushed the static memory usage over the limit of mcmodel=small. (Despite a more-than-compensating reduction in the array sizes.) 

I'm not sure how best to add the mcmodel option to the configure.wrf specifically for those two programs, since it is not needed for wrf.exe. Any suggestions? This was with gfortran, and I'll try to test again with ifort. (I have also run into the relocation truncation error with UFS and CCPP with the 3-moment code and solved by adding the mcmodel flag to ifort.)

weiwangncar:
@mgduda @islas Can you take a look and make a suggestion? Thanks!

MicroTed:
I don't think the NSSL microphysics is supported by 4dvar, anyway, (right?) so one option might be to use the WRFPLUS flag to switch the main routine to a stub or simpler version (e.g., include file of 2-moment-only routine). Or perhaps use it to exclude the module usage from physics_init and microphysics_driver?
For UFS/CCPP,  the -mcmodel=medium works there, too. It doesn't help to reduce the array fixed sizes -- I set them to 1 (instead of 500) and still get the relocation error.  I don't know if it is the total code size of the module or the size of an individual large subroutine that is pushing the static size over the edge. In other words, I don't know if it would help to try to break out smaller subroutines of code (and it would be a lot of work).

weiwangncar:
@MicroTed I like this solution. Can you give it a try?

weiwangncar:
@MicroTed Congratulations! The regression tests have now passed:

    Failed Simulations are: 
    None
    None

phys/module_physics_init.F
```diff
@@ -4653,17 +4655,52 @@ SUBROUTINE mp_init(RAINNC,SNOWNC,GRAUPELNC,config_flags,restart,warm_rain,
...
+            ccn_conc = nssl_cccn/1.225 ! set this to have correct boundary conditions
+         ENDIF
+         CALL nssl_2mom_init(nssl_params=nssl_params,ipctmp=nssl_ipconc,mixphase=0, &
+           nssl_density_on=(config_flags%nssl_density_on > 0), &
+           nssl_hail_on=config_flags%nssl_hail_on > 0, &
+           nssl_ccn_on=(config_flags%nssl_ccn_on > 0), &
+           nssl_icdx=config_flags%nssl_icdx, &
+           nssl_icdxhl=config_flags%nssl_icdxhl,ccn_is_ccna=config_flags%nssl_ccn_is_ccna)
+
+!    CASE (NSSL_2MOM)
```
islas, line 4681: @MicroTed Are the following cases now obsolete?

MicroTed: Yes, I just neglected to remove these commented lines. The associated mp_physics values (17,19,21,22) are still checked in check_a_mundo and converted to 18 with appropriate flags (along with a warning).

islas: Okay. Let's go ahead and remove these then. If you think it would be helpful for posterity the comment could be reduced to something to the effect of `! Cases 17, 19, 21, 22 superseded by 18 <extra detail if needed>`

MicroTed:
It's likely that the failure with the double precision OpenMP test could also have been resolved by setting a higher value of OMP_STACKSIZE. The default setting on linux is 4M, but is higher on Macos (with ifort) where I did my initial testing, and I think this is why it worked on the mac but not on linux. I could test this for certain if the Jenkins test be modified to set the environment value of OMP_STACKSIZE to something higher, like 8M or 16M? But it's also fine to leave it at the current solution since it works already. 

In any event, I'll add a note to the readme that for smpar or dm+sm, it may be necessary to increase that value if a seg fault occurs.

weiwangncar:
@MicroTed Thanks for documenting the solution. We would like to keep things as it is for the time being. But knowing a possible solution exists would help for later development.

phys/module_physics_init.F
MicroTed: Done

weiwangncar:
@MicroTed It looks like you didn't have any changes made for configure, is that true? But right now your configure has conflict with that in develop branch. Can you confirm?

MicroTed:
@weiwangncar  Yes, there are a couple changes to configure. They are not directly related to the microphysics update, though, so they could be split off. They solve a couple annoyances that I have run into, but it would be reasonable to move those to a separate PR.

weiwangncar:
@MicroTed I think it would be a good idea to separate the change to the configure file from the rest. Thanks!

MicroTed:
@weiwangncar  So it turned out that the conflict in configure solved the same problem, though differently. So all that is left there is a line that ensures that USENETCDFPAR is defined as 0 if it isn't 1. (I ran into a version of make that didn't do the test in Makefile correctly because USENETCDFPAR had no value.) I have merged to the latest develop branch.

weiwangncar:
@MicroTed Thanks for resolving the conflict. I would say we would like you to pull out the change in configure and make that a separate PR, since it doesn't have much to do with your MP update. I know it is a hassle, but it helps to make code change clean, easier to follow.

weiwangncar:
@MicroTed Thanks for doing that! Can you please update the list of files changed in the PR message? Also can you add a note about changing constant in module_advect_em.F? Why is that preferred for WENO?

MicroTed:
@weiwangncar  Sure, I have updated the file list -- thanks for noticing that. As for the WENO epsilon constant, do you mean to add a note to the release note section in addition to the description? Basically, epsilon is there to prevent division by zero, and it needs to be some orders of magnitude smaller than valid values of the advected scalar. The reflectivity moments are valid down to, say, 1.e-25 or so. If epsilon is at a similar order of magnitude, it will cause errors. I have been using an epsilon of 1d-40 in COMMAS and CM1 with the 3-moment scheme, and it works well.

weiwangncar:
@MicroTed I now see the sentence describing the reason for the decreased coefficient. I'm OK with it. Thanks.

doc/README.NSSLmp
```diff
@@ -0,0 +1,162 @@
...
+
+ IMPORTANT: Best results are attained using WENO (Weighted Essentially Non-Oscillatory) scalar advection option. This helps to limit oscillations at the edges of precipitation regions (i.e., sharp gradient), which in turns helps to prevent mismatches of moments that can show up as noisy reflectivity values.
+   moist_adv_opt  = 4,
+   scalar_adv_opt = 3,
+ The monotonic option (2) is less effective, but better than the default positive definite option (1)
+
+NOTE TO SMPAR or DM+SMPAR USERS: If a segmentation fault occurs, try setting the environment variable OMP_STACKSIZE to 8M or 16M (default is 4M, where M=MB). Note that this does not increase the shell stacksize limit [use 'ulimit -a unlimited' (bash) or 'unlimit stacksize' (tcsh)]
+
+CHANGES:
+June 2023 (WRF 4.5.X): Main default option change is for graupel/hail fall speed options (icdx, icdxhl; changed from 3 to 6, see below), and default maximum gr/hail droplet collection efficiencies (ehw0/ehlw0 changed from 0.5/0.75 to 0.9/0.9, see below)
```
weiwangncar, line 12: @MicroTed Can you change this to WRF 4.6?

MicroTed: Yes, got that and other instances. I also made a few small updates to the text.

weiwangncar: Thanks! I think we are ready to accept the change.

weiwangncar (approved):

dudhia (approved):
we should test this at NCAR in release testing phase.

weiwangncar:
@MicroTed We would like to test one or two NSSL MP options. What would you recommend? What is the option run in WoF at NSSL? If a user would like to use a NSSL option, what would you recommend?

dudhia:
They mention README.NSSLmp but I didn't find it. It should be in the run
directory,

weiwangncar:
@dudhia That file is in doc/ directory. But I still want to know what is commonly used.

dudhia:
I would have looked in the run directory for things related to namelist
options, especially as README.namelist refers to it, but it could also be
corrected to give the full ./doc path.

MicroTed:
It seemed logical to me to put the readme in ./doc, but that's just me.
As for what options, probably most users just use the default 2-moment (18) or the old 17 (turns off CCN concentration prediction). So a few tests could be the old 17, the 3-moment, and to turn off hail:
1. nssl_ccn_on = 0
2. nssl_3moment = 1
3. nssl_hail_on = 0 (for non-severe convection or to see the difference for severe)

The current WoFS uses mp_physics=18 (2-moment with CCN and hail), but the code is not as updated (for consistency).

dudhia:
Yes, we should just fix README.namelist to give its relative path.

PR #1881, opened 2023-06-15
New option for SLUCM to consider global distributed urban parameters

epn09:
New option for SLUCM to consider global distributed urban parameters

TYPE: new feature

KEYWORDS: SLUCM, urban parameters, anthropogenic heat

SOURCE: Do Ngoc Khanh (Tokyo Institute of Technology)

DESCRIPTION OF CHANGES:
This PR adds a new feature to WRF SLUCM by allowing consideration of spatially varying global distributed urban parameters and spatially hourly monthly varying anthropogenic heat.

ISSUE: N/A

LIST OF MODIFIED FILES: 
M       Registry/registry.dimspec
M       phys/module_sf_clm.F
M       phys/module_sf_urban.F
M       phys/noahmp
M       share/output_wrf.F

TESTS CONDUCTED: 
1. The modification has been tested and used in previous publications.
  * Initial development: Varquez, A. C. G., Nakayoshi, M., & Kanda, M. (2015). The effects of highly detailed urban roughness parameters on a sea-breeze numerical simulation. Boundary-layer meteorology, 154, 449-469. https://doi.org/10.1007/s10546-014-9985-4
  * Global extension: Khanh, D. N., Varquez, A. C., & Kanda, M. (2023). Impact of urbanization on exposure to extreme warming in megacities. Heliyon, 9, e15511. https://doi.org/10.1016/j.heliyon.2023.e15511

RELEASE NOTE: This modification add a two options (use_distributed_aerodynamics and distributed_ahe_opt) to WRF SLUCM (sf_urban_physics = 1) so that spatially varying urban morphological parameters (building height, plan area index, frontal area index, roughness length for momentum, and displacement height) can be considered.

### DIFF
--- a/Registry/Registry.EM_COMMON
+++ b/Registry/Registry.EM_COMMON
@@ -754,6 +754,7 @@ state   real   slope               ij    misc          1     -     rdu     "SLOP
 state   real   slp_azi             ij    misc          1     -     rdu     "SLP_AZI"       "ELEVATION SLOPE AZIMUTH"  "rad"
 state   real   shdmax              ij    misc          1     -     i012rhd=(interp_mask_field:lu_index,iswater)u=(copy_fcnm)   "SHDMAX"        "ANNUAL MAX VEG FRACTION" ""
 state   real   shdmin              ij    misc          1     -     i012rhd=(interp_mask_field:lu_index,iswater)u=(copy_fcnm)   "SHDMIN"        "ANNUAL MIN VEG FRACTION" ""
+state   real   shdavg              ij    misc          1     -     i012rhd=(interp_mask_field:lu_index,iswater)u=(copy_fcnm)   "SHDAVG"        "ANNUAL AVG VEG FRACTION" ""
 state   real   snoalb              ij    misc          1     -     i012rhd  "SNOALB"        "ANNUAL MAX SNOW ALBEDO IN FRACTION" ""
 state   real   toposoil            ij    misc          1     -     i12     "SOILHGT"       "ELEVATION OF LSM DATA"  "m"
 state   real   landusef            iuj   misc          1     Z     i012rdu   "LANDUSEF"      "LANDUSE FRACTION BY CATEGORY"  ""
@@ -807,13 +808,19 @@ state    real   DZR             l        em      -         Z     r        "DZR" 
 state    real   DZB             l        em      -         Z     r        "DZB"            "THICKNESSES OF WALL LAYERS"                      "m"
 state    real   DZG             l        em      -         Z     r        "DZG"            "THICKNESSES OF ROAD LAYERS"                      "m"
 state    real   URB_PARAM      i{urb}j  misc      1        -     i1       "URB_PARAM" "NUDAPT_NBSD Urban Parameters" "parameter"
-state    real   LP_URB2D       ij       misc      1        -     ir       "BUILD_AREA_FRACTION" "BUILDING PLAN AREA DENSITY"   "dimensionless"
+state    real   LP_URB2D       ij       misc      1        -     i01r     "BUILD_AREA_FRACTION" "BUILDING PLAN AREA DENSITY"   "dimensionless"
 state    real   HI_URB2D       i{uhi}j  misc      1        Z     ir       "HEIGHT_HISTOGRAMS" "DISTRIBUTION OF BUILDING HEIGHTS" "dimensionless"
 state    real   LB_URB2D       ij       misc      1        -     ir       "BUILD_SURF_RATIO" "BUILDING SURFACE AREA TO PLAN AREA RATIO" "dimensionless"
 state    real   HGT_URB2D      ij       misc      1        -     ir       "BUILD_HEIGHT" "AVERAGE BUILDING HEIGHT WEIGHTED BY BUILDING PLAN AREA" "m"
-state    real   MH_URB2D       ij       misc      1        -     ir       "MH_URB2D"  "Mean Building Height" "m"
+state    real   MH_URB2D       ij       misc      1        -     i01r     "MH_URB2D"  "Mean Building Height" "m"
 state    real   STDH_URB2D     ij       misc      1        -     ir       "STDH_URB2D" "Standard Deviation of Building Height" "m2"
 state    real   LF_URB2D       i{udr}j  misc      1        Z     ir       "LF_URB2D"   "Frontal Area Index" "dimensionless"
+
+state    real   ZD_URB2D       ij       misc        1        -     i1     "ZD_URB2D"   "Zero-plane Displacement" "m"
+state    real   Z0_URB2D       ij       misc        1        -     i01r   "Z0_URB2D"   "Roughness length for momentum"  "m"
+state    real   LF_URB2D_S     ij       misc        1        -     i01r   "LF_URB2D_S" "Frontal area index (no wind directional dependency)" ""
+# AHE with month and hour dimension flattened to one dimension, Jan = (0:23), Feb = (24:47)
+state    real   AHE            i{m_hr}j  misc       1        -     i01r   "AHE"        "Anthropogenic heat emission" "W m-2"
 
 # lsm State Variables
 
@@ -2515,6 +2522,8 @@ rconfig   integer ishallow                namelist,physics      1             0 
 rconfig   real    convtrans_avglen_m      namelist,physics      1             30      rh    "convtrans_avglen_m"  "averaging time for convective transport output variables (minutes)"      ""
 rconfig   integer num_land_cat            namelist,physics	1            21       -      "num_land_cat"                  ""      ""
 rconfig   integer use_wudapt_lcz          namelist,physics      1             0       -      "use_wudapt_lcz"                ""      ""
+rconfig   logical use_distributed_aerodynamics          namelist,physics      1             .false.       rh      "use_distributed_aerodynamics"                ""      ""
+rconfig   integer distributed_ahe_opt     namelist,physics  1            1        rh     "distributed_ahe_opt" "AHE handling: 1=add to first level temperature tendency, 2=add to surface sensible heat flux" ""
 rconfig   integer num_soil_cat            namelist,physics	1            16       -      "num_soil_cat"                  ""      ""
 rconfig   integer mp_zero_out             namelist,physics	1             0       -      "mp_zero_out"  "microphysics fields set to zero  0=no action taken, 1=all fields but Qv, 2=all fields including Qv"      "flag"
 rconfig   real mp_zero_out_thresh         namelist,physics	1          1.e-8      -      "mp_zero_out_thresh"  "minimum threshold for non-Qv moist fields, below are set to zero"  "kg/kg"
--- a/dyn_em/module_initialize_real.F
+++ b/dyn_em/module_initialize_real.F
@@ -617,6 +617,15 @@ integer::oops1,oops2
          END IF
       END IF
 
+      IF (config_flags%use_distributed_aerodynamics) THEN
+        CALL wrf_message('Adding zero-plane displacement height to topography')
+        DO j = jts, MIN(jde - 1, jte)
+        DO i = its, MIN(ide - 1, ite)
+          IF (grid%zd_urb2d(i, j) > 0) grid%ht_gc(i, j) = grid%ht_gc(i, j) + grid%zd_urb2d(i, j)
+        END DO
+        END DO
+      END IF
+
       !  Is there any vertical interpolation to do?  The "old" data comes in on the correct
       !  vertical locations already.
 
@@ -1330,6 +1339,11 @@ integer::oops1,oops2
                                 ims , ime , jms , jme , kms , kme , &
                                 its , ite , jts , jte , kts , kte )
 
+         CALL monthly_avg ( grid%greenfrac , grid%shdavg , &
+                                ids , ide , jds , jde , kds , kde , &
+                                ims , ime , jms , jme , kms , kme , &
+                                its , ite , jts , jte , kts , kte )
+
          !  The model expects the green-ness and vegetation fraction values to be in percent, not fraction.
 
          DO j = jts, MIN(jte,jde-1)
@@ -1338,6 +1352,7 @@ integer::oops1,oops2
               grid%vegfra(i,j) = grid%vegfra(i,j) * 100.
               grid%shdmax(i,j) = grid%shdmax(i,j) * 100.
               grid%shdmin(i,j) = grid%shdmin(i,j) * 100.
+              grid%shdavg(i,j) = grid%shdavg(i,j) * 100.
            END DO
          END DO
 
@@ -3089,6 +3104,16 @@ integer::oops1,oops2
 
       !  Split NUDAPT  Urban Parameters
 
+      distributed_aerodynamics_if: IF (config_flags%use_distributed_aerodynamics) THEN
+         DO j = jts , MIN(jde-1,jte)
+            DO i = its , MIN(ide-1,ite)
+              IF (grid%ivgtyp(i, j) == model_config_rec%isurban(grid%id)) THEN
+                grid%frc_urb2d(i, j) = MAX(0.1, MIN(0.9, 1 - grid%shdavg(i, j) / 100.))
+              END IF
+            END DO
+          END DO
+      ELSE
+
       IF ( ( config_flags%sf_urban_physics == 1 ) .OR. ( config_flags%sf_urban_physics == 2 ) .OR. ( config_flags%sf_urban_physics == 3 ) ) THEN
          DO j = jts , MIN(jde-1,jte)
             DO i = its , MIN(ide-1,ite)
@@ -3144,6 +3169,8 @@ integer::oops1,oops2
             END DO
          END DO
       END DO
+
+      END IF distributed_aerodynamics_if
 
       END IF
 
@@ -7921,6 +7948,34 @@ end do
       END DO
 
    END SUBROUTINE monthly_min_max
+
+!---------------------------------------------------------------------
+
+   SUBROUTINE monthly_avg ( field_in , field_avg , &
+                      ids , ide , jds , jde , kds , kde , &
+                      ims , ime , jms , jme , kms , kme , &
+                      its , ite , jts , jte , kts , kte )
+
+      IMPLICIT NONE
+
+      INTEGER , INTENT(IN)        :: ids , ide , jds , jde , kds , kde , &
+                                     ims , ime , jms , jme , kms , kme , &
+                                     its , ite , jts , jte , kts , kte
+
+      REAL , DIMENSION(ims:ime,12,jms:jme) , INTENT(IN)  :: field_in
+      REAL , DIMENSION(ims:ime,   jms:jme) , INTENT(OUT) :: field_avg
+
+      !  Local vars
+
+      INTEGER :: i , j
+
+      DO j = jts , MIN(jde-1,jte)
+        DO i = its , MIN(ide-1,ite)
+          field_avg(i, j) = SUM(field_in(i, :, j)) / 12
+        END DO
+      END DO
+
+   END SUBROUTINE monthly_avg
 
 !---------------------------------------------------------------------
 
--- a/phys/module_pbl_driver.F
+++ b/phys/module_pbl_driver.F
@@ -103,6 +103,9 @@ CONTAINS
                ,tke_adv,diss_adv,tpe_adv                   &
                ,pr_pbl,el_pbl                              &
                ,wu_tur,wv_tur,wt_tur,wq_tur                &
+! variables added for AHE
+               , gmt, xtime, julday, julyr, ahe &
+               , distributed_ahe_opt &
 ! variables  for GBM PBL
                ,exch_tke, rthraten                         &
                ,a_e_bep,b_e_bep,dlg_bep,dl_u_bep           &
@@ -199,6 +202,7 @@ CONTAINS
    USE module_bl_fogdes
    USE module_wind_fitch
 #endif
+   use module_ra_gfdleta, only: cal_mon_day
 
    !  This driver calls subroutines for the PBL parameterizations.
    !
@@ -610,6 +614,11 @@ CONTAINS
    REAL,       DIMENSION( ims:ime, kms:kme, jms:jme ),            &
                INTENT(OUT)    ::                          EL_PBL
 
+   REAL, INTENT(IN) :: gmt, xtime
+   INTEGER,  INTENT(IN) :: julday, julyr
+   REAL, OPTIONAL, DIMENSION( ims:ime, 0:287, jms:jme ), INTENT(IN)  :: ahe
+   INTEGER, INTENT(IN) :: distributed_ahe_opt
+
    REAL ,                             INTENT(IN   )  ::  u_frame, &
                                                          v_frame
 !
@@ -820,6 +829,7 @@ CONTAINS
    integer iu_bep,iurb,idiff
    real seamask,thsk,zzz,unew,vnew,tnew,qnew,umom,vmom
    REAL :: z0,z1,z2,w1,w2
+   INTEGER :: ihour, jmonth, jday
 !
 ! FASDAS
 !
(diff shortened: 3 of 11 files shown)


epn09:
This PR goes in pair with https://github.com/NCAR/noahmp/pull/96. Without it, the code will not compile.

weiwangncar:
@epn09 The compilation for the code has failed.
@cenlinhe Where is 'noahmp#96'?

cenlinhe:
I think @epn09 also modifies the Noah-MP driver part of the code, but the way of submitting this PR to include the Noah-MP submodule update may not be correct, which may be the cause for the regression test to fail.

cenlinhe:
I will come back to this PR later for a more detailed review when I have more time.

cenlinhe:
@tslin2, you may also want to review this urban code which is also related to Noah-MP PR I assigned you as a reviewer. Note that this is just for Noah-MP v4.5 not the latest refactored code.

tslin2:
> @tslin2, you may also want to review this urban code which is also related to Noah-MP PR I assigned you as a reviewer. Note that this is just for Noah-MP v4.5 not the latest refactored code.

okay, will take a look.

epn09:
@cenlinhe May I ask if there's any update?

cenlinhe:
This PR requires an update from Noah-MP driver part for v4.5 first and then connect the WRF update with the updated Noah-MP commit from Noah-MP GitHub in a new PR. I will need to update the Noah-MP GitHub commit first and then probably have to re-submit a PR with the updated Noah-MP commit for this.

Registry/Registry.EM_COMMON
```diff
@@ -807,14 +808,20 @@ state    real   DZR             l        em      -         Z     r        "DZR"
...
 state    real   LB_URB2D       ij       misc      1        -     ir       "BUILD_SURF_RATIO" "BUILDING SURFACE AREA TO PLAN AREA RATIO" "dimensionless"
 state    real   HGT_URB2D      ij       misc      1        -     ir       "BUILD_HEIGHT" "AVERAGE BUILDING HEIGHT WEIGHTED BY BUILDING PLAN AREA" "m"
-state    real   MH_URB2D       ij       misc      1        -     ir       "MH_URB2D"  "Mean Building Height" "m"
+state    real   MH_URB2D       ij       misc      1        -     i01r     "MH_URB2D"  "Mean Building Height" "m"
 state    real   STDH_URB2D     ij       misc      1        -     ir       "STDH_URB2D" "Standard Deviation of Building Height" "m2"
 state    real   LF_URB2D       i{udr}j  misc      1        Z     ir       "LF_URB2D"   "Frontal Area Index" "dimensionless"
 
+state    real   ZD_URB2D       ij       misc        1        -     i1     "ZD_URB2D"   "Zero-plane Displacement" "m"
+state    real   Z0_URB2D       ij       misc        1        -     i01r   "Z0_URB2D"   "Roughness length for momentum"  "m"
+state    real   LF_URB2D_S     ij       misc        1        -     i01r   "LF_URB2D_S" "Frontal area index (no wind directional dependency)" ""
```
cenlinhe, line 821: @epn09 what is the difference between LF_URB2D_S and LF_URB2D? Why do you need to create this new variable instead of using the existing LF_URB2D?

Registry/Registry.EM_COMMON
```diff
@@ -807,14 +808,20 @@ state    real   DZR             l        em      -         Z     r        "DZR"
...
 state    real   URB_PARAM      i{urb}j  misc      1        -     i1       "URB_PARAM" "NUDAPT_NBSD Urban Parameters" "parameter"
-state    real   LP_URB2D       ij       misc      1        -     ir       "BUILD_AREA_FRACTION" "BUILDING PLAN AREA DENSITY"   "dimensionless"
+state    real   LP_URB2D       ij       misc      1        -     i01r     "BUILD_AREA_FRACTION" "BUILDING PLAN AREA DENSITY"   "dimensionless"
 state    real   HI_URB2D       i{uhi}j  misc      1        Z     ir       "HEIGHT_HISTOGRAMS" "DISTRIBUTION OF BUILDING HEIGHTS" "dimensionless"
 state    real   LB_URB2D       ij       misc      1        -     ir       "BUILD_SURF_RATIO" "BUILDING SURFACE AREA TO PLAN AREA RATIO" "dimensionless"
 state    real   HGT_URB2D      ij       misc      1        -     ir       "BUILD_HEIGHT" "AVERAGE BUILDING HEIGHT WEIGHTED BY BUILDING PLAN AREA" "m"
-state    real   MH_URB2D       ij       misc      1        -     ir       "MH_URB2D"  "Mean Building Height" "m"
+state    real   MH_URB2D       ij       misc      1        -     i01r     "MH_URB2D"  "Mean Building Height" "m"
```
cenlinhe, line 815: @dudhia What is the difference between "i" and "i01"?

dyn_em/module_initialize_real.F
```diff
@@ -3089,6 +3104,16 @@ SUBROUTINE init_domain_rk ( grid &
 
       !  Split NUDAPT  Urban Parameters
 
+      distributed_aerodynamics_if: IF (config_flags%use_distributed_aerodynamics) THEN
```
cenlinhe, line 3107: need to add config_flags%sf_urban_physics == 1 in the if-statement too.

dyn_em/module_initialize_real.F
```diff
@@ -3089,6 +3104,16 @@ SUBROUTINE init_domain_rk ( grid &
...
+         DO j = jts , MIN(jde-1,jte)
+            DO i = its , MIN(ide-1,ite)
+              IF (grid%ivgtyp(i, j) == model_config_rec%isurban(grid%id)) THEN
```
cenlinhe, line 3110: need to also include LCZ type in the if-statement for urban pixels to allow working with LCZ capability.

dyn_em/module_initialize_real.F
```diff
@@ -3089,6 +3104,16 @@ SUBROUTINE init_domain_rk ( grid &
...
+                grid%frc_urb2d(i, j) = MAX(0.1, MIN(0.9, 1 - grid%shdavg(i, j) / 100.))
```
cenlinhe, line 3111: not sure if this is a good way to estimate urban fraction (=1-vegfrac), but may be OK for now.

phys/module_pbl_driver.F
```diff
@@ -2203,6 +2213,26 @@ SUBROUTINE pbl_driver(                                          &
               ,ITS=its,ITE=ite,JTS=jts,JTE=jte,KTS=kts,KTE=kte)
           ENDIF
 
+
+     IF (PRESENT(ahe)) THEN
```
cenlinhe, line 2217: @epn09 This may not be the best place to add this anthropogenic heating flux. First, if putting it in this pbl driver module, should it be added to a place after land surface model and urban scheme is called? Second, it may be better to include this to the SLUCM code module instead of pbl driver to avoid messing up with other physics.

phys/module_sf_noahdrv.F
```diff
@@ -1416,6 +1423,10 @@ SUBROUTINE lsm(DZ8W,QV3D,P8W3D,T3D,TSK,                      &
             if (I.eq.73.and.J.eq.125)THEN
                check = 1
             end if
+! Distributed aerodynamics
+            lf_urb_s = lf_urb2d_s(I, J)
+            z0_urb = z0_urb2d(I, J)
+            vegfrac = vegfra(I, J) / 100
```
cenlinhe, line 1429: double check consistency for the vegfrac used in urban fraction (=1-vegfrac) and the original vegfra used in Noah.

phys/module_sf_noahdrv.F
```diff
@@ -3721,7 +3743,12 @@ SUBROUTINE lsm_mosaic(DZ8W,QV3D,P8W3D,T3D,TSK,                      &
       
                 !  UTYPE_URB = UTYPE_URB2D(I,J) !urban type (low, high or industrial)
                 !  this need to be changed in the mosaic danli
-                IF (use_wudapt_lcz == 1) THEN
+                IF (use_distributed_aerodynamics) THEN
+                  IF (IVGTYP(I, J) == ISURBAN) THEN
+                    UTYPE_URB = 2
+                    FRC_URB2D(I, J) = MAX(0.1, MIN(0.9, 1 - shdavg(I, J) / 100.))
```
cenlinhe, line 3749: should this already be done above in the initialization and real part?

phys/module_sf_urban.F
```diff
@@ -783,6 +804,25 @@ SUBROUTINE urban(LSOLAR,                                           & ! L
    endif
    if(alhoption==1) ALH = ALH*alhdiuprf(tloc2)*alhseason(Kalh)
 
+   IF (distributed_aerodynamics_option) THEN
```
cenlinhe, line 807: also need to add a check for cases where distributed_aerodynamics_option is true but the input variables/data are not provided (i.e., MH_URB <=0 or LP_URB<=0).

phys/module_sf_urban.F
```diff
@@ -1319,6 +1375,23 @@ SUBROUTINE urban(LSOLAR,                                           & ! L
        DRGDTB=DRGDTB1+DRGDTB2
        DRGDTG=DRGDTG1+DRGDTG2
 
+       IF (distributed_aerodynamics_option) THEN
```
cenlinhe, line 1378: @epn09 It seems you replaced original TCP related quantity with TA, why is this? based on assumptions for your distributed parameter generation?

phys/module_sf_urban.F
```diff
@@ -1583,15 +1671,25 @@ SUBROUTINE urban(LSOLAR,                                           & ! L
    PSIX10 = ALOG(10./Z0) - PSIM10
    PSIT10 = ALOG(10./Z0H) - PSIH10
 
+   IF (distributed_aerodynamics_option .and. Z0C > 10.) THEN
+     U10 = 0.
```
cenlinhe, line 1675: @epn09 not sure if this is a good assumption. maybe still use the logscale decay? Is this assumption related to how you treat the roughness from your new input parameters?

phys/module_sf_urban.F
```diff
@@ -2647,6 +2756,14 @@ SUBROUTINE urban_var_init(ISURBAN, TSURFACE0_URB,TLAYER0_URB,TDEEP0_URB,IVGTYP,
 !m
 !FS   FRC_URB2D(I,J)=0.
       UTYPE_URB2D(I,J)=0
+
+      distributed_aerodynamics_check: IF (distributed_aerodynamics_option) THEN
+        IF (IVGTYP(I, J) == ISURBAN) THEN
+          UTYPE_URB2D(I, J) = 2
```
cenlinhe, line 2762: @epn09  your distributed parameters are only for urban type = 2?

cenlinhe:
I have reviewed this PR and have provided a few comments and questions in some code lines. This new capability of reading in and using global spatially varying distributed urban parameters are very important. However, it seems that the current implementation in this PR has a few limitations (correct me if I am wrong): 
(1) requires specific input urban parameter data from the developer's group. Did the develop also submit a PR to WPS to include their data in the WRF system? Otherwise, it is very hard for other users to use this capability. 
(2) The current implementation seems incompatible with the existing I/O capability from using NUDAPT parameter data. However, these two I/O systems should be similar but of course reading in different datasets and variables. Is there a way to synthesize these two to be more general? 
(3) This update does not work with LCZ urban type, which may be OK for now. 
(4) This update seems only be applicable for traditional urban type = 2 based on the comment I provided in one line above (maybe I misunderstood it). This strongly limits the code application.

Registry/Registry.EM_COMMON
weiwangncar: @cenlinhe Here letter i represents input data. The number 0 means it is part of wrfinput, output from real, and input to the model, and 1 is the input to real program or the met_em output data.

Registry/Registry.EM_COMMON
epn09: `S` in `LF_URB2D_S` means "single". If my understanding is right, `LF_URB2D` has 4 values for 4 wind direction, but our dataset only have one value for all wind direction. So, I created this new variable.

phys/module_pbl_driver.F
epn09: @cenlinhe The reason we didn't add AH flux to the urban scheme is because the construction of the AH dataset did not assume that AH comes from only urban. 

* Energy loss are distributed equally for all grids within a country. 
* There are strong point sources of AH (power station, steel mills) that do not necessarily belong to urban grids.

What's your opinion on this?

The full paper is available here (https://www.nature.com/articles/s41597-021-00850-w).

phys/module_sf_urban.F
epn09: @cenlinhe  Within this if statement, some default values for the parameters when they are not available are set.
```fortran
ZR = MAX(MH_URB, 3.5)
Z0C = MAX(Z0_URB, 0.1)
lambda_p = MAX(0.05, MIN(1.0, LP_URB))
lambda_f = MAX(0.05, MIN(1.0, LF_URB_S))
```

phys/module_sf_urban.F
epn09: @cenlinhe In my understanding, the difference between urban types (low intensity, high intensity, LCZ 1 to 10) in WRF right now are only the differences in the parameters in URBPARM.TBL. When we input a spatial map of these urban parameters from outside, the value of urban type doesn't matter any more. So, it's just set to 2 for no special reasons.

phys/module_sf_noahdrv.F
epn09: @cenlinhe To be honest I also don't know why I must do this. But as you can see, the code just below this line for wudapt also do the same thing for `FRC_URB2D` so I just follow the way the existing code was written.

```fortran
                  IF(UTYPE_URB==1) FRC_URB2D(I,J)=1.
                  IF(UTYPE_URB==2) FRC_URB2D(I,J)=0.99
                  IF(UTYPE_URB==3) FRC_URB2D(I,J)=1.00
                  IF(UTYPE_URB==4) FRC_URB2D(I,J)=0.65
                  IF(UTYPE_URB==5) FRC_URB2D(I,J)=0.7
                  IF(UTYPE_URB==6) FRC_URB2D(I,J)=0.65
                  IF(UTYPE_URB==7) FRC_URB2D(I,J)=0.3
                  IF(UTYPE_URB==8) FRC_URB2D(I,J)=0.85
                  IF(UTYPE_URB==9) FRC_URB2D(I,J)=0.3
                  IF(UTYPE_URB==10) FRC_URB2D(I,J)=0.55
                  IF(UTYPE_URB==11) FRC_URB2D(I,J)=1.
```

tslin2: this part is the issue in mosaic where the 2d spatially urban fraction is not used. So after the first timestep, the urban fraction is different compared to the initial 2d data.

phys/module_sf_urban.F
epn09: @cenlinhe Currently in this parameter database that we have, there is no grid with z0 higher than 10 m, so this if statement actually will never be executed. But the treatment of grid with z0 > 10 m (if exists) is something we also don't know how to properly handle because in my understanding log-law is not valid for z < z0. So if z = 10 m and z0 > 10 m, I don't really know what should be done. In the mean time, this if statement can be removed. Please let me know your opinion.

epn09:
@cenlinhe Thank you very much for reviewing. Some of your comments need code modification so I'll address them separately. But below is my response for the 4 points you raised.

> (1) requires specific input urban parameter data from the developer's group. Did the develop also submit a PR to WPS to include their data in the WRF system? Otherwise, it is very hard for other users to use this capability.

No, I haven't submit a PR. Please let me know the procedure and I'll do it promptly.

> (2) The current implementation seems incompatible with the existing I/O capability from using NUDAPT parameter data. However, these two I/O systems should be similar but of course reading in different datasets and variables. Is there a way to synthesize these two to be more general? 

In my understanding, NUDAPT only change the urban parameters but do not change the physics. However, our scheme not only change the urban parameters but also represent the urban canopy as a whole and disregard the individual effects of the roof and canyon. More details can be found here (http://www.ide.titech.ac.jp/~kandalab/download/WRF_URBAN/add_improvements.html). Please ignore section II of the link because it's obsolete thanks to the implementation of WRF mosaic option.

So, at the moment, I don't have an idea to synthesize the two approaches.

> (3) This update does not work with LCZ urban type, which may be OK for now. 

Yes. And I can't think of a valid use case of combining our approach with LCZ for now, even though both aims at the same goal of making more realistic urban representation in WRF.
* LCZ aims at providing more urban categories (from the default 3),
* Our approach uses spatial maps of urban parameters (in other words, there is no urban category anymore).

Of course, the two approach can be compared against each other.

> (4) This update seems only be applicable for traditional urban type = 2 based on the comment I provided in one line above (maybe I misunderstood it). This strongly limits the code application.

As mentioned in (3) and replied to your comment. Urban types play no role in this scheme. Type 2 is set in the code for no special reason.

Registry/Registry.EM_COMMON
cenlinhe: OK, I see. This makes sense.

Registry/Registry.EM_COMMON
cenlinhe: @weiwangncar If we do not specify "0" and/or "1" after "i", does it mean it can be in any input files or by default assume it is from wrfinput? For example, for variables like HGT_URB2D, they are only specified as "i" instead of "i01".

dyn_em/module_initialize_real.F
cenlinhe: OK, this makes sense, at least for now. more tests with the impact of shdavg vs shdmax are needed in the future.

phys/module_pbl_driver.F
cenlinhe: OK, I see. This makes sense to me, as the AH is not only for urban AH. power station may also belong to urban?

phys/module_sf_noahdrv.F
cenlinhe: I will need to take a more detailed look at this. These FRC_URB2D assignments for each LCZ are supposed to be reading from the URBPARM_LCZ.TBL table values instead of hard-coded.

phys/module_sf_urban.F
cenlinhe: OK, I see.

phys/module_pbl_driver.F
epn09: @cenlinhe Power station may be in either rural or urban. When constructing the AH dataset we did not make that distinction.

phys/module_sf_urban.F
cenlinhe: I do not have a good solution to this either. The default treatment is using logscale decay. For Z0C > 10, the log law should give a relatively small value (close to 0) at z=10. Maybe for now, just follow the default treatment.

phys/module_sf_urban.F
cenlinhe: OK

phys/module_sf_noahdrv.F
cenlinhe: @epn09 @tslin2 This original part of re-assigning FRC_URB2D is wrong. FRC_URB2D is initialized in module_physics_init.F and should only be read in as a intent(IN) variable without modifying. This has been done in both Noah and NoahMP driver. So this part should be removed from lsm_mosaic module.
@tslin2 you may want to test if removing this would solve your mosaic error.

tslin2: @cenlinhe 
Yes, I have tested already; it has the 2d spatial FRC_URB2D from the initial file

epn09: @cenlinhe @tslin2 So what's the conclusion here? Should I remove the `FRC_URB2D` calculation here, or should it be treated in a separate PR?

tslin2: maybe @DanLi-BU can also comment on this.

Thanks

phys/module_sf_urban.F
epn09: @cenlinhe According to the materials that I read, in this scheme, canopy temperature is not calculated (hence, `TCP` and `TC` are unused). Heat from all surfaces (roof, wall, roads) are released to the upper atmosphere directly. Therefore, `TA` is used.

References:
* Section I of http://www.ide.titech.ac.jp/~kandalab/download/WRF_URBAN/add_improvements.html
* https://link.springer.com/article/10.1007/s10546-004-7956-x

dyn_em/module_initialize_real.F
epn09: @cenlinhe  This is fixed.

phys/module_sf_urban.F
epn09: @cenlinhe This is fixed. The if Z0C > 10 is removed.

dyn_em/module_initialize_real.F
epn09: @cenlinhe I added if-statement and error messages to `share/module_check_a_mundo.F`.

epn09:
@cenlinhe I think I've addressed all of your comments. Please let me know if you have any further concern.

phys/module_sf_noahdrv.F
DanLi-BU: I haven't read the WRF code this closely for such a long time but it's good to be back in the coding world. I don't think I added this line of `FRC_URB2D(I, J) = MAX(0.1, MIN(0.9, 1 - shdavg(I, J) / 100.))` when I implemented the mosaic and I agree FRC_URB2D should not be changed (unless there is a strong justification).

tslin2: FRC_URB2D(I, J) = MAX(0.1, MIN(0.9, 1 - shdavg(I, J) / 100.))

is proposed by @epn09 for new option to consider global distributed urban parameteres

tslin2: @DanLi-BU you can also review if you like

dyn_em/module_initialize_real.F
DanLi-BU: @tslin2 @epn09 OK, let me comment here since this is probably the first place where it showed up. I think I understood what Khanh is trying to do here. If there is no other data source for FRC_URB2D, one can come up with a few ways to assign FRC_URB2D values. It used to be that FRC_URB2D is prescribed in the URBPARM.TBL (e.g., 0.5 for low intensity residential urban etc). With NLCD data, FRC_URB2D is actually estimated from NLCD and is no longer using prescribed values from URBPARM.TBL. Now this part of code is trying to assign FRC_URB2D based on vegetation fraction, which is not a terrible idea. But this will limit the compatibility of this code with potentially new land use/land cover data that can provide FRC_URB2D. 

Khanh (first of all, congratulations on becoming a doctor, way to go), regarding ``If a WRF grid will all MODIS grids classified as urban, the WRF grid will have LANDUSEF(urban) = 100``, this is not true actually.  LANDUSEF is always computed based on land use land cover data in WPS, and stored and passed to WRF. But it is not used when the mosaic approach is not enacted. In a somewhat similar way, FRC_URB2D can be also computed using land use land cover data (especially with high resolution data like NLCD) and it is actually used by WRF (at least before LCZ was introduced, I don't know the details after LCZ is introduced).

phys/module_sf_noahdrv.F
DanLi-BU: Khanh has to buy me a lot of coffee for me to review the entire code :) I'll try my best to comment here and there but I can't really promise.

cenlinhe:
@epn09 I am going to re-submit a PR with your modifications on urban part and refer back to this PR, because I need to update Noah-MP GitHub with your mods first and re-connect to the correct Noah-MP branch and commit.
In the near future, please submit a PR to WPS GitHub with your updated code that works with your new urban parameter dataset. Please also prepare the WPS binary data format for your dataset and test it in WPS, so that the WRF team could directly incorporate your dataset into WRF/WPS database for users to download.

Registry/Registry.EM_COMMON
cenlinhe: @weiwangncar Since these variables are optional (i.e., users may not provide these in their geo_em input file), I think it should be set to "i" instead of "i01". Otherwise, WRF will look for these variables in wrfinput or met_em files, and report errors? Is this correct?

dyn_em/module_initialize_real.F
epn09: @DanLi-BU Thanks for your note about NLCD. Well-noted.

Registry/Registry.EM_COMMON
weiwangncar: @cenlinhe Sorry I missed these comments. If you just have an 'i', that variable is only appearing in wrfinput - whether it has value will depend on what real does. If there is no input to real, the array may be all zero. If the data comes from geogrid and metgrid, it should be set to i01.

weiwangncar:
This has been replaced by PR-[1986](https://github.com/wrf-model/WRF/pull/1986).

PR #1890, opened 2023-06-27
loop permute, to help better memory access, vectorization

juntangc:
Purpose: this patch rearranges loop order (permute) to help memory access; it also breaks down larger loops into smaller ones to allow partial vectorization (GCC)

TYPE: enhancement

KEYWORDS: loop permute, vectorization

SOURCE: Jun Tang (Amazon)

DESCRIPTION OF CHANGES:
Problem:
GCC is not able to permute the loop so that the inner loop is along fast dimension of the arrays and not able to vectorize loops with conditional statement

Solution:
This patch allows memory access in the inner loop along the faster dimension of the data arrays; this patch also breaks down larger loops with conditional statement to allow partial vectorization.

ISSUE: N/A

LIST OF MODIFIED FILES: list of changed files (use git diff --name-status master to get formatted list)
M dyn_em/module_advect_em.F
M dyn_em/module_big_step_utilities_em.F
M dyn_em/module_em.F
M phys/module_ra_rrtmg_lw.F
M phys/module_ra_rrtmg_sw.F
M phys/module_radiation_driver.F

TESTS CONDUCTED:

    Do mods fix problem? How can that be demonstrated, and was that test conducted?
    Are the Jenkins tests all passing?
    I have seen 2-3% improvement in performance for CONUS 2.5km benchmark (GCC).

RELEASE NOTE: Include a stand-alone message suitable for the inclusion in the minor and annual releases. A publication citation is appropriate.

### DIFF
--- a/dyn_em/module_advect_em.F
+++ b/dyn_em/module_advect_em.F
@@ -4146,8 +4146,8 @@ SUBROUTINE advect_scalar   ( field, field_old, tendency,    &
 
    IF( (config_flags%open_ys) .and. (jts == jds) ) THEN
 
-       DO i = i_start, i_end
        DO k = kts, ktf
+       DO i = i_start, i_end
          vb = MIN( 0.5*(rv(i,k,jts)+rv(i,k,jts+1)), 0. )
          tendency(i,k,jts) = tendency(i,k,jts)                     &
                - rdy*(                                             &
@@ -4162,8 +4162,8 @@ SUBROUTINE advect_scalar   ( field, field_old, tendency,    &
 
    IF( (config_flags%open_ye) .and. (jte == jde)) THEN
 
-       DO i = i_start, i_end
        DO k = kts, ktf
+       DO i = i_start, i_end
          vb = MAX( 0.5*(rv(i,k,jte-1)+rv(i,k,jte)), 0. )
          tendency(i,k,j_end) = tendency(i,k,j_end)                   &
                - rdy*(                                               &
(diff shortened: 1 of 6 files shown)


weiwangncar:
The regression test results:

    Failed Simulations are: 
    None
    None

weiwangncar:
We will need to test if the code performs similarly on our systems.

juntangc:
Sure. Please test the GCC build.

dudhia:
This should be targeted for develop and not master

weiwangncar:
@juntangc Any idea if you get bit-for-bit outcome, perhaps with 'configure -d' compilation?

weiwangncar:
@dudhia Can you take a closer look at this PR and comment?

dudhia:
@weiwangncar I agree with your comment that we need a bit-for-bit test before we can allow this.

juntangc:
> @juntangc Any idea if you get bit-for-bit outcome, perhaps with 'configure -d' compilation?

@weiwantncar @dudhia Thanks for reviewing this.  This revision has bit-by-bit matching on WRF output for conus12km and conus 2.5km models.

dudhia:
Thanks. Could this information be added to the tests conduced section of the PR?

juntangc:
I have added a short description under the test section - https://github.com/wrf-model/WRF/pull/1890#issue-1777842539. 

I can add some model statics and snapshots if needed. It might take some time because the comparison was done a while ago.

weiwangncar:
@juntangc When you say the outputs are bit-for-bit, are the model compiled with full optimization? Can you comment on its potential impact with other compilers, such as Intel's? Some statistics for performance should be helpful to post. Thanks.

juntangc:
The comparison is done with GCC compiler with and without this patch using default/full optimization.  The performance impact with Intel compiler is very small.  We compare the optimization reports from Intel compiler and GCC to identify these loops; those loops are vectorized by Intel compiler because it can permute/exchange the inner loop with outer loop according to memory layout, but they can't be vectorized because GCC can't permute/exchange the loops.

dudhia (approved):
I will assume results are unaffected based on tests reported.

weiwangncar:
@juntangc Tests using gfortran on NCAR's Derecho didn't show significant performance gain. But we also confirmed that the change does produce bit-for-bit outcome when compiled with -d option. 
@mgduda Any other comments on this PR?

weiwangncar (approved):

PR #1896, opened 2023-07-08
CMake Build

islas:
TYPE: new feature

KEYWORDS: CMake, build, make

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
The current WRF build system is fragile with many pitfalls making it difficult for users to build & add to it without perpetuating existing problems. Many options exist across various layers of files, languages, and option control-flow.

Solution:
*This requires CMake version 3.20 or newer*
A redesign of the build system from the ground up, maintaining the interfacing feel and knowledge accumulated in `arch/configure.defaults`. Condense option selection and control to single locations and as best as possible reduce the complexity of this control.

This will be a work in progress as gaps are identified in reproducing the full functionality of the makefile build system. Currently only `em_real` and `em_ideal` have limited supported

Brief how to use:
As this is a work in progress, the original `configure` and `compile` scripts have been left as-is. Alongside them are now `configure_new` and `compile_new` which walk a user through a slightly similar experience of configuring & compiling WRF.

A simple usage example would be:
```bash
# Ensure you have cmake 3.20+ and configuration environment set up
./configure_new
# Follow prompts to select configuration
./compile_new [-j N]
```

Notable differences are :
* Submodule code must be checked out beforehand and is not checked out during the compile process
* Stanzas presented to a user are only those for which the compiler exists in the current environment
* `!!` warnings appear for subconfigurations (MPI) that would not be supported in the current environment
* DM/SM selection is now done after selecting a base configuration rather than an individual configuration # within a family of compilers
* Compilation via `compile_new` does not take target to build as an argument - parallel `-j N` jobs still supported
* Users do not need to set `NETCDF` or `LD_LIBRARY_PATH` variables
* Base binaries do not have `.exe` extension, but symlinks are provided
* Binaries, test setups, and everything else generated from compilation is copy-placed (not softlinked) to a separate location - default is `./install`. This means the equivalent `test/em_real/wrf.exe` is now at `install/test/em_real/wrf.exe`

LIST OF MODIFIED FILES: 
A       CMakeLists.txt
A       arch/configure_reader.py
A       chem/CMakeLists.txt
A       cleanCMake.sh
A       cmake/c_preproc.cmake
A       cmake/confcheck.cmake
A       cmake/gitinfo.cmake
A       cmake/m4_preproc.cmake
A       cmake/modules/FindJasper.cmake
A       cmake/modules/FindRPC.cmake
A       cmake/modules/FindnetCDF-Fortran.cmake
A       cmake/modules/FindnetCDF.cmake
A       cmake/modules/FindpnetCDF.cmake
A       cmake/printOption.cmake
A       cmake/target_copy.cmake
A       cmake/template/WRFConfig.cmake.in
A       cmake/template/arch_config.cmake
A       cmake/template/commit_decl.cmake
A       cmake/wrf_case_setup.cmake
A       cmake/wrf_get_version.cmake
A       compile_new
A       confcheck/CMakeLists.txt
A       configure_new
A       dyn_em/CMakeLists.txt
A       external/CMakeLists.txt
A       external/RSL_LITE/CMakeLists.txt
A       external/atm_ocn/CMakeLists.txt
A       external/esmf_time_f90/CMakeLists.txt
A       external/fftpack/fftpack5/CMakeLists.txt
A       external/io_adios2/CMakeLists.txt
A       external/io_esmf/CMakeLists.txt
A       external/io_grib1/CMakeLists.txt
A       external/io_grib1/MEL_grib1/CMakeLists.txt
A       external/io_grib1/WGRIB/CMakeLists.txt
A       external/io_grib1/grib1_util/CMakeLists.txt
A       external/io_grib2/CMakeLists.txt
A       external/io_grib2/bacio-1.3/CMakeLists.txt
A       external/io_grib2/g2lib/CMakeLists.txt
A       external/io_grib2/g2lib/utest/CMakeLists.txt
A       external/io_grib_share/CMakeLists.txt
A       external/io_int/CMakeLists.txt
A       external/io_netcdf/CMakeLists.txt
A       external/io_netcdfpar/CMakeLists.txt
A       external/io_phdf5/CMakeLists.txt
A       external/io_pio/CMakeLists.txt
A       external/io_pnetcdf/CMakeLists.txt
A       external/ioapi_share/CMakeLists.txt
A       frame/CMakeLists.txt
A       inc/CMakeLists.txt
A       main/CMakeLists.txt
A       phys/CMakeLists.txt
A       share/CMakeLists.txt
A       test/em_b_wave/CMakeLists.txt
A       test/em_convrad/CMakeLists.txt
A       test/em_fire/CMakeLists.txt
A       test/em_grav2d_x/CMakeLists.txt
A       test/em_heldsuarez/CMakeLists.txt
A       test/em_hill2d_x/CMakeLists.txt
A       test/em_les/CMakeLists.txt
A       test/em_quarter_ss/CMakeLists.txt
A       test/em_real/CMakeLists.txt
A       test/em_scm_xy/CMakeLists.txt
A       test/em_seabreeze2d_x/CMakeLists.txt
A       test/em_squall2d_x/CMakeLists.txt
A       test/em_squall2d_y/CMakeLists.txt
A       test/em_tropical_cyclone/CMakeLists.txt
A       tools/CMakeLists.txt
A       tools/CodeBase/CMakeLists.txt
A       doc/README.cmake_build
M       tools/fseek_test.c
M       README

- Modified file include an adjustment to a compile test to allow the test to be conducted in an out-of-source build manner as prescribed by CMake. Default logic of this test to still test on the existence of `Makefile`

TESTS CONDUCTED: 
1. In various instances this build is faster and more reliable with meaningful diagnostics when errors occur

RELEASE NOTE: 
Introduction of a CMake build system for em_real and em_ideal


islas:
Additional notes:
While most files are additions, there are a few key edits/removals comprising of
* Analogous `#ifdef` syntax to better support automated dependency graph generation
* Removal of redundant code (such as wrf_io_flags.h)
* Missing characters that were being addressed through standard.exe/sed/some other form of preprocessing to avoid such need
* Comments to note why certain things were done a particular way (mainly XDEX)

phys/module_mp_fast_sbm.F
```diff
@@ -6,6 +6,20 @@ SUBROUTINE SBM_fast
...
+module module_mp_SBM_Collision
+end module module_mp_SBM_Collision
+
+module module_mp_SBM_Auxiliary
+end module module_mp_SBM_Auxiliary
+
+module module_mp_SBM_Nucleation
+end module module_mp_SBM_Nucleation
+
 #else
```
dudhia, line 23: This is related to 
#if( BUILD_SBM_FAST != 1)
Can you remind me when this stub is used if it is in any of our builds?

islas: From what I could tell, it is used only when double-precision builds are requested as the subsequent code has a lot of mismatched complex types
The setting is [done in configure](https://github.com/wrf-model/WRF/blob/3dd1546ba425b5e77d682dd92253738416b3cb0a/configure#L602C8-L602C92)

dudhia:
Thanks for checking.

dudhia:
Note: #1896 and #1899 have similar changes. Not sure which to review.

islas:
#1899 Is the broken out portions that don't necessarily depend on the new files. It takes precedence as an incremental set of changes that push toward making this easier. 

note: I left the changes in here to avoid too much git history manipulation though I imagine I will pay for that later if #1899 goes through

weiwangncar:
The Jenkins test results:

    Failed Simulations are: 
    None
    None

phys/module_sf_clm.F
```diff
@@ -60994,7 +60994,7 @@ subroutine clminit(VEGFRA,SNOW,SNOWC,SNOWH,CANWAT,SMSTAV,        &
      ENDDO
    ENDDO
    IF ( errflag .EQ. 1 ) THEN
-      CALL wrf_error_fatal( "CLM: clminit: out of range value "// &
+      CALL wrf_error_fatal( "CLM: clminit: out of range value ", &
```
mgduda, line 60997: What is the reason for this change?

dyn_em/module_big_step_utilities_em.F
```diff
@@ -5733,11 +5733,11 @@ SUBROUTINE moist_physics_finish_em( t_new, t_old, t0, mut,     &
 !                                300                                     ! term 4
 !                    
          IF ( ( config_flags%use_theta_m .EQ. 1 ) .AND. (P_Qv .GE. PARAM_FIRST_SCALAR) ) THEN
-            t_new(i,k,j) =  h_diabatic(i,k,j)*(1. + (R_v/R_d)*qv_diabatic(i,k,j)) + \
-                            mpten*(1. + (R_v/R_d)*qv(i,k,j)) + \
+            t_new(i,k,j) =  h_diabatic(i,k,j)*(1. + (R_v/R_d)*qv_diabatic(i,k,j)) + &
+                            mpten*(1. + (R_v/R_d)*qv(i,k,j)) + &
```
mgduda, line 5737: Since these changes aren't strictly related to the new build system, if they haven't been already, should they be introduced through a separate PR?

mgduda (changes requested):
Is the addition of the `.ci` directory necessary for the new build system? If not, it may be better to add that through a separate PR.

islas:
> Is the addition of the `.ci` directory necessary for the new build system? If not, it may be better to add that through a separate PR.

I debated adding it as it is not necessary per se, however it contains the first drafts of tests to actually test this new build system. I'd argue that adding tests directly with the new feature is a great addendum to a PR

phys/module_sf_clm.F
mgduda: I think we've typically used flags like `-P -traditional` with `cpp` to avoid various issues when processing Fortran. Given that `//` is a standard concatenation operator in Fortran, rather than forevermore having to ensure that new code doesn't make use of this, it seems better to adjust our preprocessing if at all possible.

mgduda:
> > Is the addition of the `.ci` directory necessary for the new build system? If not, it may be better to add that through a separate PR.
> 
> I debated adding it as it is not necessary per se, however it contains the first drafts of tests to actually test this new build system. I'd argue that adding tests directly with the new feature is a great addendum to a PR. That said as the new proposed testing framework is not yet officially part of the baseline, I can see how this might not be an ideal addition.

I 100% agree that adding tests along with new code is a great idea in principle. In this case specifically, since the testing framework isn't in place -- and since that framework could change between the time this PR is merged and the time that it becomes operational -- my inclination would be to omit these tests for the present. Until we can actually run the tests, we don't know that they work; for example on line 26 of `.ci/tests/buildCMake.sh` there's a reference to `.ci/env/helpers.sh`, which doesn't appear to exist in this PR.

phys/module_sf_clm.F
islas: I see, I think that flag slipped past me, as I had `-P -nostdinc` but not `-traditional` added to my `CMAKE_C_PREPROCESSOR_FLAGS` @ line 25 at the top-level CMakeLists.txt

islas:
@mgduda Removal was done via `git rm` as a rebase was proving to be rather gnarly with the merge commit in the middle. 

Only edit should be fseek_test.c to allow testing files other than `Makefile`, which does fall into the realm of specific to this build. I adjusted the description of the PR to include this info

islas:
Push `6684d468` fixes the commit message typo, apologies

mgduda (approved):

rcabell:
One potential source of confusion for users (it confused me at first!) is that in the `configure_new` script there are several yes/no questions that are posed with `[Y/n]` at the end of the line. The capital 'Y' typically indicates that this is the default answer, however, the script defaults answers to false because the optional is set using  `input().lower() in yesValues`. So pressing return after these questions, assuming the default unanswered value will be 'yes', leads to unexpected results.

arch/configure_reader.py
```diff
@@ -0,0 +1,617 @@
...
+  yesValues    = [ "yes", "y", "true", "1" ]
+
+  ##############################################################################
+  # Decompose the weird way to write the logic for DM/SM 
+  USE_MPI = False
+  if ( stanzaCfg.serialOpt_ or stanzaCfg.smparOpt_ ) and ( stanzaCfg.dmparOpt_ or stanzaCfg.dmsmOpt_ ) :
+    # togglable
+    # we can safely check this since the user would not have been able to select this stanza if it couldn't be disabled
+    if stanzaCfg.dmCompilersAvailable() :
+      useMPI       = input( "[DM] Use MPI?    [Y/n] : " ).lower() in yesValues
```
rcabell, line 582: One potential source of confusion for users (it confused me at first!) is that in the script there are several yes/no questions that are posed with [Y/n] at the end of the line. The capital 'Y' typically indicates that this is the default answer, however, the script defaults answers to false because the optional is set using `input().lower() in yesValues`. So pressing return after these questions, assuming the default unanswered value will be 'yes', leads to unexpected results.

rcabell (changes requested):

rcabell (changes requested):
Eventually (not this version, probably), it would be nice to have the configure script be wrapped non-interactively into the CMakeLists.txt build script. One of the strengths of CMake is that it can query and enumerate the build environment (which compiler is being used, paths to libraries including MPI, etc.), so I don't see why this couldn't be reduced to a simple `cmake -B _build -DWRF_CORE=0 -DWRF_NESTING=1` in the future.

arch/configure_reader.py
islas: I genuinely didn't know this! I thought the alternation of capital/lower `Y/n` was meant to indicate to the user that any capital or lower input would be acceptable. This is definitely confusing then.

rcabell: Another option would be to not allow empty input, that would make things more verbose but eliminate any ambiguity.

islas:
> Eventually (not this version, probably), it would be nice to have the configure script be wrapped non-interactively into the CMakeLists.txt build script. One of the strengths of CMake is that it can query and enumerate the build environment (which compiler is being used, paths to libraries including MPI, etc.), so I don't see why this couldn't be reduced to a simple `cmake -B _build -DWRF_CORE=0 -DWRF_NESTING=1 -DWRF_CASE=em_real` in the future.
> 
> Note: I noticed this comment is listed as 'requested changes' but that is for the above review. This is just a comment about potential future development...

I did not like that the script required interactive input especially when devising automated tests so `./configure_new` has some extra features that allow no interaction to happen at all - these advanced features are what I'm going to add to a readme here shortly.

The gist is that you can do:
```
./configure_new [options] [-- <CMake configuration options>]
  -p                  Preselect a stanza configuration with matching description
  -x                  Skip CMake options prompt, meant to be used in conjunction with direct pass-in options
  -d directory        Use as alternate build directory
  -i directory        Use as alternate install directory
  -- <CMake options>  Directly pass CMake options to configuration, equivalent to cmake <source dir> <CMake Options>
  -h                  Print this message

# And using the above
./configure_new -p GNU -d _build -i $PWD/install_here -x -- -DWRF_CORE=0 -DWRF_NESTING=1 -DWRF_CASE=em_real ...and any other options...
```

I avoid using `-S` or `-B` options from cmake as those require higher versions than necessary, and preselection of stanza did in fact make sorting which ancillary flags to use much easier rather than embedding the logic in CMake itself.

arch/configure_reader.py
islas: Would swapping to `y/N` be acceptable, or do you think some of these options should be on by default while some off?

rcabell: I think at the very least MPI should default to true given that most non-trivial cases require it. Regarding the others, I'm not sure. Others closer to the code may have a more valid opinion...

islas: @mgduda Do you have a particular preference on which way we go with default options like this?

islas:
New pushes address documentation, default values properly reflected (for now all defaults are `N` and `0` where appropriate), and minor modification to cleaning script to match output to command execution

arch/configure_reader.py
mgduda: I would agree with @rcabell that MPI should probably default to `Y`. I would also suggest that `BASIC` nesting (nesting option `1`) should be the default as well.

CMakeLists.txt
```diff
@@ -0,0 +1,919 @@
...
+##
+################################################################################
+# Mode configuration
+
+set( OPTIMIZATION_LEVEL "" )
+set( WRF_OS             "" )
+set( WRF_MACH           "" )
+
+if ( "${CMAKE_BUILD_TYPE}" STREQUAL "" )
+  set( CMAKE_BUILD_TYPE Debug )
```
mgduda, line 68: If my understanding is correct, and the default build type is `Debug`, I would propose we make the default `Release` (or something with optimizations enabled).

mgduda:
As a general comment / question, since we have `configure_new` and `compile_new`, would it help users if we renamed the `cleanCMake.sh` script to `clean_new`?

islas:
> As a general comment / question, since we have `configure_new` and `compile_new`, would it help users if we renamed the `cleanCMake.sh` script to `clean_new`?

I'm fine with either way on this.

arch/configure_reader.py
```diff
@@ -0,0 +1,628 @@
...
+    # First get os & archs
+    osarchMatch = osAndArch.match( self.osArchLine_ )
+
+    if osarchMatch is None :
+      osarchMatch = osAndArchAlt.match( self.osArchLine_ )
+      if osarchMatch is None :
+        print( "Could not find OS and architecture info in " + self.osArchLine_ )
+      
+    self.os_    = osarchMatch.group(1)
+    self.archs_ = osarchMatch.group(2).strip(",").split( " " )
```
rscabell, line 49: This RE breaks on macOS, because it reads the (MACOS) comment from the `configure.defaults` file. You could either make a special-case check here, or, update `configure.defaults` with something like this:

```diff
diff --git a/arch/configure.defaults b/arch/configure.defaults
index 36ca1b6e..34038a87 100644
--- a/arch/configure.defaults
+++ b/arch/configure.defaults
@@ -1000,7 +1000,7 @@ CC_TOOLS        =      $(SCC)
 NETCDFPAR_BUILD        =      CONFIGURE_NETCDFPAR_BUILD
 
 ###########################################################
-#ARCH    Darwin (MACOS) gfortran with gcc #serial smpar dmpar dm+sm
+#ARCH    Darwin x86_64, aarch64 (MACOS) gfortran with gcc #serial smpar dmpar dm+sm
 #
 DESCRIPTION     =       GNU ($SFC/$SCC)
 DMPARALLEL      =       # 1
@@ -1045,7 +1045,7 @@ CC_TOOLS        =      $(SCC)
 NETCDFPAR_BUILD        =      CONFIGURE_NETCDFPAR_BUILD
 
 ###########################################################
-#ARCH    Darwin (MACOS) gfortran with clang #serial smpar dmpar dm+sm
+#ARCH    Darwin x86_64, aarch64 (MACOS) gfortran with clang #serial smpar dmpar dm+sm
 #
 DESCRIPTION     =       GNU ($SFC/clang)
 DMPARALLEL      =       # 1
```

islas: I don't think I follow. The `osAndArch` regex properly catches the lines you mentioned in my tests when enabling (via changing the skip logic) cross-platform stanza selection. The last bit of the regex has a non-capture group for the `(...)` to deal with `(MACOS)`

It would be nice though to not have to deal with special cases for all the inconsistent formatting in comments/descriptions of things in `configure.defaults`

rcabell: On my MacBook Pro, the script failed to find any stanzas until I manually added the architectures to the `configure.defaults` file. Printing out the `Stanza.archs_` variable is how I found that it was capturing the `(MACOS)` token. I did not try using the cross-platform flag.  

I also found that using the gfortran/clang stanza failed as well because the `-fallow-argument-mismatch` flag was being passed to both gfortran and clang (where it's an invalid argument). Using gcc/gfortran worked fine after the edits to the configure file. I'll dig into that clang issue and post another comment when I determine the underlying issue.

islas: I see. I think I was focused on simply getting the regex to capture something, but that something is eventually compared to `platform.machine()` in python. I think a change would be best.

rcabell: I tried again but got the same result (no stanzas) when building on a M1 Mac -- as it turns out, macOS uses `arm64` as an architecture identifier vs `aarch64` that Linux uses. My patch worked on a Intel Mac but not Arm-based. My apologies; I should have double-checked that.

islas: Just double checking, adding `arm64` to the list of archs in the line does properly show the modified stanza?

rcabell: > Just double checking, adding `arm64` to the list of archs in the line does properly show the modified stanza?

Yes, but I would change `aarch64` to `arm64` for the Darwin stanzas vs. adding it. The macOS reports it only as `arm64` or `x86_64` from the `uname()` call (which is what Python is using under the hood).

rcabell:
One additional issue I found was that when building with `gfortran/clang` the Clang version of `cpp` failed when preprocessing Fortran source files. Making sure that the GCC version was first in the $PATH fixed the issue. This is probably not anything that the CMake system needs to worry about, but it might be worth mentioning in the documentation somewhere.

islas:
> One additional issue I found was that when building with `gfortran/clang` the Clang version of `cpp` failed when preprocessing Fortran source files. Making sure that the GCC version was first in the $PATH fixed the issue. This is probably not anything that the CMake system needs to worry about, but it might be worth mentioning in the documentation somewhere.

Interesting. Does the make build system not fail with that same setup? CMake doesn't have a built in variable documented for the C preprocessor, so the build currently takes verbatim what is listed in the `configure.defaults` file for that stanza.

rcabell:
> Does the make build system not fail with that same setup?

It probably does, as I made no change to the CMake setup other than to change the order of things in my `$PATH`. It's been a long time since I've built WRF on a Mac! Both are called `cpp` so there's not much the build script can do other than call it with `--version` and warn the user if it greps to clang. Probably more trouble than it's worth...

arch/configure_reader.py
islas: Okay, should be fixed now

rcabell:
I found another potential issue: the Cray (CCE) stanzas don't get detected on Derecho because `platform.system()` returns `Linux` instead of `Cray`. Given that the default compiler is Intel, it's not a show-stopper, but maybe something to look into i in the next batch of CMake updates.

rcabell (dismissed):
Issues identified have been fixed; approved.

islas:
@mgduda When you get a chance, would you mind reviewing the updates?

mgduda:
With both the Intel and GNU compilers on Derecho, I'm not able to get successful builds. With the Intel compilers (default modules on Derecho), compilation fails with:
```
[ 51%] Building Fortran object CMakeFiles/WRF_Core.dir/frame/module_comm_nesting_dm.F.o
ifort: error #10106: Fatal error in /glade/u/apps/common/23.08/spack/opt/spack/intel-oneapi-compilers/2023.2.1/compiler/2023.2.1/linux/bin/intel64/../../bin/intel64/fortcom, terminated by kill signal
compilation aborted for /glade/derecho/scratch/duda/pull/1896/frame/module_comm_nesting_dm.F (code 1)
make[2]: *** [CMakeFiles/WRF_Core.dir/build.make:4268: CMakeFiles/WRF_Core.dir/frame/module_comm_nesting_dm.F.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:508: CMakeFiles/WRF_Core.dir/all] Error 2
make: *** [Makefile:136: all] Error 2
```
With the GNU compilers (just running `module load gcc` starting from the default modules), compilation fails with:
```
[ 65%] Building Fortran object CMakeFiles/WRF_Core.dir/phys/module_cu_ntiedtke.F.o
/glade/derecho/scratch/duda/pull/1896/phys/module_cu_ntiedtke.F:5:6:

    5 |  use cu_ntiedtke,only: cu_ntiedtke_run, &
      |      1
Fatal Error: Cannot open module file 'cu_ntiedtke.mod' for reading at (1): No such file or directory
compilation terminated.
make[2]: *** [CMakeFiles/WRF_Core.dir/build.make:960: CMakeFiles/WRF_Core.dir/phys/module_cu_ntiedtke.F.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:508: CMakeFiles/WRF_Core.dir/all] Error 2
make: *** [Makefile:136: all] Error 2
```
One important thing to note is that I've merged (locally) the `cmakeBuild` branch onto the head of the `develop` branch; so perhaps there are just a few updated needed to the PR branch so that it will work once merged?

mgduda:
> > As a general comment / question, since we have `configure_new` and `compile_new`, would it help users if we renamed the `cleanCMake.sh` script to `clean_new`?
> 
> I'm fine with either way on this.

If the purpose of the `cleanCMake.sh` script is roughly the same as the `clean` script, then I'd advocate for renaming `cleanCMake.sh` to `clean_new`. If not, though, perhaps the purpose of the `cleanCMake.sh` script (including when it is to be used) could be documented?

rcabell:
> One important thing to note is that I've merged (locally) the `cmakeBuild` branch onto the head of the `develop` branch; so perhaps there are just a few updated needed to the PR branch so that it will work once merged?

I was able to build on Derecho with the PR as it is, so I suspect you're right about needing to rebase it to the current develop.

weiwangncar:
@mgduda I encountered some error when compiling this branch. I was then able to complete the compile with the use of qcmd command.

islas:
> [T]he purpose of the `cleanCMake.sh` script (including when it is to be used) could be documented

Its use is more general than current `clean`, so in that regard it is slightly different as it does not contain any WRF-specifics. The usage is documented in the `doc/README.cmake_build`

mgduda:
> > [T]he purpose of the `cleanCMake.sh` script (including when it is to be used) could be documented
> 
> Its use is more general than current `clean`, so in that regard it is slightly different as it does not contain any WRF-specifics. The usage is documented in the `doc/README.cmake_build`

I missed the documentation in `doc/README.cmake_build`. I'm still wondering if there's a different name for this script that might make its role clearer?

islas:
@mgduda Best I can tell, `ifort: error #10106` is a memory issue - this might be due to the fact that super large files like `module_comm_nesting_dm.F` are no longer preprocessed separately and instead passed to the Fortran compiler with native preprocessing flags.

mgduda:
Building with the HEAD of this PR branch (983f82a9) does succeed with the GNU compilers on Derecho for me, so it would appear that the nTiedtke error is related to recent changes in the `develop` branch.

mgduda:
I'd like to get some clarification on how the expected workflow under this build system will look for users. Right now, the `real` and `wrf` binaries are installed in `install/bin` with copies in `install/test`. However, in the `test/install` directory, although there are binaries and physics tables, example namelists are in a `em_real` subdirectory. Do we expect users to copy namelists from the `install/test/em_real` directory into the `install/test` directory?

Perhaps it's a bit late in the PR process to raise this issue, but we have a large amount of documentation that will need to be updated with the current organization of install and run directories. What if we were to try to mirror the workflow implied by the current build system, but just under the `install` subdirectory (recognizing that we could support out-of-source builds). So for example, the `install/test/em_real` directory would contain binaries, physics tables, and namelists -- basically everything in the current `test/em_real` directory.

Also, is there a strong argument for dropping the `.exe` suffix from binaries? I don't personally prefer this suffix (as evidenced by the names of executable files in a certain other model), but I do think there's value in trying to minimize unnecessary changes at this point in WRF's lifecycle, especially given the untold number of users who may have scripts for running complex WRF workflows, and given the large body of tutorial material and other documentation.

weiwangncar:
> Building with the HEAD of this PR branch ([983f82a](https://github.com/wrf-model/WRF/commit/983f82a9ee602abab66aa93cf620673fcb354f8b)) does succeed with the GNU compilers on Derecho for me, so it would appear that the nTiedtke error is related to recent changes in the `develop` branch.

Could this be related to the newly added physics_mmm/ directory and several physics modules now residing in that directory?

weiwangncar:
@mgduda @islas From a user's perspective, I'd agree with Michael's suggestion. It would be great if the final executables are named the same as before, perhaps even linked to the current run/ and test/em_real/ (or any of the test/em_<ideal> case directory.

islas:
> I'd like to get some clarification on how the expected workflow under this build system will look for users. Right now, the `real` and `wrf` binaries are installed in `install/bin` with copies in `install/test`. However, in the `install/test` directory, although there are binaries and physics tables, example namelists are in a `em_real` subdirectory. Do we expect users to copy namelists from the `install/test/em_real` directory into the `install/test` directory?

This is a bug, everything was meant to be installed under test/em_real. In shuffling things around it seems that the binaries and tables from main/ are getting installed one directory higher than they ought to. I'll fix this as soon as I merge develop into this branch - currently running into weird unforeseen issues.
 

 
> Also, is there a strong argument for dropping the `.exe` suffix from binaries? I don't personally prefer this suffix (as evidenced by the names of executable files in a certain other model), but I do think there's value in trying to minimize unnecessary changes at this point in WRF's lifecycle, especially given the untold number of users who may have scripts for running complex WRF workflows, and given the large body of tutorial material and other documentation.

It's standard practice for *nix binaries to not have an extension, more so the `.exe` suffix is most generally applied to Windows binaries. If the workflow can more or less be replicated in the out-of-source build I think this is a much smaller change that could be tolerated.


> perhaps even linked to the current run/ and test/em_real/ (or any of the test/em_ case directory

With fixing the install location of tables & binaries we shouldn't need to link back into run/ or test/<selected case>. The workflow should ideally be going into the out-of-source build (install/) and then from there everything should be the same - though right now a run/ directory does not exist as I was under the impression that almost all runs are started from test/<selected case>.

dudhia:
My vote would be to
1. have executables linked to the existing run and test directories where
the data and namelist files reside.
2. at least have the .exe suffix in the linked files in those directories
to reduce user confusion.
I prefer suffixes if only to distinguish files from directories.
Jimy

dudhia:
For me the default intel compiler was successful with qcmd, but I still
have problems with configuring when I switch to gcc.
It hangs after this. Not sure what the loadedmodules message means.
--------------------
Default [0] :
[DM] Use MPI?    Default [Y] [Y/n] :
[SM] Use OpenMP? Default [N] [y/N] :
Configure additional options? Default [N] [y/N] :
-- Cray Programming Environment 2.7.23 C
-- NOTE: LOADEDMODULES changed since initial config!
-- NOTE: this may cause unexpected build errors.
-- Cray Programming Environment 2.7.23 CXX
-- Cray Programming Environment 2.7.23 Fortran

islas:
> It hangs after this. Not sure what the loadedmodules message means.
> Default [0] :
[DM] Use MPI?    Default [Y] [Y/n] :
[SM] Use OpenMP? Default [N] [y/N] :
Configure additional options? Default [N] [y/N] :
-- Cray Programming Environment 2.7.23 C
-- NOTE: LOADEDMODULES changed since initial config!
-- NOTE: this may cause unexpected build errors.

While I'm not exactly certain what is under the hood causing this the issue arises from swap/switching modules _and_ not clearing the previous build entirely. `./cleanCMake.sh` on its own only does make clean, so if this is done then a module is changed, the previous configuration is being used for compilation checks causing cmake to hang indefinitely. 

Solution: Do `./cleanCMake.sh -a` when switching modules as that invalidates the previous configuration.

dudhia:
> Solution: Do `./cleanCMake.sh -a` when switching modules as that invalidates the previous configuration.

Update: This worked for configure and compile after switching to gcc.

dudhia:
Even though gcc compiled there were some errors in copies of 4 files as the target directory looks wrong.
---
Error copying file (if different) from "/glade/u/home/dudhia/cmake-test/WRF/run/p3_lookupTable_1.dat-5.3-2momI" to "/glade/u/home/dudhia/cmake-test/WRF/install/test//p3_lookupTable_1.dat-5.3-2momI".

weiwangncar:
@islas I would still recommend removing COAMPS in the list of models to compile. Yes, it is in the list of things to compile in the old make system, but nobody sees it, and it never was a real option (but no one bothered to remove it, true). But in cmake, it shows up as an option for users. It can be confusing.

mgduda (changes requested):
I have just two more requests, and then I think any remaining cleanup issues can be ironed out before the release:
1. Right now it looks like namelists are placed in `install/test/em_real/em_real`, and I think these should go in `install/test/em_real`.
2. Would it make sense to symlink the binaries from `install/bin` to the `install/test/em_real` directory rather than copying them? I would also propose that we have symlinks to the binaries in the `install/test/em_real` directory with the `.exe` suffix. I know this isn't the Unix way, but I really think there's value in not forcing unnecessary updates to documentation, workflows, etc. As a compromise, I wouldn't object to having both names in the `install/test/em_real` directory (with and without the `.exe` suffix).

mgduda (approved):

weiwangncar:
@rcabell Can you help review and approve this PR?

weiwangncar (approved):

islas:
There isn't a short-hand for selecting debug (e.g. the old `-d` option for debug) but you can use the `CMAKE_BUILD_TYPE` to select debug compile options.

This, and any other cmake options, can be done with the `--` delimiter into `./configure_new` :
`./configure_new -- -DCMAKE_BUILD_TYPE=Debug`

PR #1904, opened 2023-07-27
Fix Triggering Condition Bug for CAPE Calculation

### LINKED ISSUE #1903: CAPE/CIN bug in afwa diagnostics
@weiwangncar - This PR introduced a bug.

`plfc` is initialized to a dymmy value here:

https://github.com/wrf-model/WRF/blob/3dd1546ba425b5e77d682dd92253738416b3cb0a/phys/module_diag_functions.F#L518-L526

Once the `lfclev` is found then `plfc` is then assigned it final value:
https://github.com/wrf-model/WRF/blob/3dd1546ba425b5e77d682dd92253738416b3cb0a/phys/module_diag_functions.F#L808-L813

Notice that `CAPE` is only updated when `p(k) <= plfc`:

https://github.com/wrf-model/WRF/blob/3dd1546ba425b5e77d682dd92253738416b3cb0a/phys/module_diag_functions.F#L774-L777

This makes sense based on the PR description but the problem is that the `CAPE` assignment happens before `plfc` is assigned. Therefore the condition on line 774 uses the dummy value for `plfc` and it always ends up being false. The overall effect of this is that `cape` is never updated and the WRF outputs for `AFWA_CAPE`, `AFWA_CAPE_MU`, etc. are always zero now.

All we need to do is move the `plfc = p(lfclev)` assignment to somewhere before it is used for the `cin`/`cape` assignment loop at line 821.

_Originally posted by @sael9740 in https://github.com/wrf-model/WRF/issues/1775#issuecomment-1652512844_

zhixiaozhang:
Assign PLFC before calculating CAPE since PLFC is the triggering condition for the CAPE calculation

TYPE: bug fix

KEYWORDS: CAPE, PLFC

SOURCE: Zhixiao Zhang (University of Oxford)

DESCRIPTION OF CHANGES:
Problem:
CAPE can't be updated

Solution:
Assign PLFC value before calculating CAPE because PLFC is the triggering condition for CAPE calculation.

ISSUE: Fixes #1903

LIST OF MODIFIED FILES: WRF/phys/module_diag_functions.F

RELEASE NOTE: Fix triggering condition code bug for CAPE calculation

### DIFF
--- a/phys/module_diag_functions.F
+++ b/phys/module_diag_functions.F
@@ -760,9 +760,10 @@ CONTAINS
             flag = .false.
          END IF
       END DO
-      IF (ellev >= 0) THEN !Modified by Zhixiao
-         pel = p(ellev) !Modified by Zhixiao
-         CIN=REAL ( 0 ) !Modified by Zhixiao
+      IF ((ellev >= 0) .and. (lfclev >= 0)) THEN !Modified by Zhixiao
+         plfc = p (lfclev)
+         pel = p (ellev)
+         CIN=REAL ( 0 )
          DO k = sfc+1, nz
             ! Make CAPE and CIN consistent with AMS definition
             ! https://glossary.ametsoc.org/wiki/Convective_available_potential_energy


dudhia:
We will create a branch for V4.5.2 soon. This should probably go in there. It just missed 4.5.1.

weiwangncar:
@sael9740 Can you see if this PR fixes the problem you saw? Thanks!

weiwangncar:
The regression test results:

    Failed Simulations are: 
    None
    None

sael9740:
@weiwangncar - Yes it does. It is a little funky that it is setting `plfc` twice though. Once at line 764 and once at line 812. I would instead recommend moving lines 809-819 
https://github.com/wrf-model/WRF/blob/3dd1546ba425b5e77d682dd92253738416b3cb0a/phys/module_diag_functions.F#L808-L818

to just before the `IF ((ellev >= 0) .and. (lfclev >= 0)) THEN` on line 763. That would look a lot cleaner to me. But this is just me being picky about readability/cleanliness since the proposed changes work fine.

weiwangncar:
@zhixiaozhang What do you think of the comment by @sael9740?

dudhia (approved):
I will approve it to get in 4.5.2. We can look at suggested mod later.

weiwangncar (approved):

PR #1923, opened 2023-10-12
Improve NetCDF detection and linking

rjdave:
TYPE: enhancement

KEYWORDS: NetCDF4, Spack-Stack, gfortran

SOURCE: David Robertson, Rutgers University

DESCRIPTION OF CHANGES:
Problem:

- NetCDF4 detection and linking is unreliable and impossible if NetCDF-C and NetCDF-Fortran reside in different directories
- gfortran is intolerant of non-initialized variables in **`mediation_integrate.F`**.

Solution:
- Refactor the detection and configuration of NetCDF dependencies and add the capability to use NetCDF-C and NetCDF-Fortran libraries that don't reside in the same directory by introducing a new variable for the NetCDF-C path.
 -  The **fname** and **n2**  `intent(out)` variables are not assigned under certain circumstances so we should initialize them to empty strings.
``` 
   ierr = 0
   fname = ""
   n2 = ""

   ! Note that computation of fname and n2 are outside of the oid IF statement
```     

LIST OF MODIFIED FILES: list of changed files (used `git diff --name-status release-v4.5.2` to get formatted list)
```
M       Makefile
M       arch/Config.pl
M       arch/postamble
M       arch/preamble
M       configure
M       share/mediation_integrate.F
```
TESTS CONDUCTED: 
1. Compiled and ran with several different configurations of NetCDF, including both shared and static versions of both split (NetCDF-C and NetCDF-Fortran in separate directories) and unified library locations. The shared library versions were from Spack-Stack; the static libraries were custom-built for my systems.
2. Ran the coupled WRF-ROMS standard test case for Hurricane Irene.
3. The Jenkins tests are all passing.

RELEASE NOTE: This PR improve the detection and linking with shared and static configurations of the NetCDF-4 libraries (split or unified) inside and outside Spack-Stack.

PR #1924, opened 2023-10-12
Remove arbitrary limiting of heat fluxes

rjdave:
Arbitrary limitation in the computation of **QFX** and **HFX**.

TYPE: bug fix

KEYWORDS: SBL

SOURCE: David Robertson, Rutgers University

DESCRIPTION OF CHANGES:
Problem:

- Arbitrary limitation in the computation of **QFX** and **HFX** in modules  **`module_sf_sfclay.F`** and **`module_sf_sfclayrev.F`** that affects the computation latent heat fluxes and sensible heat flux in coastal areas that limit fog layers.

Solution:
- Comment out the lines in **`phys/module_sf_sfclay.F`** :
 ```f90
       DO 370 I=its,ite
         QFX(I)=FLQC(I)*(QSFC(I)-QX(I))
!       QFX(I)=AMAX1(QFX(I),0.)
         LH(I)=XLV*QFX(I)
  370 CONTINUE

...

        ELSEIF(XLAND(I)-1.5.LT.0.)THEN
          HFX(I)=FLHC(I)*(THGB(I)-THX(I))
!        HFX(I)=AMAX1(HFX(I),-250.)
        ENDIF
  400 CONTINUE
```
 - Similar changes in **`phys/module_sf_sfclayrev.F`**.

LIST OF MODIFIED FILES: 
M       phys/module_sf_sfclay.F
M       phys/module_sf_sfclayrev.F

TESTS CONDUCTED: 
1. Ran the coupled WRF-ROMS standard test case for Hurricane Irene.
2. It passed regression tests.

RELEASE NOTE: This PR removed zero negative latent heat flux limit (atmosphere to water) in the revised MM5 and original MM5 surface layer schemes.

### DIFF
--- a/share/mediation_integrate.F
+++ b/share/mediation_integrate.F
@@ -1701,6 +1701,8 @@ SUBROUTINE open_hist_w ( grid , config_flags, stream, alarm_id, &
    ENDIF
 
    ierr = 0
+   fname = ""
+   n2 = ""
 
    ! Note that computation of fname and n2 are outside of the oid IF statement 
    ! since they are OUT args and may be used by callers even if oid/=0.  
--- a/share/mediation_integrate.F
+++ b/share/mediation_integrate.F
@@ -1701,8 +1701,6 @@ SUBROUTINE open_hist_w ( grid , config_flags, stream, alarm_id, &
    ENDIF
 
    ierr = 0
-   fname = ""
-   n2 = ""
 
    ! Note that computation of fname and n2 are outside of the oid IF statement 
    ! since they are OUT args and may be used by callers even if oid/=0.  
--- a/phys/module_sf_sfclay.F
+++ b/phys/module_sf_sfclay.F
@@ -892,7 +892,7 @@ CONTAINS
 !                                                                               
       DO 370 I=its,ite
         QFX(I)=FLQC(I)*(QSFC(I)-QX(I))                                     
-        QFX(I)=AMAX1(QFX(I),0.)                                            
+!       QFX(I)=AMAX1(QFX(I),0.)                                            
         LH(I)=XLV*QFX(I)
   370 CONTINUE                                                                 
                                                                                 
@@ -910,7 +910,7 @@ CONTAINS
 !         ENDIF
         ELSEIF(XLAND(I)-1.5.LT.0.)THEN                                       
           HFX(I)=FLHC(I)*(THGB(I)-THX(I))                                
-          HFX(I)=AMAX1(HFX(I),-250.)                                       
+!         HFX(I)=AMAX1(HFX(I),-250.)                                       
         ENDIF                                                                  
   400 CONTINUE                                                                 
 
--- a/phys/module_sf_sfclayrev.F
+++ b/phys/module_sf_sfclayrev.F
@@ -1041,7 +1041,7 @@ CONTAINS
 
       DO 370 I=its,ite
         QFX(I)=FLQC(I)*(QSFC(I)-QX(I))                                     
-        QFX(I)=AMAX1(QFX(I),0.)                                            
+!       QFX(I)=AMAX1(QFX(I),0.)                                            
         LH(I)=XLV*QFX(I)
   370 CONTINUE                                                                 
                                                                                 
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

**Comment by weiwangncar** [comment]
@dudhia Can you comment on why the limits were needed historically?
@rjdave Can you show some comparison plots of before and after?

**Comment by hga007** [comment]
Our group has reported this limiter hack to WRF developers several times over the years, but it is still in the current version.  We have to make this correction to every version of WRF that we have used over the years.  This issue is documented in Seroka et al (2016), https://journals.ametsoc.org/view/journals/mwre/144/9/mwr-d-15-0452.1.xml. See Figure 10:

image

You may search for **250** in that publication to see an explanation in the text:

> (v) Sensitivity to latent heat flux < 0 over water

> In the WRF surface layer scheme code, a switch exists that disallows any latent heat flux < 0 W m−2. (There is also a 
switch that disallows any sensible heat flux less than −250 W m−2.) WRF convention for negative heat flux is downward, or from atmosphere to land or water surface. This sensitivity involves removing the switch disallowing negative latent heat flux. This switch removal only results in changes in latent heat flux over water, because the subsequent WRF land surface scheme modifies fluxes and already allows for latent heat flux to be negative over land.

File: share/mediation_integrate.F [hunk 1]
```diff
@@ -1701,6 +1701,8 @@ SUBROUTINE open_hist_w ( grid , config_flags, stream, alarm_id, &
    ENDIF
 
    ierr = 0
+   fname = ""
+   n2 = ""
```
dudhia, line 1706: this change looks irrelevant to the PR

dudhia:
I think these limiters have been in the code since the MM5 days. As mentioned, this only affects water when an LSM is used for land. I see no obvious need for them over oceans. It may help cooling over cold water.

share/mediation_integrate.F
hga007: @dudhia, True, but it is also needed because it triggers an error in some strict compilers. The Fortran standard dictates that every **INTENT(OUT)** variable needs to be initialized. Putting a PR for adding two lines to the code that does not affect any of the computations is somewhat absurd. Otherwise, it will be ignored during updating. However, we have a lot of users worldwide that will make a fuzz for a simple error like that. We intend to use WRF as it is distributed in this repository without change. We are trying to avoid patching WRF every time there is a release. Many users have WRF-ROMS coupled applications using the ESMF/NUOPC or MCT coupling frameworks. 

We separated our PR contribution (https://github.com/wrf-model/WRF/pull/1923) to compile WRF with the NetCDF-4 library. Are you suggesting that to make this correction, we need to separate adding these inconsequential but need to lines to **mediation_integrate.F**?

dudhia: OK, I was just making sure it was unrelated but needed. Since it is so small it can be combined in this PR. It was described in the PR I now see.

weiwangncar:
@rjdave @hga007 Even though the two line change is small, but it is un-related to the physics change. To help us to track code change easily and cleanly, I would suggest that the lines to be removed from this PR. If you don't want to make another PR, you could add it to PR1923. Either way, there is no need to replace this PR or PR1923, simply push up a new change to remove the change in mediation_integrate.F. 
@hga007 Thank you for your providing justification for this PR.

hga007:
@weiwangncar, Sorry, it is too inconvenient for us to remove it.  I hope that you can overlook that level of rigidity.

dudhia:
If you put another PR with just this change first, it might be sufficient.

rjdave:
Removed minor variable initialization in **`share/mediation_integrate.F`** from this PR and added to #1923 as requested.

Should I remove the notes about **`share/mediation_integrate.F`** from the original PR above and add to #1923?

Hopefully splitting these PRs as requested multiple times will finally get these necessary changes into WRF and help broaden the environments WRF can be compiled under.

dudhia:
Thanks. This also allows separate review by our software engineers.

weiwangncar:
@dudhia Should we consider this in the develop branch? Even though it is just two-line change for each of the subroutines, but the outcome could be significant.

dudhia:
Yes, I prefer it in the develop branch as a result changer.

weiwangncar:
@dudhia Is this ok to go?

dudhia (approved):

dudhia:
I prefer RH as being suited to the more common case of pressure-level input. Note could be added in the README that SH should probably be used for native level input.

weiwangncar (approved):

PR #1928, opened 2023-10-20
fix a typo in the single-layer urban canopy model 

DanLi-BU:
TYPE: bug fix

KEYWORDS: single-layer urban canopy model, saturated specific humidity for ground

SOURCE: Dan Li (Boston University)

DESCRIPTION OF CHANGES:
Problem:
When computing the derivative of the saturated specific humidity with respect to surface temperature for the ground in single-layer urban canopy model, the equation had a typo which involved a coefficient. It should be 0.622 (as in for example the same calculations for roof and wall) instead of 0.22. 

Solution:
Change the coefficient from 0.22 to 0.622. 

LIST OF MODIFIED FILES: 
M    phys/module_sf_urban.F

TESTS CONDUCTED: 
1. case study.

RELEASE NOTE: Fix a typo in computing the saturated specific humidity in the single-layer urban canopy model. The effect is likely small due to the calculation is relevant only in the case of rain and on impervious surfaces when using default evaporation scheme (which is set in URBPARM.TBL).

### DIFF
--- a/phys/module_sf_urban.F
+++ b/phys/module_sf_urban.F
@@ -1279,7 +1279,7 @@ ENDIF
        ES=6.11*EXP( (2.5*10.**6./461.51)*(TGP-273.15)/(273.15*TGP) ) 
        DESDT=(2.5*10.**6./461.51)*ES/(TGP**2.)
        QS0G=0.622*ES/(PS-0.378*ES)        
-       DQS0GDTG=DESDT*0.22*PS/((PS-0.378*ES)**2.) 
+       DQS0GDTG=DESDT*0.622*PS/((PS-0.378*ES)**2.) 
 
        RG1=EPSG*( RX*VFGS          &
        +EPSB*VFGW*SIG*TBP**4./60.  &


dudhia:
Agreed. Any idea of the effect?

DanLi-BU:
I have to confess that I didn't do a thorough test on its effect. My guess is that the effect of this typo is probably small, because evaporation over impervious ground in the single-layer urban canopy model only kicks in when it rains (assuming one uses the default IMP_SCHEME = 1). When it doesn't rain, this typo does not make any difference because any evaporation related calculations will be multiplied by 0 (i.e., BETG).  The effect might be stronger if one uses IMP_SCHEME = 2.

weiwangncar:
The regression test results:

    Failed Simulations are: 
    None
    None

weiwangncar:
@cenlinhe Comments?

cenlinhe:
I agree. This is a bug and the proposed fix is correct.

weiwangncar (approved):
Cenlin has reviewd this PR.

dudhia (approved):

PR #1931, opened 2023-10-27

kkeene44:
TYPE: bug fix

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:

Solution:

LIST OF MODIFIED FILES: 
M   share/module_check_a_mundo.F

TESTS CONDUCTED: 
1. Now, if non-supported radiation options are used with ghg_input, the model stops with the following message:
                  --- ERROR: ghg_input available only for these radiation schemes: CAM, RRTM, RRTMG, RRTMG_fast 
                             OK = CAM LW with CAM SW
                  FATAL CALLED FROM FILE:  <stdin>  LINE:    2794 

### DIFF
--- a/share/module_check_a_mundo.F
+++ b/share/module_check_a_mundo.F
@@ -928,7 +928,7 @@
 #endif
 
 !-----------------------------------------------------------------------
-! Climate GHG from an input file requires coordinated pairing of
+! Climatr GHG from an input file requires coordinated pairing of
 ! LW and SW schemes, and restricts which schemes are eligible.
 ! Only radiation schemes CAM, RRTM, RRTMG, RRTMG_fast may be used.
 ! CAM LW and CAM SW must be used together.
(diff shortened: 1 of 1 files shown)


kkeene44:
@dudhia @weiwangncar 
Some of the tests failed here because if we have the model stop if anyone is using ghg_input and a physics scheme that doesn't work with it (e.g., goddard), it will fail. ghg_input is turned on, by default, in the registry. Is that what we want?

dudhia:
Good question but also dangerous to change defaults. Need to modify the
tests.

kkeene44:
@dudhia 
So maybe something like "if using anything other than CAM, RRTM, RRTMG, and RRTMG_fast, set ghg_input = 0" instead of stopping the model?

dudhia:
The problem is they won't notice and think they used it.

weiwangncar:
@kkeene44 Is it possible to add ghg_input = 0 to the namelist.input file for the failed runs in regression tests?

kkeene44:
@weiwangncar 
I've made the changes to the namelists in the regression tests. Is there a way to manually resubmit the Jenkins tests without having to modify a file?

dudhia:
Maybe you can add a blank space in a README?

weiwangncar:
@kkeene44 Unfortunately you cannot start a new test without changing the code. README files do not count.

kkeene44:
Closing and opening a different PR #1936 so that reg tests will run.

PR #1936, opened 2023-11-09

kkeene44:
TYPE: bug fix

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:

Solution:

LIST OF MODIFIED FILES:

TESTS CONDUCTED:

Now, if non-supported radiation options are used with ghg_input, the model stops with the following message:
--- ERROR: ghg_input available only for these radiation schemes: CAM, RRTM, RRTMG, RRTMG_fast
OK = CAM LW with CAM SW
FATAL CALLED FROM FILE: LINE: 2794
NOTE: 1 namelist settings are wrong. Please check and reset these options

Are the Jenkins tests all passing?

### DIFF
--- a/share/module_check_a_mundo.F
+++ b/share/module_check_a_mundo.F
@@ -963,6 +963,7 @@
             CALL wrf_message ( TRIM( wrf_err_message ) )
             wrf_err_message = '           OK = RRTM, RRTMG LW or SW, RRTMG_fast LW or SW may be mixed'
             CALL wrf_message ( TRIM( wrf_err_message ) )
+            count_fatal_error = count_fatal_error + 1
          END IF
       END IF
 


dudhia (dismissed):

dudhia:
Is this one not passing due to some regtest namelists not being consistent with the GHG switch?

dudhia:
Can someone look at failed checks?

kkeene44:
@dudhia 
I'm working on it. The tests that are failing are due to reg-test namelists that use radiation. ghg_input = 0 was added to all other necessary namelists, but not those set to 0. I'm adding that now, then will do another test with the updated reg-test.

smileMchen (dismissed):

kkeene44:
This reg-test only passed because the change was commented out. Don't approve it!

PR #1941, opened 2023-11-24
debug tslist file reinitialized when nest starts

### LINKED ISSUE #1927: tslist file reinitialized when nest begins
Note : I posted this issue on the forum : [https://forum.mmm.ucar.edu/threads/error-with-tslist-reinitialize-when-nest-begins.14228/#post-35888](https://forum.mmm.ucar.edu/threads/error-with-tslist-reinitialize-when-nest-begins.14228/#post-35888)

**Describe the bug**
I am running a 5-nested domains simulation. Each domain starts a certain time after its parent. Each domain has a ts output location in the center of the domain. At the end of the simulation, in the ts output file of each domain, the time series start approximately at the date when its nest starts instead of when itself starts except for d05 which behaves perfectly since he has no nest inside. If I look at the ts file during the simulation before d02 starts, I notice that d01 ts output are written normally which indicates that the ts file is reinitialized when the nest begins. 

**To Reproduce**
I am running a real case with nested domain starting later than its parent. The nest initialization options input_from_file and fine_input_stream are activated. 

**A possible solution**
I investigated a bit and found a possible solution (in the file attached).
[start_domain.txt](https://github.com/wrf-model/WRF/files/12925549/start_domain.txt)
The ts file is reinitialized when calling *start_domain(parent)* in the subroutine *med_nest_initial*. in this *parent* object the variable *have_calculated_tslocs* is .False.. I reset it to True depending on the timestep value.
It might not be the best solution but at least it indicates where the problem comes from.

Mathieu

mathieulandreau:
Correction of the bug described in https://github.com/wrf-model/WRF/issues/1927

TYPE: bug fix

KEYWORDS: tslist,nests,

SOURCE: Mathieu Landreau (LHEEA, Centrale Nantes, France)

DESCRIPTION OF CHANGES:
Problem:
When running a simulation with a nest starting later than its parent and with tslist option activated, the parent tslist file is reinitialized when the nest starts.
Minor limit : when nest have more vertical levels than parent (with vertical nesting), the number of tslist levels is limited by the parent number of levels

Solution:
Adding a test in share/start_em.F to check whether the tslist file of "grid" is already initialized.
Minor limit : adding the possiblity for max_ts_levels to be different for each level and setting the default value to -1 which is later converted to the maximum number of levels for each domain.

ISSUE:
Fixes #1927

LIST OF MODIFIED FILES: 
Registry/Registry.EM_COMMON
share/module_check_a_mundo.F
share/start_domain.F

TESTS CONDUCTED: 
The problem was observed on a personal test case and the modifications fixed it.

RELEASE NOTE:

### DIFF
--- a/Registry/Registry.EM_COMMON
+++ b/Registry/Registry.EM_COMMON
@@ -2358,7 +2358,7 @@ rconfig   real      ocean_s         namelist,domains    max_ocean      -1      -
 rconfig   integer   num_traj        namelist,domains    1               1000    irh   "num_traj" "#of trajectory" ""
 
 # variable for time series of vertical profile of U, V, Theta, GHT,a nd QVAPOR
-rconfig   integer   max_ts_level    namelist,domains    max_domains     -1     -       "max_ts_level"   "Highest model level for time series output"
+rconfig   integer   max_ts_level    namelist,domains    1              15       -       "max_ts_level"   "Highest model level for time series output"
 
 # track input
 rconfig   integer   track_loc_in    namelist,domains    1               0       -    "Number of track locations input"   ""    ""
--- a/Registry/Registry.EM_COMMON
+++ b/Registry/Registry.EM_COMMON
@@ -2358,7 +2358,7 @@ rconfig   real      ocean_s         namelist,domains    max_ocean      -1      -
 rconfig   integer   num_traj        namelist,domains    1               1000    irh   "num_traj" "#of trajectory" ""
 
 # variable for time series of vertical profile of U, V, Theta, GHT,a nd QVAPOR
-rconfig   integer   max_ts_level    namelist,domains    1              15       -       "max_ts_level"   "Highest model level for time series output"
+rconfig   integer   max_ts_level    namelist,domains    max_domains     -1     -       "max_ts_level"   "Highest model level for time series output"
 
 # track input
 rconfig   integer   track_loc_in    namelist,domains    1               0       -    "Number of track locations input"   ""    ""
--- a/share/module_check_a_mundo.F
+++ b/share/module_check_a_mundo.F
@@ -2647,13 +2647,18 @@
 !-----------------------------------------------------------------------
 !   Check that max_ts_level is smaller than the number of half levels
 !-----------------------------------------------------------------------
-      IF ( model_config_rec % max_ts_level .gt. model_config_rec %e_vert(1)-1 )  then
-        wrf_err_message = ' max_ts_level must be <= number of znu half layers '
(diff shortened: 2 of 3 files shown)


weiwangncar:
@mathieulandreau Prior to your change, the output of vertical profiles are default to be at 15 levels, which will probably cover the boundary layers. With your change from 15 to -1 and reset it to evert - 1, it means if a user didn't set max_ts_level, they'll get vertical profile for the entire model layers. My questions are 1. do we want to change the default behavior? 2. If we do, I actually prefer when max_ts_level is -1, no time series of vertical profiles are output - not all users probably need that output when turning on time series option.

mathieulandreau:
@weiwangncar 
I understand your point. The modification of the default value is not necessary. 
Are the other modifications correct ?
Mathieu

weiwangncar:
@mgduda Can you review this PR? I tested the modification in share/start_domain.F, and it fixes the problem with a nest starting at a later time, and it also works for a nest starting at hour 0.

weiwangncar:
@mathieulandreau Can you revert the changes made to Registry.EM_COMMOM and share/module_check_a_mundo.F to limit the fix only to share/start_domain.F? Thanks.

mathieulandreau (commented):
These lines fix the bug in my case

weiwangncar:
@mathieulandreau Thanks for reverting the un-related change.

mathieulandreau:
Hi,
Once again, I am not familiar with Github so could you tell me if I did it right ?
Mathieu

weiwangncar:
The regression test results:

    Failed Simulations are: 
    None
    None

dudhia (approved):

weiwangncar:
@mgduda Can you review this PR? Thanks.

mgduda:
Although the code diff is larger, I have an alternative set of changes that I think may leave the code in a slightly more maintainable state.

@mathieulandreau Would you be willing to try the changes here https://github.com/mgduda/WRF/commit/3295f43a64799ffa288e105d740c46be8a5eab4f to see if the correct the issue you're seeing?

mathieulandreau:
Hi, 
These modifications seem correct to me at first sight. I won't have so much time in the next weeks to check it.
Thank you for the work you did about this issue.
Mathieu

weiwangncar:
@mathieulandreau @mgduda Thanks, Michael! I did two tests, one with two domains starting at the same time, and one with the nest starting at a later time. The code works.

mgduda:
@weiwangncar for the v4.5.2 bugfix release, I've created PR #1953 as an alternative to this PR.

weiwangncar:
This PR is replaced by PR-1953, and hence no longer needed.

PR #1949, opened 2023-12-15

kkeene44:
TYPE: bug fix

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:

Solution:

LIST OF MODIFIED FILES:

TESTS CONDUCTED:

Now, if non-supported radiation options are used with ghg_input, the model stops with the following message:
-- ERROR: ghg_input available only for these radiation schemes: CAM, RRTM, RRTMG, RRTMG_fast
OK = CAM LW with CAM SW
FATAL CALLED FROM FILE: LINE: 2794
NOTE: 1 namelist settings are wrong. Please check and reset these options

Are the Jenkins tests all passing? Waiting for results.


kkeene44:
This is a duplicate PR to #1936, which fails the regression tests. Just checking to see if there was something funky with that PR and if this one is capable of passing (probably not, but worth a shot).

PR #1950, opened 2023-12-18
Rectify many of the missing dependencies

islas:
TYPE: bug fix

KEYWORDS: make, dependencies

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
Occasionally and non-deterministically the WRF make compilation will fail due to erroneous depend rules or missing rules which result in essentially either (1) race conditions relying on modules to just happen to be compiled in time for use (2) when rules are followed, subsequent dependencies are missing and compilation fails or (3) complete recompilation of certain files.

For instance, if one takes a look at a build log, one could see that `module_mp_thompson.o` is compiled twice - once as `module_mp_thompson.o` and again as `../phys/module_mp_thompson.o`. 

This is not or rarely observed in the current regression tests because [compilation is done three times](https://github.com/wrf-model/SCRIPTS/blob/35f81c34d6d17799b3b53e5a3f75c86bd9b96cac/script.csh#L135C2-L141C7) as a stop-gap to avoid this issue. 

Solution:
Add missing dependencies and fix malformed dependencies.

LIST OF MODIFIED FILES: 
M main/depend.common

TESTS CONDUCTED: 
1. Cannot be observed in current regression tests as they are designed to obscure this very issue. Can be observed with many single individual compilations

RELEASE NOTE: 
Fix missing or erroneous dependencies in make rules


weiwangncar:
@islas Do we care if the dependency files are not lined up neatly?

mgduda:
> @islas Do we care if the dependency files are not lined up neatly?

I'd agree. Although the current `depend.common` file isn't always perfectly tidy, it would be good to try to match what seems to be the most prevalent style, with one dependency per line and consistent indentation.

mgduda (changes requested):
It would be good to clean up the formatting.

islas:
@mgduda @weiwangncar file has been cleaned up to be completely consistent, but the diff will be large so it might be worth inspecting with whitespace diff highlighting turned off

weiwangncar:
@islas Did you also clean up the correct dependencies? If that's case, it would be hard to tell what have been corrected.

islas:
Yes, all the dependency listings were corrected to have consistent format, as it was going to be easier to process all formatting inconsistencies rather than selectively modifying only the ones coming from the missing dependency edits

dudhia (approved):
Certainly a large change.

mgduda (approved):
Although I can't say that I've verified every single line that's changed, I have done some spot checking of the dependencies, and I've tried parallel builds with the current build system using different numbers of make jobs (-j N), and all seems in order.

weiwangncar:
@islas When compiling WRF after this change, I see that phys/module_madwrf.F is always compiled even when I modified something like module_initialize_real.F, because it is now dependend on module_wrf_top.o. Can you explain why module_wrf_top is added to the dependency of module_madwrf.o?

islas:
@weiwangncar I'm not too sure. Re-running the scripts I used to generate the deps doesn't output `module_wrf_top.o` as a dependency, so I suspect it may have been a typo on my part from final manual edit and cleanup of the generated files.

PR #1959, opened 2023-12-27
Use Qv or specific humidity in interpolation if they are available

weiwangncar:
TYPE: enhancement

KEYWORDS: Qv, specific humidity, vertical interpolation

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
In current real program, if either water mixing ratio (Qv) or specific humidity (SH) is available, it first converts it to RH, perform vertical interpolation using RH, and then compute Qv for the model input. Assuming Qv and SH are coming from other model's native levels, which tends to be plenty and generally true, it should not be necessary to go through this Qv (SH) to RH, and RH to Qv steps.

Solution:
Use the flag_qv and flag_sh to avoid using interpolated RH (even though RH is still computed and vertically interpolated). If flag_sh is 1, SH is converted to Qv or mixing ratio. Then Qv is interpolated vertically. A namelist, use_sh_qv, is added for user to choose to do so if SH or Qv data is available. A test case shows that there is some differences in the final Qv in the lower levels (a bit less Qv), particularly near terrain slopes.

LIST OF MODIFIED FILES: 
M      Registry/Registry.EM_COMMON
M      dyn_em/module_initialize_real.F
M      run/README.namelist

TESTS CONDUCTED: 
1. It doesn't affect existing option if RH is the only incoming variable. Tested the code, and it gives bit-for-bit results before and after this change.

RELEASE NOTE: If water mixing ratio or specific humidity is available in the metgrid output, they can be used directly in vertical interpolation. Previously these fields are first converted to RH, interpolated vertically, and the water vapor mixing ratio is computed at model levels. Now one can use the new namelist, use_sh_qv = T to choose to use SH or Qv data directly.

### DIFF
--- a/Registry/Registry.EM_COMMON
+++ b/Registry/Registry.EM_COMMON
@@ -2282,7 +2282,7 @@ rconfig   integer interp_method_type      namelist,domains	1              2     
 rconfig   logical aggregate_lu            namelist,domains	1              .false. irh    "aggregate_lu"   "T/F aggregate the grass, shrubs, trees in LU"
 rconfig   logical rh2qv_wrt_liquid        namelist,domains	1              .true.  irh    "rh2qv_wrt_liquid" "T = rh=>Qv assumes RH wrt liquid water, F = allows ice"
 rconfig   integer rh2qv_method            namelist,domains	1              1       irh    "rh2qv_method" "1=old MM5 method, 2=new WMO method"
-rconfig   logical use_sh_mr               namelist,domains	1              .false. irh    "use_sh_mr" "T/F whether to use SH or mixing ratio in input"
+rconfig   logical use_sh_qv               namelist,domains	1              .false. irh    "use_sh_mr" "T/F whether to use SH or mixing ratio in input"
 rconfig   real    qv_max_p_safe           namelist,domains      1              10000   irh    "qv_max_p_safe" "Threshhold pressure, Qv > flag set to value"      "Pa"
 rconfig   real    qv_max_flag             namelist,domains      1              1.E-5   irh    "qv_max_flag" "Qv flag for max"  "kg kg{-1}"
 rconfig   real    qv_max_value            namelist,domains      1              3.E-6   irh    "qv_max_value" "Qv value for max"  "kg kg{-1}"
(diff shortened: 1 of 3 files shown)


dudhia:
When incoming pressure levels are coarse it is always better to interpolate RH.

jimbresch:
That is correct (since qv and sh decrease exponentially with height). But then we will need a definition of “coarse”.Also, the wrf system assumes moisture is computed with respect to water. For some models qv and sh might need to be recomputed wrt water. These kind of details are often poorly documented by the operational centers. Currently the rh is adjusted for the GFS and EC.On Dec 27, 2023, at 5:36 AM, dudhia ***@***.***> wrote:﻿
When incoming pressure levels are coarse it is always better to interpolate RH.

weiwangncar:
The regression test results:

    Failed Simulations are: 
    None
    None

dudhia:
We need to think about this one more. We should convert to RH when the vertical resolution is not coarse. Maybe we need a check for whether native levels are coming in.

weiwangncar:
@dudhia So far the datasets we've seen which provide SH data are from NCEP, EC model level data, which tend to have many more vertical levels. The capability of using MPAS model data to drive WRF also falls in this category. We could add a namelist here (MPAS also has one) to confirm that the user wants to use SH or mixing ratio data instead of converting them to RH and back to mixing ratio.

dudhia:
I would prefer it to be a choice to use SH rather than changing the default.

weiwangncar:
@dudhia I will add a namelist to this option and make the default to use RH. But I think many users could miss the opportunity and benefit to use SH/mixing ratio data directly.

dudhia (approved):

dudhia:
approved assuming it passes tests

weiwangncar:
@dudhia Yes, it has passed the tests.

weiwangncar:
@jimbresch Do you have more comment on this PR?

weiwangncar:
> @dudhia Yes, it has passed the tests.

Yes, it passed the tests.

smileMchen (approved):

PR #1963, opened 2023-12-29
add salinity effect in sfclay and sfclayrev

weiwangncar:
TYPE: enhancement

KEYWORDS: salinity effect, saturation vapor pressure, ocean

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
The salinity effect of ocean is not considered when computing saturation vapor pressure.

Solution:
Add the effect in the MM5 and revised MM5 surface layer schemes. The 0.98 factor is an approximation for salinity of 34 part per thousands, and applied for saturation specific humidity. Here the approximation is applied to saturation mixing ratio. The lakemask field is used to exclude this effect from inland lakes.

LIST OF MODIFIED FILES: 
M     phys/module_sf_sfclay.F
M     phys/module_sf_sfclayrev.F
M     phys/module_surface_driver.F

TESTS CONDUCTED: 
1. Tested in many tropical cyclone cases.

RELEASE NOTE: Add salinity effect in MM5 and revised MM5 surface layer schemes. The effect is lower the saturation vapor pressure over ocean by about 2%.

### DIFF
--- a/phys/module_sf_sfclay.F
+++ b/phys/module_sf_sfclay.F
@@ -20,7 +20,7 @@ CONTAINS
                      GZ1OZ0,WSPD,BR,ISFFLX,DX,                     &
                      SVP1,SVP2,SVP3,SVPT0,EP1,EP2,                 &
                      KARMAN,EOMEG,STBOLT,                          &
-                     P1000mb,                                      &
+                     P1000mb,LAKEMASK,                             &
                      ids,ide, jds,jde, kds,kde,                    &
                      ims,ime, jms,jme, kms,kme,                    &
                      its,ite, jts,jte, kts,kte,                    &
@@ -136,6 +136,7 @@ CONTAINS
                 INTENT(IN   )               ::             MAVAIL, &
                                                              PBLH, &
                                                             XLAND, &
+                                                         LAKEMASK, &
                                                               TSK
 
       REAL,     DIMENSION( ims:ime, jms:jme )                    , &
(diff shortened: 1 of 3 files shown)


dudhia:
Not doing sfclayrev?
Could 0.98 be made a parameter?
Should this be a switch to allow easy comparisons between versions?

weiwangncar:
The regression test results:

    Failed Simulations are: 
    None
    None

dudhia (dismissed):
I will approve it, but maybe we should have it as a switch and a parameter as in my comments.

weiwangncar:
@dudhia To reply to your earlier comments: 1. I would like to add this to sfclayrev through shared physics. 2. 0.98 is the only parameter I know, and cited in several document and papers. 3. I'm trying to avoid adding namelist, unless it is absolutely necessary.

smileMchen (dismissed):

dudhia:
OK, so the sfclayrev in shared physics will have this?
I was suggesting 0.98 to be a parameter because then it can have a variable name too which could be descriptive.

dudhia (approved):

weiwangncar:
@smileMchen Can you approve this PR one more time? Thanks.

weiwangncar:
> @smileMchen Can you approve this PR one more time? Thanks.

@smileMchen Can you help?

smileMchen (approved):

smileMchen:
@weiwangncar 
I agree that we should be cautious not to add too many namelist options.

PR #1973, opened 2024-01-05
Bug fix in gen_allocs to use correct conditional compilation logic on allocatables vs pointers

islas:
TYPE: bug fix

KEYWORDS: gen_allocs, registry, defines

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
Conditionally compiled code in `gen_allocs` is flipped to use wrong `#define` for when to use allocatables or pointers

Solution:
Flip logic of `#ifdef` to `#ifndef` to avoid using Fortran `nullify()` statement on allocatables

LIST OF MODIFIED FILES: 
M tools/gen_allocs.c

RELEASE NOTE: 
Bug fix in gen_allocs to use correct conditional compilation logic on allocatables vs pointers


mgduda (approved):
Agreed that this appears to be a bug, and that the changes in this PR are an appropriate fix.

weiwangncar:
The regression test results:

    Failed Simulations are: 
    None
    None

weiwangncar (approved):

PR #1980, opened 2024-01-10
Update module_diffusion_em.F

Lyonon:
A coefficient error in calculating BN2

TYPE: bug fix

KEYWORDS: 1.61, 0.61


dudhia:
This is not theta_v. We define theta_m this way. See documents.

weiwangncar:
@Lyonon See WRF Tech Note. This PR is closed.

PR #1942, opened 2023-11-28
Resolves building issues (WRF/WRF-chem) with Intel compilers (ifx/icx)

ahheo:
Resolving building issues with Intel compilers (ifx/icx) for WRF/WRF-chem

TYPE: bug fix

KEYWORDS: Missing/Wrong Prototypes

SOURCE: Changgui Lin

DESCRIPTION OF CHANGES:
Problem:
Problem: Most C code written long ago. No prototypes were used. See #1823. And, this pr is kind of an extension to #1823 addressing the same issue for building WRF-chem.

Solution:
Add missing prototypes; rearrange function order.

LIST OF MODIFIED FILES: list of changed files (use `git diff --name-status master` to get formatted list)
	modified:   arch/configure.defaults
	modified:   chem/KPP/kpp/kpp-2.1/src/code.c
	modified:   chem/KPP/kpp/kpp-2.1/src/code.h
	modified:   chem/KPP/kpp/kpp-2.1/src/code_c.c
	modified:   chem/KPP/kpp/kpp-2.1/src/code_f77.c
	modified:   chem/KPP/kpp/kpp-2.1/src/code_f90.c
	modified:   chem/KPP/kpp/kpp-2.1/src/code_matlab.c
	modified:   chem/KPP/kpp/kpp-2.1/src/gdata.h
	modified:   chem/KPP/kpp/kpp-2.1/src/gen.c
	modified:   chem/KPP/kpp/kpp-2.1/src/gen_org.c
	modified:   chem/KPP/kpp/kpp-2.1/src/kpp.c
	modified:   chem/KPP/kpp/kpp-2.1/src/scan.h
	modified:   chem/KPP/kpp/kpp-2.1/src/scan.l
	modified:   chem/KPP/kpp/kpp-2.1/src/scan.y
	modified:   chem/KPP/util/wkc/Makefile
	modified:   chem/KPP/util/wkc/change_chem_Makefile.c
	modified:   chem/KPP/util/wkc/compare_kpp_to_species.c
	modified:   chem/KPP/util/wkc/gen_kpp.c
	modified:   chem/KPP/util/wkc/gen_kpp_args_to_Update_Rconst.c
	modified:   chem/KPP/util/wkc/gen_kpp_interf_utils.c
	modified:   chem/KPP/util/wkc/gen_kpp_interface.c
	modified:   chem/KPP/util/wkc/gen_kpp_mech_dr.c
	modified:   chem/KPP/util/wkc/gen_kpp_utils.c
[...]

PR #1987, opened 2024-01-17
Add if test for ftn compiler version (for derecho) in compile file

kkeene44:
TYPE: no impact

KEYWORDS: compile, version, compiler, ftn, derecho

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
When compiling WRF on NCAR's Derecho HPC, and choosing a Cray option during configuration (for e.g., "INTEL (ftn/icc): Cray XC"), the compile script isn't able to determine the compiler version because "ftn" is not included in the "if" test to checks versions in the "compile" script. This means the compiler version is not printed in the compile log, which could be problematic for several reasons.

Solution:
Added a line to the "compile" file compiler version "if" test to print out the compiler version information when "ftn" is used.

LIST OF MODIFIED FILES: 
M    compile

TESTS CONDUCTED: 
1. Tested a compile when choosing "INTEL (ftn/icc): Cray XC" during configuration. Now the compiler version prints out to the compile log.

### DIFF
--- a/compile
+++ b/compile
@@ -414,6 +414,8 @@ else
     pgf90 --version
   else if ( "$comp[1]" == "ifort" ) then
     ifort -V
+  else if ( "$comp[1]" == "ftn" ) then
+    ftn -V
   else
     echo "Not sure how to figure out the version of this compiler: $comp[1]"
   endif


kkeene44:
@islas 
I added you as a reviewer in case you think there is a better way to do this, or in case I'm not considering something that could cause an issue down the road.

islas:
I think at this point it might be easier to write something more generic that doesn't rely on a hard-coded string. Maybe something like:
```
compVersion=$( $comp[1] --version 2>/dev/null || $comp[1] -V 2>/dev/null )
if [ -n "$compVersion" ]; then
  printf "%s\n" "$version"
else
  ...message about can't find version...
fi
```
That would, for any compiler, attempt first `--version` and then `-V` if that fails forwarding any error messages to `/dev/null` so as to not pollute the compiler version output. Using `printf` should faithfully reproduce the output as if the correct command had just be found without relying on non-POSIX standard `echo -e`

islas:
Okay, not as clean but I think the least convoluted way that doesn't involve different shell invocations, file redirection, or other fragile setups:
```csh
$comp[1] -V >& /dev/null
if ( $status == 0 ) then
  $comp[1] --version
else
  $comp[1] --version >& /dev/null
  if ( $status == 0 ) then
    $comp[1] -V
  else
    ...msg...
  endif
endif
```

kkeene44:
@islas 
Thank you so much for looking into this. The way you propose makes a lot more sense and keeps it universal. I'll modify the code with your suggestion and update the PR. Thanks, again!

kkeene44:
It looks like this was already taken care of by #1942, so I'm closing this.

PR #1743, opened 2022-06-18
check if USENETCDFPAR is empty first

wkliao:
Env variable USENETCDFPAR may not be set when running make.

TYPE: bug fix

KEYWORDS: USENETCDFPAR

SOURCE: Wei-keng Liao

DESCRIPTION OF CHANGES:
When env USENETCDFPAR is not set, running make spews the following error message.
```
/bin/sh: line 0: [: -eq: unary operator expected
```
Solution: Check if USENETCDFPAR  is set first, and then its value if set.

LIST OF MODIFIED FILES: Makefile

TESTS CONDUCTED:  running command make

PR #1988, opened 2024-01-19
Make sure that USENETCDFPAR is not undefined

MicroTed:
Some versions of make appear to fail on the USENETCDFPAR logic in the top level Makefile if the variable is undefined. Making sure it is at least set to 0 by the configure script fixes the issue.

TYPE: bug fix

KEYWORDS:  make

SOURCE: Ted Mansell (NOAA/NSSL)

DESCRIPTION OF CHANGES:
Problem: Logic failure in top level Makefile with some versions of make (don't remember which) if USENETCDFPAR is undefined

Solution:
Making sure that at USENETCDFPAR is set to either 0 or 1 in the configure script fixes the issue.

LIST OF MODIFIED FILES: configure

TESTS: it passed regression tests.

### DIFF
--- a/configure
+++ b/configure
@@ -219,6 +219,8 @@ if [ -n "$NETCDFPAR" ] ; then
   export NETCDF
   export NETCDF4
   export USENETCDFPAR
+else
+  export USENETCDFPAR=0
 fi
 
 if  test -z "$NETCDF"  ; then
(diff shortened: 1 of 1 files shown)


islas:
While this is technically no longer an issue since merging of #1743, I think setting the precedent to fixing the issue upstream and simply having the env var defined is good practice - especially when WRF is the one setting the variable to begin with.

islas:
@MicroTed Can you reword the title  the PR to be more descriptive of the issue?

For the description - I think it would be better to talk about the underlying issue which isn't so much about versions of `make` as it is about the `nc4_test` target previously using logic that assumed `USENETCDFPAR` had been set, leading to malformed syntax of `if [  -eq 0 ]; then...` in the recipe expansion.

MicroTed:
@islas Ah, OK, this was something I had encountered a while ago and didn't realize that the Makefile logic had been fixed. If this is therefore moot, it can just be closed.

MicroTed:
OK, sounds good. And I think the original issue was actually my fault, when the parallel netcdf4 option was added 😄

weiwangncar:
The regression test results:

    Failed Simulations are: 
    None
    None

islas (approved):

weiwangncar (approved):

PR #1997, opened 2024-01-28
Shared physics sfclayrev

weiwangncar:
TYPE: enhancement, new feature

KEYWORDS: shared physics, revised MM5 sfclay scheme

SOURCE: internal

DESCRIPTION OF CHANGES:
Add revised MM5 surface layer scheme from the shared physics directory, which is ccpp-compliant. Also added salinity effect in the shared physics subroutine.

Solution:
The CCPP-compliant revised MM5 surface layer code will be residing in phys/physics_mmm/ directory. Currently it is a copy of the code from the future MMM-physics repo, plus the salinity change made recently in WRF.

LIST OF MODIFIED FILES: 
M       arch/postamble
M       arch/preamble
M       clean
M       main/depend.common
M       phys/Makefile
M       phys/module_sf_sfclayrev.F
A       phys/physics_mmm/LICENSE
A       phys/physics_mmm/README.md
A       phys/physics_mmm/sf_sfclayrev.F90
A       phys/physics_mmm/sf_sfclayrev.meta

TESTS CONDUCTED: 
1. The runs before and after this change do not produce bit-for-bit results when compiled with -d option. The code does produce identical results when run with 36 vs 16 processors.

RELEASE NOTE: Add revised MM5 surface layer scheme from the future shared MMM-physics repository.


weiwangncar:
The regression test results:

    Failed Simulations are: 
    None
    None

smileMchen:
@weiwangncar 
For codes saved in  phys/physics_mmm/, in this PR it is "sf_sfclayrev.F90", can we change the name to "phys/physics_mmm/sf_sfclayrev.F" so that they are consistent with the name convention of other codes?  Another benefit is that we don't need to  modify arch/postamble and arch/preamble. 
Is there any special reason we need to name the code .F90?

weiwangncar:
@smileMchen Thanks for reviewing this PR. Unfortunately the decision to make CCPP-compliant physics code using .F90 is not ours, but the CCPP community. It will apply to MPAS as well.

dudhia (dismissed):

dudhia (dismissed):

phys/module_physics_init.F
```diff
@@ -2622,7 +2622,7 @@ SUBROUTINE bl_init(STEPBL,BLDT,DT,RUBLTEN,RVBLTEN,RTHBLTEN,  &
                                                                 ) !Optional oml
 !--------------------------------------------------------------------
    USE module_sf_sfclay
-   USE module_sf_sfclayrev
+   USE sf_sfclayrev
```
islas, line 2625: The depend.common line for this under module_physics_init should change to this new module

weiwangncar: It is already there. But module_sf_sfclayrev.o could be removed.

smileMchen (approved):

smileMchen (approved):

smileMchen (approved):

dudhia (approved):

PR #1986, opened 2024-01-16
[Resubmit for PR #1881] New option for SLUCM to use global distributed urban aerodynamic parameters

cenlinhe:
This is a re-submitted PR for https://github.com/wrf-model/WRF/pull/1881 to connect with latest Noah-MP GitHub commit and branch for v4.6 release. (Below is copied from https://github.com/wrf-model/WRF/pull/1881). For all review comments, please refer back to the original PR: https://github.com/wrf-model/WRF/pull/1881

New option for SLUCM to consider global distributed urban parameters

TYPE: new feature

KEYWORDS: SLUCM, urban parameters, anthropogenic heat

SOURCE: Do Ngoc Khanh (Tokyo Institute of Technology)

DESCRIPTION OF CHANGES:
This PR adds a new feature to WRF SLUCM by allowing consideration of spatially varying global distributed urban parameters and spatially hourly monthly varying anthropogenic heat.

LIST OF MODIFIED FILES:
M Registry/Registry.EM_COMMON
M Registry/registry.dimspec
M dyn_em/module_first_rk_step_part1.F
M dyn_em/module_initialize_real.F
M phys/module_pbl_driver.F
M phys/module_physics_init.F
M phys/module_sf_clm.F
M phys/module_sf_noahdrv.F
M phys/module_sf_urban.F
M phys/module_surface_driver.F
M phys/noahmp
M share/output_wrf.F

TESTS CONDUCTED:

The modification has been tested and used in previous publications.
Initial development: Varquez, A. C. G., Nakayoshi, M., & Kanda, M. (2015). The effects of highly detailed urban roughness parameters on a sea-breeze numerical simulation. Boundary-layer meteorology, 154, 449-469. https://doi.org/10.1007/s10546-014-9985-4
Global extension: Khanh, D. N., Varquez, A. C., & Kanda, M. (2023). Impact of urbanization on exposure to extreme warming in megacities. Heliyon, 9, e15511. https://doi.org/10.1016/j.heliyon.2023.e15511
Are the Jenkins tests all passing?
RELEASE NOTE: This modification add a two options (use_distributed_aerodynamics and distributed_ahe_opt) to WRF SLUCM (sf_urban_physics = 1) so that spatially varying urban morphological parameters (building height, plan area index, frontal area index, roughness length for momentum, and displacement height) can be considered.

PR #2005, opened 2024-02-07
Fix AHE option 2 and a problem with mosaic

epn09:
Related to PR#1881.

TYPE: bug fix

KEYWORDS: anthropogenic heat, mosaic

SOURCE: Do Ngoc Khanh (Tokyo Institute of Technology)

DESCRIPTION OF CHANGES:
Problem 1:
* In https://github.com/wrf-model/WRF/pull/1881#discussion_r1446523703, @cenlinhe said that `FRC_URB2D` should be `IN` (not `INOUT`) and removed a portion of code. See https://github.com/wrf-model/WRF/pull/1986/files#diff-4cf900c383221426961400b9c706d4a7f551833ccff64af20fef15ef6bdc0f1bL3721.
* However, that change causes `UTYPE_URB` passed to urban module to become 0 when mosaic is turn on, and array index out of bound error occurs.

Solution 1:
* As I said before, I don't know why FRC_URB2D and UTYPE should be set there. But apparently the model does not work if they are not set there (even when the new options are turned off and only standard SLUCM is used).
* In this PR, I revert that portion of change. Converting `FRC_URB2D` back to `INOUT`. 

Problem 2:
* @weiwangncar said that adding AH to HFX after PBL driver has no effect. https://github.com/wrf-model/WRF/pull/1986#discussion_r1480479849

Solution 2:
* I moved the addition part to surface driver, right after the computation of HFX.

This PR also removes a few unused variables in their declaration.

LIST OF MODIFIED FILES:

TESTS CONDUCTED: 
The Jenkins tests are all passing.

RELEASE NOTE: Fix AHE option 2 and a problem with mosaic

### DIFF
--- a/phys/module_sf_noahdrv.F
+++ b/phys/module_sf_noahdrv.F
@@ -2896,7 +2896,7 @@ SUBROUTINE lsm_mosaic(DZ8W,QV3D,P8W3D,T3D,TSK,                      &
      REAL, OPTIONAL, DIMENSION( ims:ime, jms:jme ), INTENT(OUT) :: AKMS_URB2D
 !
      REAL, DIMENSION( ims:ime, jms:jme ), INTENT(OUT) :: UST_URB2D
-     REAL, DIMENSION( ims:ime, jms:jme ), INTENT(INOUT) :: FRC_URB2D                ! change this to inout, danli mosaic
+     REAL, DIMENSION( ims:ime, jms:jme ), INTENT(IN) :: FRC_URB2D
      INTEGER, DIMENSION( ims:ime, jms:jme ), INTENT(IN) :: UTYPE_URB2D
 
 ! output variables urban --> lsm
@@ -3745,7 +3745,6 @@ SUBROUTINE lsm_mosaic(DZ8W,QV3D,P8W3D,T3D,TSK,                      &
                   IF (slucm_distributed_drag) THEN
                     IF (IVGTYP(I, J) == ISURBAN) THEN
                       UTYPE_URB = 2
-                      FRC_URB2D(I, J) = MAX(0.1, MIN(0.9, 1 - shdavg(I, J) / 100.))
                     END IF
                   ELSE IF (use_wudapt_lcz == 1) THEN
                     IF(IVGTYP(I,J)==ISURBAN) UTYPE_URB=5
@@ -3760,28 +3759,11 @@ SUBROUTINE lsm_mosaic(DZ8W,QV3D,P8W3D,T3D,TSK,                      &
                     IF(IVGTYP(I,J)==LCZ_9) UTYPE_URB=9
                     IF(IVGTYP(I,J)==LCZ_10) UTYPE_URB=10
                     IF(IVGTYP(I,J)==LCZ_11) UTYPE_URB=11
-
-
-                    IF(UTYPE_URB==1) FRC_URB2D(I,J)=1.
-                    IF(UTYPE_URB==2) FRC_URB2D(I,J)=0.99
-                    IF(UTYPE_URB==3) FRC_URB2D(I,J)=1.00
-                    IF(UTYPE_URB==4) FRC_URB2D(I,J)=0.65
-                    IF(UTYPE_URB==5) FRC_URB2D(I,J)=0.7
-                    IF(UTYPE_URB==6) FRC_URB2D(I,J)=0.65
-                    IF(UTYPE_URB==7) FRC_URB2D(I,J)=0.3
-                    IF(UTYPE_URB==8) FRC_URB2D(I,J)=0.85
-                    IF(UTYPE_URB==9) FRC_URB2D(I,J)=0.3
-                    IF(UTYPE_URB==10) FRC_URB2D(I,J)=0.55
-                    IF(UTYPE_URB==11) FRC_URB2D(I,J)=1.
                   ELSE
                     IF(IVGTYP(I,J)==ISURBAN) UTYPE_URB=2
                     IF(IVGTYP(I,J)==LCZ_1) UTYPE_URB=1 ! LOW_DENSITY_RESIDENTIAL
                     IF(IVGTYP(I,J)==LCZ_2) UTYPE_URB=2 ! HIGH_DENSITY_RESIDENTIAL
                     IF(IVGTYP(I,J)==LCZ_3) UTYPE_URB=3 ! HIGH_INTENSITY_INDUSTRIAL
-
-                    IF(UTYPE_URB==1) FRC_URB2D(I,J)=0.5
-                    IF(UTYPE_URB==2) FRC_URB2D(I,J)=0.9
-                    IF(UTYPE_URB==3) FRC_URB2D(I,J)=0.95
                   END IF
 
                   TA_URB    = SFCTMP           ! [K]
--- a/phys/module_sf_noahdrv.F
+++ b/phys/module_sf_noahdrv.F
@@ -2896,7 +2896,7 @@ SUBROUTINE lsm_mosaic(DZ8W,QV3D,P8W3D,T3D,TSK,                      &
      REAL, OPTIONAL, DIMENSION( ims:ime, jms:jme ), INTENT(OUT) :: AKMS_URB2D
 !
      REAL, DIMENSION( ims:ime, jms:jme ), INTENT(OUT) :: UST_URB2D
-     REAL, DIMENSION( ims:ime, jms:jme ), INTENT(IN) :: FRC_URB2D                ! change this to inout, danli mosaic
+     REAL, DIMENSION( ims:ime, jms:jme ), INTENT(INOUT) :: FRC_URB2D                ! change this to inout, danli mosaic
      INTEGER, DIMENSION( ims:ime, jms:jme ), INTENT(IN) :: UTYPE_URB2D
 
 ! output variables urban --> lsm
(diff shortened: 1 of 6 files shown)


dudhia:
Yes, it is needed in the surface driver.

phys/module_sf_noahdrv.F
```diff
@@ -3740,7 +3740,49 @@ SUBROUTINE lsm_mosaic(DZ8W,QV3D,P8W3D,T3D,TSK,                      &
                      IVGTYP(I,J) == LCZ_6      .or. IVGTYP(I,J) == LCZ_7 .or. IVGTYP(I,J) == LCZ_8 .or. &
                      IVGTYP(I,J) == LCZ_9      .or. IVGTYP(I,J) == LCZ_10 .or. IVGTYP(I,J) == LCZ_11 ) THEN
 
-                  UTYPE_URB = UTYPE_URB2D(I,J) !urban type (low, high or industrial)
+                  !  UTYPE_URB = UTYPE_URB2D(I,J) !urban type (low, high or industrial)
```
cenlinhe, line 3743: I still don't understand why this UTYPE_URB and FRC_URB2D are needed here to be re-assigned values. This will cause inconsistency. I remember that in @tslin2 test, he removed this and it worked fine. If this is not removed, then the UTYPE_URB and FRC_URB2D will be different from the input map.

tslin2: This should not happen, as I see the new implementation has been tested. https://github.com/wrf-model/WRF/pull/1986#issuecomment-1918313488

I will give a test when the machine is back

epn09: @tslin2 @epn09 If mosaic is turned off. Everything worked fine in my machine.

tslin2: It seems mosaic do all the calculations not only for dominant urban land cover grids. So if the grids are not dominant urban types but have some urban areas, there is no information for the urban type in the initial stage for these grids, so the UTYPE_URB2D(i,j) is initialed as zero for these grids.

https://github.com/wrf-model/WRF/blob/a8eb846859cb39d0acfd1d3297ea9992ce66424a/phys/module_sf_noahdrv.F#L3149

https://github.com/wrf-model/WRF/blob/a8eb846859cb39d0acfd1d3297ea9992ce66424a/phys/module_sf_noahdrv.F#L3191


perhaps it's better to use IVGTYP_dominant(I,J) here to only consider mosaic in urban dominant grids?!

or 

do the lsm_mosaic_init first, then urban_var_init in the module_physics_init.F

phys/module_sf_noahdrv.F
```diff
@@ -3740,7 +3740,49 @@ SUBROUTINE lsm_mosaic(DZ8W,QV3D,P8W3D,T3D,TSK,                      &
...
+                  ELSE
+                    IF(IVGTYP(I,J)==ISURBAN) UTYPE_URB=2
+                    IF(IVGTYP(I,J)==LCZ_1) UTYPE_URB=1 ! LOW_DENSITY_RESIDENTIAL
+                    IF(IVGTYP(I,J)==LCZ_2) UTYPE_URB=2 ! HIGH_DENSITY_RESIDENTIAL
+                    IF(IVGTYP(I,J)==LCZ_3) UTYPE_URB=3 ! HIGH_INTENSITY_INDUSTRIAL
+
+                    IF(UTYPE_URB==1) FRC_URB2D(I,J)=0.5
+                    IF(UTYPE_URB==2) FRC_URB2D(I,J)=0.9
+                    IF(UTYPE_URB==3) FRC_URB2D(I,J)=0.95
+                  END IF
```
tslin2, line 3785: I think it is okay to have UTYPE_URB here for those grids that are not dominant urban types (it's better to do urban_var_init). But not for frc_urb2d as the value comes from the geo file, so is different from the input map.

epn09: @tslin2 Will FRC_URB2D be read or set for grids which aren't dominant urban types if we remove it from here?

cenlinhe: @epn09 Yes, FRC_URB2D will be read from wrfinput file for all grids including grids that are not dominated by urban types. You could try to remove the FRC_URB2D part here and do a test to see if the model runs fine.

epn09: @tslin2 I'm not so sure about the default SLUCM, BEP-BEM, and LCZ, but for our new options, I think the `frc_urb2d` statement can be removed if the following
https://github.com/wrf-model/WRF/blob/3cadf04277ac3a050e65461efb6aa939349c60a8/dyn_em/module_initialize_real.F#L3140-L3142
is changed to
```fortran
IF (grid%landusef(i,grid%isurban,j) > 0) THEN 
```
I have tested it but I feel like it should work.

tslin2: > @epn09 Yes, FRC_URB2D will be read from wrfinput file for all grids including grids that are not dominated by urban types. You could try to remove the FRC_URB2D part here and do a test to see if the model runs fine.

yes it work only to set urban type here. In fact, my previous test only removed the FRC_URB2D. I test again today; setting the UTYPE_URB to the exact number is needed, not required to set the FRC_URB2D. This is because a single-layer urban model requires an urban type to read the parameter from the table; an urban type cannot be zero.

tslin2: > @tslin2 I'm not so sure about the default SLUCM, BEP-BEM, and LCZ, but for our new options, I think the `frc_urb2d` statement can be removed if the following
> 
> https://github.com/wrf-model/WRF/blob/3cadf04277ac3a050e65461efb6aa939349c60a8/dyn_em/module_initialize_real.F#L3140-L3142
> 
> 
> is changed to
> ```fortran
> IF (grid%landusef(i,grid%isurban,j) > 0) THEN 
> ```
> 
> I have tested it but I feel like it should work.

the frc_urb2d is not the issue here, but the urban type, UTYPE_URB. If the model has urban type 0, it cannot get the parameter from the urban table, and the model will crash. BEP and BEP-BEM are not yet compatible with mosaic.

tslin2: > @tslin2 Will FRC_URB2D be read or set for grids which aren't dominant urban types if we remove the statements from here?

the FRC_URB2D is first read from the wrfinput file, then set in the urban_var_init.
In urban_var_init, if the dominant land cover type is not urban, the FRC_URB2D is set to zero. 
So for the mosaic case, if we remove the FRC_URB2D here, for the grids that urban is not the dominant land cover type, the FRC_URB2D is zero, which means the impervious fraction is zero.

epn09:
@cenlinhe @tslin2 I have converted `INOUT` back to `IN` for FRC_URB2D and made a slight modification to initialize_real. For the new options, the output was identical.

epn09:
@cenlinhe @tslin2 Happy Chinese new year. Any updates on this?

cenlinhe:
The modifications look good to me.
@dudhia Are we going to include this PR to the v4.6 release or do you prefer to merge this PR for v4.6.1 minor release? Have we frozen the v4.6 release branch?

dudhia (dismissed):

dudhia:
There is still time to get this in the release. We have the test phase to consider any remaining questions.

phys/module_surface_driver.F
```diff
@@ -395,6 +396,7 @@ SUBROUTINE surface_driver(                                         &
    USE module_sf_tmnupdate
    USE module_sf_lake
    USE module_cpl, ONLY : coupler_on, cpl_rcv
+   use module_ra_gfdleta, only: cal_mon_day
```
weiwangncar, line 399: This requires adding dependency in main/depend.common.

epn09: @weiwangncar Please check if I've added the dependency correctly. I think this the `cal_mon_day` function can be move to some library module instead of radiation module. But let's save it for later.

epn09:
@weiwangncar @dudhia @cenlinhe Do you have any further concerns about this PR? I'll be on leave from Thursday.

phys/module_surface_driver.F
```diff
@@ -1455,14 +1459,10 @@ SUBROUTINE surface_driver(                                         &
 
    real,    optional,  dimension(ims:ime,jms:jme ),intent(inout) :: XLAIDYN
 ! IRRIGATION
-   INTEGER :: tloc, jmonth,timing  
-   REAL, PARAMETER    :: PI_GRECO=3.14159 
-   INTEGER  :: end_hour, irr_start,xt24,irr_day  
-   REAL :: constants_irrigation
```
dudhia, line 1461: OK, this looks like valid clean-up

dudhia (approved):

phys/module_surface_driver.F
weiwangncar: @dudhia @epn09 Why should these lines be removed for this PR?

epn09: @weiwangncar They are unused variables.

weiwangncar: @epn09 I understand. But we generally do not want un-related change to be put in a single PR. We may allow it for this time, if @dudhia and @cenlinhe are ok with it. If so, please document the change in your PR.

dudhia: I think as it has no effect, it can be combined but the PR should list it.

cenlinhe: I agree with Jimy.

weiwangncar (approved):

Issue #2007, opened 2024-02-09
mp_zero_out is applied to non-moist arrays, but should it be?


MicroTed:
README.namelist describes mp_zero_out as applying only to the 'moist' variables:

```
 mp_zero_out                    = 0,      ! no action taken, no adjustment to any moist field
                                = 1,      ! except for Qv, all other moist arrays are set to zero
                                                 if they fall below a critical value
                                = 2,      ! Qv is .GE. 0, all other moist arrays are set to zero
                                                 if they fall below a critical value
```

But in solve_em.F, **all** the 3D arrays (moist, scalar, chem, tracer) have mp_zero_applied with calls to microphysics_zero_outa and microphysics_zero_outb.

So the question is should this be applied to non-moist arrays? Or should the other arrays have separate flags and thresholds? (Or simply update the readme to say what it is actually doing.)

weiwangncar:
@dudhia Any thoughts?

dudhia:
Note that  by default this is off, but point taken. Not sure of any reason it should be applied to other 4d arrays, especially as it gives different treatment for p_qv. 
@MicroTed Do you have a reason you need to apply this switch for microphysics?

MicroTed:
@dudhia  I would not normally use it, but mp_zero_out is being used in the Warn-on-Forecast ensemble system to suppress noisy values along boundaries. We started doing some tests with 3-moment physics, though, and the reflectivity moments were getting zeroed out by the flag because they are O(-16 to -10). I am testing separate flags for each array (e.g., scalar_zero_out, scalar_zero_out_thresh, etc.) so that they can be controlled individually, which seems to be a reasonable solution. (This is working with 3.9, but 4.5 is the same.)

It seems that the original intent (based on the subroutines) was to apply only to the moist array. But perhaps somewhere along the line it was applied to all the arrays.

dudhia:
Yes, I agree and if there is no reason to use it for other arrays, we could just confine it to moist and not need other switches.

dudhia:
Thanks. So, it gave problems for WRF-Chem too. Limiting it to moist should be added as a PR.

MicroTed:
If you're thinking of just removing the extra calls to microphysics_zero_outa and microphysics_zero_outb in solve_em.F, that would be pretty easy and give the expected behavior. Would here or a PR be the place to solicit input from the WRF-CHEM folks?

dudhia:
[...]

PR #2010, opened 2024-02-22
 Makes mp_zero_out affect only the 'moist' array. New separate flags for scalar/chem/tracer

MicroTed:
TYPE:  bug fix

KEYWORDS: mp_zero_out

SOURCE: Ted Mansell (NOAA)

DESCRIPTION OF CHANGES:

This fixes the side-effect of mp_zero_out being applied not only to the moist array, but also to the scalar/chem/tracer arrays using the same threshold. This behavior would be unexpected from the documentation (readme) which indicated that only mixing ratios were affected.  This PR restricts mp_zero_out to the moist array and adds a separate mp_zero_out_all flag to apply it to all the arrays in the off chance that somebody needs to replicate the previous behavior.

ISSUE: Addresses https://github.com/wrf-model/WRF/issues/2007

LIST OF MODIFIED FILES: 
  Registry/Registry.EM_COMMON 
  dyn_em/solve_em.F 
  run/README.namelist
  wrftladj/solve_em_ad.F
  wrftladj/solve_em_tl.F

TESTS CONDUCTED: 
The Jenkins tests have passed.

RELEASE NOTE: The behavior of the  mp_zero_out flag was changed affect only the 'moist' array, whereas previously it also caused the scalar/chem/tracer arrays to also be set to zero for values below threshold. Now there is a separate flag (mp_zero_out_all) if one wishes to reproduce the old behavior.


dudhia:
Thanks. I would advocate eliminating the other new zero_out options completely. The existing routine has too many moisture references and even uses p_qv. Hard to anticipate anyone needing this for other 4d arrays, as I doubt it is used much anyway.

dudhia:
I understand your caution about backward compatibility. Another alternative is to leave the old calls in there but commented out in case someone wants to reactivate them easily.

MicroTed:
Or maybe just a 'mp_zero_out_all' flag to turn it on for all the arrays? (Basically the same as commenting out the extra calls)

dudhia:
I like that idea.

weiwangncar:
@MicroTed The regression tests are failing for DA/4DVAR compilation. When the changes are finalized, some of the registry changes should be added to registry.wrfplus and/or registry.var.

MicroTed:
@weiwangncar Thanks for noting that. I tend to forget about var.
@dudhia I'll make that change to a single extra flag -- much simpler

weiwangncar:
The regression test results:

    Failed Simulations are: 
    None
    None

weiwangncar (approved):

dudhia (approved):

PR #2020, opened 2024-03-12
PX LSM Update for MODIS 61 class LCZ 

coastwx:
Pleim-Xiu LSM Compatibility with MODIS LCZ NUM_LAND_CAT 61

TYPE: bug fix

KEYWORDS: MODIS, LCZ, P-X LSM, NUM_LAND_CAT, 61

SOURCE: Robert Gilliam, US EPA

DESCRIPTION OF CHANGES:
Problem: Approached that P-X LSM errored for MODIS LCZ 61 NUM_LAND_CAT configuration. 
 
Solution:
Logic checks in module_physics_init.F and module_sf_pxlsm.F were adjusted for 61 category inputs. P-X LSM data table, module_sf_pxlsm_data.F was updated for MODIS 61 categories. Default for LCZ 51-61 was set to MODIS urban class.

Fixes #123

LIST OF MODIFIED FILES: 
M       phys/module_sf_pxlsm.F
M       phys/module_sf_pxlsm_data.F

TESTS CONDUCTED: 
1. Tested 61 class LCZ with PX LSM for a 1 day simulation with updated codes. Ran base MODIS 21 class scheme with same code before and after LCZ update. The results were identical after a 24 hour simulation. This confirms updates do not impact other MODIS settings in the P-X LSM. The MODIS 21 was by nature not identical to the MODIS 61, but similar enough and differences follow underlying differences in MODIS datasets.
2. Are the Jenkins tests all passing? No Jenkins test

RELEASE NOTE: 
Pleim-Xiu LSM is now compatible with 61 category MODIS LCZ landuse dataset.


weiwangncar:
The regression test results:

    Failed Simulations are: 
    None
    None

weiwangncar:
@coastwx Is there a reason to put LANDUSEF, RA and RS in the history file? Note this will show up in all history output even if a user is not using PXLSM.

dudhia:
I think it needs to be done at your end.

weiwangncar:
@coastwx No need to close this PR. I was going to check, but I think RA and RS are only used with PX LSM. If so, they can be packaged on this line:
``
package   pxlsmscheme    sf_surface_physics==7
``
and as for LANDUSEF, without it being in the history file, I know it would fail when people use ndown using this scheme. So that may be an argument to add it to the history file.

If we agree that RA and RS can be packaged, you can simply do that in your local sandbox and do git push again.

coastwx:
Wei,

I'm in progress on this. Thanks for the guidance and sorry for any waste of time on your side.

For clarity, I'm going to add Jon Pleim's PX update from some testing over the last year. And a bug in PX where some NaNs were found in GRDFLX at Arctic grid cells where lakes or sea turn to ice. A soil parameter initially zero is used in the denominator of a calc.

Do you think it would be ok to include all in the update? It only changes a few lines within PX. Nothing outside of that module. If ok, I'll do in the morning after this last round of unit testing within the current master branch code. I was going to do two pull request. But now do not want to waste time of NCAR on two vs. one pull request.

Thanks,

Rob

dudhia:
I agree with Wei's ideas.

coastwx:
Wei,

I recompiled WRF with RA and RS added in the PX package def. Then ran a quick 1 day sim. RA and RS do not show in the output. I turned other variables in PX like VEGF_PX and LAI_PX off in the WRF output via Registry.EM_COMMON as well. None of these show up in the output. I was under the impression that the package def would put those variables in the output stream.

package   pxlsmscheme    sf_surface_physics==7       -             state:t2_ndg_new,q2_ndg_new,t2_ndg_old,q2_ndg_old,t2obs,q2obs,vegf_px,imperv,canfra,lai_px,wwlt_px,wfc_px,wsat_px,clay_px,csand_px,fmsand_px,ra,rs

Not a big deal if this does not work. We'll just have to be more clear on using the dynamic output control at runtime using iofields_filename.

Just wanted to get your view on the package issue in case I missed a step somewhere else. It seems these are just making variable available for the model and not necessarily output.

Thanks, I'll get the pull request issued today.

Rob

dudhia:
You still need the h in the Registry. The advantage of package is that
options not using PX will not have them.

Issue #1965, opened 2023-12-30
lcz cant use  sf_surface_physics=7? 


wbfaded:

wbfaded:
INPUT LandUse = "MODIFIED_IGBP_MODIS_NOAH"
 LANDUSE TYPE = "MODIFIED_IGBP_MODIS_NOAH" FOUND          61  CATEGORIES           2  SEASONS WATER CATEGORY =           17  SNOW CATEGORY =           15
-------------- FATAL CALLED ---------------
FATAL CALLED FROM FILE:  <stdin>  LINE:    3258
module_physics_init: PX LSM option requires USGS, MODIS, or NLCD

weiwangncar:
@coastwx Just wondering if you are interested in addressing the use of LCZ data issue.

coastwx:
Noted. I see that recent LANDUSE.TBL iterations have expanded the num_land_cat for NOAH variations for this LCZ paradigm. RIght now the PX LSM init stops the initialization if num_land_cat does not fit MMINLU that are defined in PX LSM internal landuse tables (phys/module_sf_pxlsm_data.F). These tables may still work even with larger num_land_cat that includes LCZ definitions and a lot of unassigned classes. It may just be adding the condition for the num_land_cat = 61.

To be clear though, this would not use any LCZ information in the PX LSM. We have some internal efforts adding some urban processes in the PX LSM, but that is not near release at this point. Any modification for the use of MODIFIED_IGBP_MODIS_NOAH 61 class would just be compatibility with Noah LSM MODIS 20 class table. 

I'll need to test and may have time this week, but cannot give a clear time line. This may be simple couple hours of testing with most of that setting up a MODIS test case. I could test the exact domain @wbfaded is attempting to run if provided namelist.wps and namelist.input files.

coastwx:
![image](https://github.com/wrf-model/WRF/assets/15362904/be25e06a-b3f2-48b8-aa9b-d2c1007206e1)

PX may work just fine if the condition for 61 landuse cats is added here. This is where I'll start FYI.

coastwx:
@wbfaded
@weiwangncar 

I'm starting to work on this issue. Need some details on the namelist that was used for LCZ initiation. I created a geogrid file using 

 geog_data_res = 'default',

for the MODIS landuse. I've created a one week case using UNGRIB & METGRID. 

Next step is the wrfinput_d01 file. 

1) Are the steps above correct for the case above?
[...]

PR #2023, opened 2024-03-19
Pxlsm lcz61 ii

coastwx:
Pleim-Xiu LSM MODIS LCZ Compatibility  & Surface Evaporation Update

TYPE: bug fix & physics refinement

KEYWORDS: MODIS, LCZ, P-X LSM, Latent Heat Flux

SOURCE: Robert Gilliam & Jon Pleim, US EPA

DESCRIPTION OF CHANGES:
Problem: 

1. User indicated that the P-X LSM errored for MODIS LCZ 61 NUM_LAND_CAT configuration.
2. Currently, we account for evaporation from transpiration, soil in both vegetated and non-veg parts, and wet leaves. But we only account for the latent heat effects on Tg from transpiration and evaporation from non-veg soil.
This fix adds latent heat effects on Tg from soil in vegetated parts and from wet leaves.
3. Fix for rare case where GRDFLX goes NaN because of a divide by zero based on a soil parameter when a water cell turns to sea ice.

Solution:
Logic checks in module_physics_init.F and module_sf_pxlsm.F were adjusted for 61 category inputs. P-X LSM data table, module_sf_pxlsm_data.F was updated for MODIS 61 categories. Default for LCZ 51-61 was set to MODIS urban class. We also added updates for the evaporation from vegetation and wet canopy. 

ISSUE: 
Fixes: [#1965](https://github.com/wrf-model/WRF/issues/1965)

LIST OF MODIFIED FILES:
M phys/module_physics_init.F
M phys/module_sf_pxlsm.F
M phys/module_sf_pxlsm_data.F

TESTS CONDUCTED:

- Tested 61 class LCZ with PX LSM for a 1 day simulation with updated codes. Ran base MODIS 21 class scheme with same code before and after LCZ update. The results were identical after a 24 hour simulation. This confirms updates do not impact other MODIS settings in the P-X LSM. The MODIS 21 was by nature not identical to the MODIS 61, but similar enough and differences follow underlying differences in MODIS datasets.

Followed with another longer July 2018 test of the update surface evaporation in the P-X LSM.
image

- The Jenkins tests are all passing.

RELEASE NOTE:
Pleim-Xiu LSM is now compatible with 61 category MODIS LCZ landuse dataset. A mode of latent heat effects on Tg from vegetated parts and from wet leaves is added to Pleim-Xiu LSM.


weiwangncar:
The regression test results:

    Failed Simulations are: 
    None
    None

coastwx:
Thanks Wei. Looks like it passed the testing.

With the reservations of adding RA and RS and sensitivity to other user's default output size, I just left the registry out of the pull request. Also thought about LANDUSEF more. Since our domains are static, this can be used from other inputs. MCIP that processes WRF for CMAQ has an option to use the geogrid file for these geographical data inputs. RA and RS are different obviously, but sensitive to other user's. We plan to update the P-X LSM run notes documentation with the other method of using text file to remove and add variables to the output. This is preferable anyhow because we can reduce the output size a lot.

weiwangncar:
@coastwx That works with us. Thanks!

coastwx:
This is great news. Now the packaged variables make more sense.

I’ll test and if appropriate I can do a simple pull request on that change.

Rob

dudhia (approved):
I will approve it. I mentioned it to @cenlinhe but I think it is OK.

weiwangncar (approved):

coastwx:
Super. Thanks Wei and Jimy for the advice and consultation.

PR #2025, opened 2024-03-21
Pleim-Xiu Package & Registry update for RA, RS, LANDUSEF, etc.

coastwx:
Key variables for CMAQ AQ modeling from the WRF Pleim-Xiu LSM are not default in WRF outputs when LSM=7

TYPE: bug fix

KEYWORDS: Pleim-Xiu, RA, RS, LANDUSEF, Registry

SOURCE: Robert Gilliam, US EPA

DESCRIPTION OF CHANGES:
RA, RS, LANDUSEF, T2OBS and Q2OBS are not default in WRF output when the Pleim-Xiu LSM package is used.

Solution:
We changed the Registry to add "h" next to variable declarations and added these to the "package   pxlsmscheme    sf_surface_physics==7" definition.

ISSUE: N/A

LIST OF MODIFIED FILES: 
Registry.EM_COMMON

TESTS CONDUCTED: 
1. We recompiled the code after clean -a and then ran a NOAH and P-X LSM simulation for a 3-hour run. Verified that P-X variables only show up in the WRF output when the Pleim-Xiu LSM is used. These variables were not in the NOAH run's output.
2. No Jenkins testing

RELEASE NOTE: Pleim-Xiu updated so all variable used by CMAQ are put in WRF outputs.


weiwangncar:
@coastwx Thanks for adding this. You can package RA and RS, but you cannot package LANDUSEF as other LSMs use it too. When you package something with a particular option, those arrays should only be used by that option alone.

weiwangncar:
@coastwx I edited your file to remove landusef from package.

weiwangncar:
The regression test results:

    Failed Simulations are: 
    None
    None

coastwx:
Sorry Wei. I was in meetings earlier. Perfect.

This is good. Thank you.

Rob

dudhia (approved):

weiwangncar:
@coastwx @dudhia It looks like RS is passed into Noah LSM and set there too. So it cannot be packaged with PXLSM.

dudhia:
I see RS also passed to NoahMP

dudhia:
Am I missing where they are used in PXLSM?

weiwangncar:
@dudhia Should we just package RA or should we let it go of this PR altogether?

dudhia:
An alternative is to name them specially like RA_PX and RS_PX and package those. There is a precedent in that LSM.

coastwx:
Thanks.

Renaming with the RA_PX and RS_PX would cause a chain of changes in the CMAQ system, MCIP in particular so not preferable on our end. Then potential issues with WRF versions in the future where RS is used for pre-v4.6 and RS_PX v4.6+. We'd probably just have to hope users know how to force in the WRF outputs either by the dynamic or compiled methods unless there is another good way. We generally account for this internally, but a lot of external groups have a lot less knowledge of WRF and tend to run out of the box.

Regards,

Rob

dudhia:
Renaming could be confined to the model down to the driver with internal physics names the same as before, but yes the output names would be changed, so we can leave it as is.

weiwangncar:
@coastwx Considering the complication of adding these variables to package, can we just not do this for now?

coastwx:
Sure. Disregard the pull request. I doubt there will be much enthusiasm to change the variable names in the WRF output. I'll discuss with our group and perhaps do this in later versions. It would require testing and the window is almost closed for this cycle. And we've gone many years the way it is now. Will update the PX LSM suggested run configuration document. I may be in touch very soon with an update to this hosted document.

https://www2.mmm.ucar.edu/wrf/users/docs/PX-ACM.pdf

weiwangncar:
@coastwx Thanks. I will close this PR and we will see what a future solution might be. Please send us any update of the document, probably before the end of April if possible.

PR #2029, opened 2024-03-25
Remove unneeded depends

weiwangncar:
TYPE: bug fix, clean up

KEYWORDS: dependencies, unused

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
Some dependency is declared twice, some are no longer needed.

Solution:
Remove un-used depends.

LIST OF MODIFIED FILES: 
M      main/depend.common

TESTS CONDUCTED: 
1. Tested on Derecho with 6 proc to compile

RELEASE NOTE:

### DIFF
--- a/main/depend.common
+++ b/main/depend.common
@@ -389,7 +389,6 @@ module_bl_camuwpbl_driver.o: \
 
 
 module_sf_mynn.o: \
-	module_sf_sfclay.o \
 	module_bl_mynn.o \
 	../share/module_model_constants.o \
 	../frame/module_wrf_error.o 
@@ -928,7 +927,6 @@ module_physics_init.o: \
 	module_ra_hs.o \
 	module_ra_flg.o \
 	module_sf_sfclay.o \
-	module_sf_sfclayrev.o \
 	physics_mmm/sf_sfclayrev.o \
 	module_sf_slab.o \
 	module_sf_myjsfc.o \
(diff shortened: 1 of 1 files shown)


main/depend.common
```diff
@@ -1417,10 +1415,6 @@ module_fdda_spnudging.o: \
 	../frame/module_wrf_error.o 
 
 
-module_sf_bep.o: \
```
islas, line 1420: I believe this is still needed, from `phys/module_sf_bep.F`:
```
MODULE module_sf_bep
#if defined(mpas)
use mpas_atmphys_utilities, only: physics_error_fatal
#define FATAL_ERROR(M) call physics_error_fatal( M )
#else
use module_wrf_error
#define FATAL_ERROR(M) call wrf_error_fatal( M)
!USE module_model_constants
#endif
 USE module_sf_urban
 USE module_bep_bem_helper, ONLY: nurbm
```

islas (approved):
Disregard previous comment, I saw after that the removed dep was a duplicate

weiwangncar:
The regression test results:

    Failed Simulations are: 
    None
    None

main/depend.common
```diff
@@ -1417,10 +1415,6 @@ module_fdda_spnudging.o: \
...
-	module_sf_urban.o 
-
-
```
dudhia, line 1423: can you explain why this one is removed?

weiwangncar: @dudhia This dependency is declared elsewhere in this file, hence redundant.

dudhia (approved):

PR #2031, opened 2024-03-26
Fix a logic for using specific humidity or mixing ratio directly

weiwangncar:
TYPE: bug fix

KEYWORDS: real, use_sh_qv

SOURCE: internal

DESCRIPTION OF CHANGES:
The logic to control whether specific humidity or mixing ratio is used instead of RH is wrong (PR-1959). It is corrected in this PR.

LIST OF MODIFIED FILES: 
M     dyn_em/module_initialize_real.F

TESTS CONDUCTED: 
1. Tested before and after this change. The data without SH is not affected, and the data with SH (EC model level data) is now used directly if use_sh_qv is set to true.

RELEASE NOTE:

### DIFF
--- a/dyn_em/module_initialize_real.F
+++ b/dyn_em/module_initialize_real.F
@@ -1110,6 +1110,7 @@ integer::oops1,oops2
             ELSE
                k = num_metgrid_levels
             END IF
+            config_flags%use_sh_qv = .FALSE.
 
             IF      ( config_flags%rh2qv_method .eq. 1 ) THEN
                CALL rh_to_mxrat1(grid%rh_gc, grid%t_gc, grid%p_gc, grid%qv_gc ,         &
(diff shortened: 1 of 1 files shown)


weiwangncar:
The regression test results:

    Failed Simulations are: 
    None
    None

dudhia:
So you don't need to test on those flags?

weiwangncar:
> So you don't need to test on those flags?

No.

dudhia (approved):

weiwangncar:
@kkeene44 @islas Can you review and approve this PR?

kkeene44 (approved):

kkeene44:
@weiwangncar 
I'm just curious what happened before the fix. Would something fail, down the line?

weiwangncar:
> @weiwangncar I'm just curious what happened before the fix. Would something fail, down the line?

real.exe would not fail. When the input has one of the SH or Qv input, it doesn't matter if one sets use_sh_qv or not, it will always use computed RH (if RH isn't in the input. If RH is in the input, it uses it too).

dudhia:
Does this fix anything in the last release or is it just fixing things added since?

weiwangncar:
@dudhia This change is only added for this release.

PR #2044, opened 2024-05-10
Updated global attributes for auxiliary output as well as runs with WRF-CMAQ

### LINKED ISSUE #1476: missing metadata in auxiliary history files
The suite of global attributes is incomplete in the auxiliary history files. I have a proposed fix that would be inserted into share/output_wrf.F. I made the change in the version in WRFv4.2.2 (which did not change from WRFv4.2.1, for example), and it works.

Please let me know if you'd like the updated code.

tlspero:
Updates global attributes in auxiliary history files and separately for WRF-CMAQ

TYPE: bug fix

KEYWORDS: metadata, global attributes, auxiliary output, auxiliary history, WRF-CMAQ

SOURCE: Tanya Spero and Jeff Willison (U.S. EPA)

DESCRIPTION OF CHANGES:
Problem:
The suite of global attributes is incomplete in auxiliary history files. Notably, the precipitation tipping bucket is among the fields not included. Separately, there is no metadata in the header to indicate when WRF-CMAQ is used.

Solution:
Changes are introduced in share/output_wrf.F to expand the use cases to write the full suite of global attributes to the metadata of auxiliary output files. In addition, two global attributes are added to the output only when the WRF-CMAQ model is invoked (WRF_CMAQ_OPTION and DIRECT_SW_FEEDBACK).

ISSUE: Fixes #1476 

LIST OF MODIFIED FILES: share/output_wrf.F

TESTS CONDUCTED: 
1. Tests were conducted using the WRF model both with and without compiling with CMAQ. The base code was pulled from the branch "release-v4.6.0" on 8 May 2024. In both sets of tests on the EPA supercomputing system, an auxiliary output file was created in auxhist7. The global attributes were properly expanded to include the full suite of attributes in the auxiliary output file. The two new WRF-CMAQ coupled model attributes appeared when the WRF-CMAQ model was invoked.
2.  Jenkins testing is all passing.

RELEASE NOTE: Expanded global attributes to the full suite for auxiliary history files. Also added two global attributes when the WRF-CMAQ model is invoked (WRF_CMAQ_OPTION and DIRECT_SW_FEEDBACK).

### DIFF
--- a/share/output_wrf.F
+++ b/share/output_wrf.F
@@ -638,6 +638,7 @@ endif
     ibuf(1) = wrf_cmaq_option
     CALL wrf_put_dom_ti_integer ( fid , 'WRF_CMAQ_OPTION' ,  ibuf , 1 , ierr )
     lbuf(1) = direct_sw_feedback
+! Note the code will turn logical T/F to integer 1/0 in the output
     CALL wrf_put_dom_ti_logical ( fid , 'DIRECT_SW_FEEDBACK' ,  lbuf , 1 , ierr )
 #endif
     ibuf(1) = mp_physics
--- a/share/output_wrf.F
+++ b/share/output_wrf.F
@@ -356,8 +356,6 @@
     CALL wrf_put_dom_ti_char ( fid , 'START_DATE', TRIM(start_date) , ierr )
     IF ( switch .EQ. input_only) THEN
        CALL wrf_put_dom_ti_char ( fid , 'SIMULATION_START_DATE', TRIM(start_date) , ierr )
-!!! Expanded to include auxiliary history files.  TLS 20 Jan 2021
-!!!    ELSE IF ( ( switch .EQ. restart_only ) .OR. ( switch .EQ. history_only ) ) THEN
     ELSE IF ( ( switch .EQ. restart_only ) .OR. ( switch .EQ. history_only ) .OR.  &
               ( ( switch .GE. auxhist1_only ) .AND. ( switch .LE. auxhist24_only ) ) ) THEN
        CALL nl_get_simulation_start_year   ( 1, simulation_start_year   )
@@ -564,8 +562,6 @@ endif
        END IF
     END IF
 
-!!! Expanded to include auxiliary history files.  TLS 20 Jan 2021
-!!!    IF (switch .EQ. history_only) THEN
     IF ( ( switch .EQ. history_only ) .OR.  &
          ( ( switch .GE. auxhist1_only ) .AND. ( switch .LE. auxhist24_only ) ) ) THEN
        CALL    wrf_put_dom_ti_integer( fid, 'SKEBS_ON'           , config_flags%skebs_on           , 1, ierr )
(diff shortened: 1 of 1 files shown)


tlspero:
Hi @islas!

I submitted this PR with the intent of moving into WRFv4.6, but I did not realize that you were prepping for release at the time that I put this forth. This PR has been idle, other than you reassigned it to WRFv4.6.1 and some regression tests are allegedly still "running". Please let me know how you'd like me to proceed on this. There are no reviewers assigned, and I think someone at NCAR needs to intervene to make that happen. This PR addresses an Issue that I opened three years ago but did not complete with Dave Gill.

--Tanya

share/output_wrf.F
```diff
@@ -810,8 +836,13 @@ SUBROUTINE output_wrf ( fid , grid , config_flags, switch , ierr )
 #endif
 
 #if (EM_CORE == 1) && ( DA_CORE != 1)
+!!! Expanded to include auxiliary history files.  TLS 20 Jan 2021
+!!!      IF ( ( switch .EQ. input_only   ) .OR. &
+!!!           ( switch .EQ. history_only ) .OR. &
+!!!           ( switch .EQ. restart_only ) ) THEN
```
weiwangncar, line 842: I'd suggest to remove these commented lines. One can retrieve old code fairly easily with git.

tlspero: Thank you, @weiwangncar!

Would you like me to remove my comments that include the nature of the change, e.g., L839 in the example above?

weiwangncar: @tlspero I'd say we can remove that too. Making the purpose of your change clear in the commit message is sufficient. Thanks!

tlspero:
Updated share/output_wrf.F to reflect comments from @weiwangncar.

weiwangncar:
@tlspero Does adding lbuf allow logical values to write out to the global attributes? Can you show an example? The reason I'm asking is that writing a logical namelist variable had not been an option in the past. If what you did here adds that function, that would be very helpful. We could potentially add more to the global attributes.

tlspero:
@weiwangncar --

We used the existing function, WRF_PUT_DOM_TI_LOGICAL, to add the logical variable to the global attributes. We had to create the local variable, LBUF, which we patterned after IBUF and RBUF. I believe the translation to the global attributes flipped the logical T/F to integer 1/0, but I would need to go back to recreate those output files to double-check. I think that we could continue using that function.

If you want me to recreate the output, I'll do that and add the output from "ncdump -h <file>" here. It will be next week, though, due to schedule conflict this weekend.

weiwangncar:
@tlspero Thanks! Next week would be fine.

weiwangncar:
The regression test results:

    Failed Simulations are: 
    None
    None

weiwangncar:
@tlspero Thanks..

weiwangncar:
@dudhia @islas If we cannot really define logical in global attributes, should we add lbuf and let the code turn it to integer, or should we use integer directly?

islas:
@weiwangncar I think leaving as logical is fine. Switching to integer opens up to multiple interpretations and various values on fields that are otherwise only one of two options. 

I think if there's an appropriate location we should just note that there is an assumption in what true and false correspond to as integers. The 1/0 to true/false interpretation is widely used but it's best to be explicit about any assumption/interpretations made.

weiwangncar (approved):

weiwangncar:
@dudhia Can you review this PR? Thanks!

dudhia:
I had the question about how 'switch' is used with GT, but I looked at it and it seems OK.

dudhia (approved):

PR #2053, opened 2024-05-28
Change internal flag organization in CMake build to not be global

islas:
TYPE: enhancement

KEYWORDS: cmake, flags, compilation

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
The current iteration of the cmake build places all configuration flags in the global properties of the project. While this works when just building WRF, integration with other projects' cmake builds if placed under WRF pollutes their respective build flags.

Solution:
Adjust the layout of the flags to instead carry them in a variable, and prefer using `target_*` calls for flag usage.

LIST OF MODIFIED FILES: 
M       CMakeLists.txt
M       chem/CMakeLists.txt
M       external/CMakeLists.txt
M       external/io_adios2/CMakeLists.txt
M       external/io_netcdf/CMakeLists.txt
M       external/io_netcdfpar/CMakeLists.txt
M       external/io_pnetcdf/CMakeLists.txt
M       frame/CMakeLists.txt
M       main/CMakeLists.txt
M       phys/CMakeLists.txt
M       tools/CMakeLists.txt

RELEASE NOTE: 
Change internal flag organization in CMake build to not be global

PR #2055, opened 2024-05-31
Fix fseek test

islas:
TYPE: bug fix

KEYWORDS: fseek, compilation, cmake

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
The fseek test lacks correct syntax causing false negative reports of feature not existing when newer compiler standards disallow this.

Solution:
Add `int` to the main program in the fseek test. Also to add further robustness in detecting this feature, use the `-D_FILE_OFFSET_BITS=64 -D_LARGEFILE_SOURCE=1` generally *nix standard defines.

LIST OF MODIFIED FILES: 
M       CMakeLists.txt
M       confcheck/CMakeLists.txt
M       tools/fseek_test.c

RELEASE NOTE: Fix fseek test

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -792,10 +792,13 @@ endif()
 #!TODO Leaving as is in WRF for now but investigate why we don't do this
 # https://stackoverflow.com/a/1035713
 # If fseeko64 succeeds, use that, else check if we can fall back to fseeko, and if not just use fseek
-if ( ${FSEEKO64} )
-  list( APPEND PROJECT_COMPILE_DEFINITIONS FSEEKO64_OK )
-elseif( "${FSEEKO}" )
-  list( APPEND PROJECT_COMPILE_DEFINITIONS FSEEKO_OK )
+if ( "${FSEEKO64}" OR "${FSEEKO}" )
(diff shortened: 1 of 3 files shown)


islas:
@Plantain Thanks for the catch! The issue was some assumptions made in the use of cmake generator expressions. PR #2056 should fix that issue so that this will build properly.

islas:
Requires #2056 and #2053

amstokely (approved):
I’ll approve these changes, but I noticed that the use of void main() in fseek_test.c is not ISO C compliant. Is there any reason why the contents of main() can’t be moved to a new function, which can then be called from an int main()? This wouldn't alter the current code's behavior and would make it ISO C compliant.

islas:
> the use of void main() in fseek_test.c is not ISO C compliant

I didn't actually know that `void` return on main wasn't ISO C. I think we don't even need to move it to a function if instead of the `exit( retval )` call we replace that with `return retval`

mgduda:
@islas Can you update the PR description to note that `main` returns an `int` rather than `void`?

mgduda (dismissed):

islas:
Rebased to v4.6.1 to resolve conflicts from duplicate changes

mgduda (approved):

amstokely (approved):

PR #2056, opened 2024-05-31
CMake build no longer uses generator expressions in defines

islas:
TYPE: bug fix

KEYWORDS: cmake, compilation

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
The use of generator expressions in the defines compacts the logic neatly but removes the ability to evaluate these conditionals at configuration time. As such, assumptions must either be made or defines wholly dropped when adding configure-time commands like C preprocessing, both of which are wrong.

Solution:
Switch the logic to a more verbose `if()`-style that guarantees defines that can be known at configure time are resolved. 

LIST OF MODIFIED FILES: 
M       CMakeLists.txt
M       cmake/c_preproc.cmake

RELEASE NOTE: CMake build no longer uses generator expressions in defines

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -619,55 +619,8 @@ add_compile_definitions(
                         DWORDSIZE=${DWORDSIZE}
                         LWORDSIZE=${LWORDSIZE}
                         RWORDSIZE=${RWORDSIZE}
-                        # Only define if set, this is to use #ifdef/#ifndef preprocessors
-                        # in code since cmake cannot handle basically any others :(
-                        # https://gitlab.kitware.com/cmake/cmake/-/issues/17398
-                        $<$<BOOL:${ENABLE_CHEM}>:WRF_CHEM=$<BOOL:${ENABLE_CHEM}>>
-                        $<$<BOOL:${ENABLE_CHEM}>:BUILD_CHEM=$<BOOL:${ENABLE_CHEM}>>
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -619,89 +619,189 @@ add_compile_definitions(
...
+#!TODO Leaving as is in WRF for now but investigate why we don't do this
+# https://stackoverflow.com/a/1035713
+# If fseeko64 succeeds, use that, else check if we can fall back to fseeko, and if not just use fseek
+if ( NOT ${FSEEKO64} )
+  add_compile_definitions( FSEEKO64_OK )
+elseif( "${FSEEKO}" )
+  add_compile_definitions( FSEEKO_OK )
+else()
+  add_compile_definitions( FSEEK_OK )
+endif()
```
amstokely, line 798: Instead of calling add_compile_definitions for every "compile definition", I would append all compile definitions to a list and then call add_compile_definitions once using that list. At least for me, this would make the code easier to follow and it would be easy to print out all compile definitions via message.

amstokely:
These changes look good to me. My only suggestion is to use target_compile_definitions where possible and to group related compile definitions into a list before passing them to add_compile_definitions or target_compile_definitions. However, I see that you've already implemented this approach in [PR 2053](https://github.com/wrf-model/WRF/pull/2053)

amstokely (approved):
All of these changes look good to me.

mgduda (approved):

PR #2059, opened 2024-06-07
Remove unsupported -auto icx flag from Intel oneAPI stanza

islas:
TYPE: bug fix

KEYWORDS: intel, compilation, oneapi

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
The `-auto` flag is not supported by the Intel C compilers. See https://www.intel.com/content/www/us/en/docs/cpp-compiler/developer-guide-reference/2021-10/alphabetical-option-list.html for full list of supported icx flags.

Solution:
Remove the flag, no output changes should happen as the flag was previously being ignored if it didn't cause an error.


weiwangncar:
The regression test results:

    Failed Simulations are: 
    None
    None

weiwangncar:
@islas For an older version of Intel C compiler, would removing -auto cause any trouble?

islas:
@weiwangncar I can't find any reference to `-auto` being part of the icc compiler, so I suspect removing it would have no impact. Also, on Derecho I get this warning as well :
`icc: command line warning #10006: ignoring unknown option '-auto'`

weiwangncar (approved):

mgduda (approved):
I can confirm that the `icx` compiler rejects the `-auto` option:
```
icx: error: unknown argument: '-auto'
```

PR #2060, opened 2024-06-07
Fix line continuation syntax causing problems with certain compilers

islas:
TYPE: bug fix

KEYWORDS: syntax, compilation

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
dyn_em/module_big_step_utilities_em.F contains nonstandard line continuation symbols that are most likely being ignored by the make build system by the in situ file preprocessing steps. The cmake build does not use sed or standard.exe to preprocess files, so these lines get fed directly to compilers. Fortran compilers that do not support C-style line continuation then fail (nvhpc as an example).

Solution:
Change line continuation characters to standard Fortran.


weiwangncar:
The regression test results:

    Failed Simulations are: 
    None
    None

weiwangncar (approved):

mgduda (approved):

PR #1944, opened 2023-12-04
Add three new wind farm parameterizations and their ensembles 

YulongMa:
Add three new wind farm parameterizations and their ensembles

TYPE: new feature

KEYWORDS: wind farm parameterization, windfarm option

SOURCE: Cristina Archer (UDEL), Yulong Ma(UDEL;GWA-MWF) and Ahmad Vasel-Be-Hagh(Tennessee Technological University)

DESCRIPTION OF CHANGES:
Three new wind farm parameterizations and their ensembles have been added. The key innovation of these parameterizations is their ability to directly account for the individual and overlapping sub-grid wakes of wind turbines within a wind farm. This feature was absent in the Fitch parameterization previously used in WRF. Additionally, the three parameterizations differ in their representation of wakes (e.g., top-hat or Gaussian) and their superposition methods (e.g., sum of squared deficits, squared velocities, or empirical fittings).
The new namelist option windfarm_wake_model can be used to select one of the three wind farm parameterizations, the windfarm_overlap_method namelist option can be used to select the desired wake superposition method.
We recommend to use the new wind farm parameterizations, particularly for coarse resolution, high turbine density, and wind directions aligned with the turbine columns.

LIST OF MODIFIED FILES: 
dyn_em/module_first_rk_step_part1.F
phys/module_pbl_driver.F
phys/module_physics_init.F
phys/module_wind_jensen.F
phys/Makefile
Registry/Registry.EM_COMMON
run/README.namelist

TESTS CONDUCTED: 
1. Code tested as shown in references.
2. The Jenkins tests are all passing.

RELEASE NOTE: 
This PR adds options for three new wind farm parameterizations and their ensembles to account for the individual and overlapping sub-grid wakes of wind turbines within a wind farm. It is recommended to use the new wind farm parameterizations, particularly for coarse resolution, high turbine density, and wind directions aligned with the turbine columns. References:
Ma, Yulong, Cristina L. Archer, and Ahmadreza Vasel-Be-Hagh. "The Jensen wind farm parameterization." Wind Energy Science 7.6 (2022): 2407-2431.
Ma, Yulong, Cristina L. Archer, and Ahmad Vasel‐Be‐Hagh. "Comparison of individual versus ensemble wind farm parameterizations inclusive of sub‐grid wakes for the WRF model." Wind Energy 25.9 (2022): 1573-1595.

PR #2061, opened 2024-06-07
Replace nonstandard isnan() call with value input checks instead

islas:
TYPE: bug fix

KEYWORDS: compilation, nonstandard

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
PR #1944 introduced a nonstandard function call `isnan()` which breaks compilation for compilers which don't support this extension. 

Solution:
A potential alternative is to use the `ieee_is_nan()` call from `ieee_arithmetic`. As there is conditional compilation of this module which requires Fortran 2003 standard, an easier alternative is just to check the input values before calling `acos()`

### DIFF
--- a/phys/module_wind_mav.F
+++ b/phys/module_wind_mav.F
@@ -1231,11 +1231,11 @@ MODULE module_wind_mav
       end if
 
       tmp1 = tmp1 / tmp2
-      if ( tmp1 < -1 .or. tmp1 > 1 ) then
-        angle = 0
+      if ( tmp1 < -1.0_8 .or. tmp1 > 1.0_8 ) then
+        angle = 0.
       else
         angle = real(acos(tmp1), kind = 4)
-      endif
+      end if
  
       if (abs(angle) <= sr_angle) then
           ft = .true.
--- a/phys/module_wind_mav.F
+++ b/phys/module_wind_mav.F
@@ -1231,10 +1231,10 @@ MODULE module_wind_mav
       end if
 
       tmp1 = tmp1 / tmp2
-      if ( tmp1 < -1.0_8 .or. tmp1 > 1.0_8 ) then
-        angle = 0.
+      if (tmp1 < -1.0_8 .or. tmp1 > 1.0_8) then
+          angle = 0.
       else
-        angle = real(acos(tmp1), kind = 4)
+          angle = real(acos(tmp1), kind = 4)
       end if
  
       if (abs(angle) <= sr_angle) then
--- a/phys/module_wind_mav.F
+++ b/phys/module_wind_mav.F
@@ -1231,7 +1231,7 @@ MODULE module_wind_mav
       end if
 
       tmp1 = tmp1 / tmp2
-      if ( tmp1 < -1 .or. tmp1 > 1 )
+      if ( tmp1 < -1 .or. tmp1 > 1 ) then
         angle = 0
       else
         angle = real(acos(tmp1), kind = 4)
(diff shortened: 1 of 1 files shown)


weiwangncar:
The regression test results:

    Failed Simulations are: 
    None
    None

weiwangncar (dismissed):

phys/module_wind_mav.F
```diff
@@ -1230,11 +1230,12 @@ function find_turb(xc, yc, xt, yt, u, v, sr_angle, sr_dis)  result(ft)
...
-
-      if (isnan(angle)) then
-         angle = 0.
-      end if
+      tmp1 = tmp1 / tmp2
+      if ( tmp1 < -1 .or. tmp1 > 1 ) then
+        angle = 0
+      else
+        angle = real(acos(tmp1), kind = 4)
+      endif
```
mgduda, line 1238: To match the style of the rest of the code, I'd suggest `end if` here.

phys/module_wind_mav.F
```diff
@@ -1230,11 +1230,12 @@ function find_turb(xc, yc, xt, yt, u, v, sr_angle, sr_dis)  result(ft)
...
       end if
 
-      angle = real(acos(tmp1/tmp2), kind = 4)
-
-      if (isnan(angle)) then
-         angle = 0.
-      end if
+      tmp1 = tmp1 / tmp2
+      if ( tmp1 < -1 .or. tmp1 > 1 ) then
+        angle = 0
```
mgduda, line 1235: If we're assigning to a `real`, I think it would be preferable to uses `0.` rather than `0`.

phys/module_wind_mav.F
```diff
@@ -1230,11 +1230,12 @@ function find_turb(xc, yc, xt, yt, u, v, sr_angle, sr_dis)  result(ft)
           tmp2 = sign(tmp1,tmp2)
       end if
 
-      angle = real(acos(tmp1/tmp2), kind = 4)
-
-      if (isnan(angle)) then
-         angle = 0.
-      end if
+      tmp1 = tmp1 / tmp2
+      if ( tmp1 < -1 .or. tmp1 > 1 ) then
```
mgduda, line 1234: I think it would be preferable to use `-1._8` and `1._8`, since `tmp1` is a `real (kind = 8)` variable.

mgduda (changes requested):
As a final comment, I wonder whether the original code is correct? Would setting `angle = pi` be better when `tmp1 / tmp2` is less than `-1`?

islas:
> As a final comment, I wonder whether the original code is correct? Would setting `angle = pi` be better when `tmp1 / tmp2` is less than `-1`?

I'm unsure. That does make sense, but  I was hesitant to modify the logic in a way that deviates from the original. Since the final result is a `logical`, there are probably more implications [than I understand] based on what a close-to or actual out of bounds evaluation should be.

mgduda:
Agreed that we shouldn't try to "correct" the computation, as we may be missing some important details.

As a final comment, I'd suggest trying to match the whitespace convention used in the rest of the `find_turb`: four spaces of indentation, and no space between the parentheses in conditional statements.

islas:
> trying to match the whitespace convention used in the rest of the `find_turb`: four spaces of indentation, and no space between the parentheses in conditional statements.

Fixed, though I do wish we had a format checker, as a significant portion of the file's code (and the original logic that this PR replaces) don't follow this.

mgduda (approved):

weiwangncar (approved):

PR #2062, opened 2024-06-10
Add explicit casts to mismatched pointer types in rsl_lite

### LINKED ISSUE #2047: GCC 14 leads to compile errors in WRF/WPS compilation
Hej there. 
I am maintaining a project to provide wrf/wps binaries for deployment in an ArchLinux environment.

**Describe the bug**
With the update to gcc 14 several new changes regarding compile flags were done.
These lead to a more restrict error handling since several flags that would only lead to warnings now lead to compile errors:

Looking on the changes with gcc 14 it seems they have change the flag: [GCC 14](https://gcc.gnu.org/gcc-14/porting_to.html)

> Implicit function declarations (-Werror=implicit-function-declaration)
> It is no longer possible to call a function that has not been declared. In general, the solution is to include a header file with an appropriate function prototype. Note that GCC will perform further type checks based on the function prototype, which can reveal further type errors that require additional changes.
> 
> For well-known functions declared in standard headers, GCC provides fix-it hints with the appropriate #include directives:

```
error: implicit declaration of function ‘strlen’ [-Wimplicit-function-declaration]
    5 |   return strlen (s);
      |          ^~~~~~
note: include ‘<string.h>’ or provide a declaration of ‘strlen’
  +++ |+#include <string.h>
    1 |
```

When compiling the wrf model now these new restriction lead to these compile errors, e.g.
```
gcc  -I. -w -O3 -c   -DDM_PARALLEL -DLANDREAD_STUB=1 -DMAX_HISTORY=25 -DNMM_CORE=0   -c get_region_center.c
get_region_center.c: In function ‘get_region_center_’:
get_region_center.c:40:3: error: implicit declaration of function ‘memcpy’ [-Wimplicit-function-declaration]
   40 |   memcpy(MemoryOrder,MemoryOrderIn,strlen1);
      |   ^~~~~~
```

**To Reproduce**
Testing environment is an ArchLinux VM  with Linux Kernel 6.8.9.
Compiler is gcc/gfortran with version [14.1.1](https://archlinux.org/packages/core/x86_64/gcc/).
Current compile flags: [Env. Variables](https://github.com/SettRaziel/wrf_archlinux/blob/development/wrf_install/linux/set_env.sh)
Additional parameter for compilation:  35 gfortran dm+sm 
Precondition: Successful compilation of netcdf, netcdf-fortran, mpich, hdf5
Running my compile routines lead to a reproducible number of around 140 of these implicit compile errors.

**Workaround**
Adding to the compile flags: -Wimplicit-function-declaration (issue tracked in: [wrf_archlinux](https://github.com/SettRaziel/wrf_archlinux/issues/135))

**Expected behavior**
No implicit functions declaration throughout the code resolving the workaround to address these issues as warnings and not as errors with the flag which might lead to other side effects during the code compilation.

**Additional context**
Up to gcc/gfortran 13.2.2 these implicit declarations only lead to warnings, so up to that point the compilation runs successfully.
Are there plans to refactor the code base?

islas:
TYPE: bug fix

KEYWORDS: gcc14, compilation, c99

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
From #2047 :
Compilation of WRF v4.6.0 fails with GCC 14 due to new standard changes

> Type checking on pointer types (-Werror=incompatible-pointer-types)
> GCC no longer allows implicitly casting all pointer types to all other pointer types. This behavior is now restricted to the void * type and its qualified variations.
> 
> To fix compilation errors resulting from that, you can add the appropriate casts, and maybe consider using void * in more places (particularly for old programs that predate the introduction of void * into the C language).
> 
> Programs that do not carefully track pointer types are likely to contain aliasing violations, so consider building with -fno-strict-aliasing. (Whether casts are written manually or performed by GCC automatically does not make a difference in terms of strict aliasing violations.)
> 
> A frequent source of incompatible function pointer types involves callback functions that have more specific argument types (or less specific return types) than the function pointer they are assigned to. For example, old code which attempts to sort an array of strings might look like this: 
```
#include <stddef.h>
#include <stdlib.h>
#include <string.h>

int
compare (char **a, char **b)
{
  return strcmp (*a, *b);
}

void
sort (char **array, size_t length)
{
  qsort (array, length, sizeof (*array), compare);
}
```
Example of failure : 
```
/home/aislas/wrf-model/wrf/external/RSL_LITE/c_code.c: In function ‘rsl_lite_pack_’:
/home/aislas/wrf-model/wrf/external/RSL_LITE/c_code.c:487:27: error: passing argument 1 of ‘f_pack_lint_’ from incompatible pointer type [-Wincompatible-pointer-types]
  487 |             F_PACK_LINT ( buf, p+yp_curs, imemord, &js, &je, &ks, &ke, &is, &ie,
      |                           ^~~
      |                           |
      |                           char *
In file included from /home/aislas/wrf-model/wrf/external/RSL_LITE/c_code.c:31:
/home/aislas/wrf-model/wrf/external/RSL_LITE/rsl_lite.h:200:25: note: expected ‘long int *’ but argument is of type ‘char *’
  200 | void F_PACK_LINT (long *inbuf, long *outbuf, int* memorder, int* js, int* je, int* ks, int* ke, int* is, int* ie, int* jms, int* jme, int* kms, int* kme, int* ims, int* ime, int* curs);
      |                   ~~~~~~^~~~~
/home/aislas/wrf-model/wrf/external/RSL_LITE/c_code.c:487:33: error: passing argument 2 of ‘f_pack_lint_’ from incompatible pointer type [-Wincompatible-pointer-types]
  487 |             F_PACK_LINT ( buf, p+yp_curs, imemord, &js, &je, &ks, &ke, &is, &ie,
      |                                ~^~~~~~~~
      |                                 |
      |                                 char *
/home/aislas/wrf-model/wrf/external/RSL_LITE/rsl_lite.h:200:38: note: expected ‘long int *’ but argument is of type ‘char *’
  200 | void F_PACK_LINT (long *inbuf, long *outbuf, int* memorder, int* js, int* je, int* ks, int* ke, int* is, int* ie, int* jms, int* jme, int* kms, int* kme, int* ims, int* ime, int* curs);
      |                                ~~~~~~^~~~~~
/home/aislas/wrf-model/wrf/external/RSL_LITE/c_code.c:492:26: error: passing argument 1 of ‘f_pack_int_’ from incompatible pointer type [-Wincompatible-pointer-types]
  492 |             F_PACK_INT ( buf, p+yp_curs, imemord, &js, &je, &ks, &ke, &is, &ie,
      |                          ^~~
      |                          |
      |                          char *
/home/aislas/wrf-model/wrf/external/RSL_LITE/rsl_lite.h:201:23: note: expected ‘int *’ but argument is of type ‘char *’
  201 | void F_PACK_INT (int *inbuf, int *outbuf, int* memorder, int* js, int* je, int* ks, int* ke, int* is, int* ie, int* jms, int* jme, int* kms, int* kme, int* ims, int* ime, int* curs);
      |                  ~~~~~^~~~~
```

Error was reproduced AlmaLinux release 9.4 with GCC 14.1.0

Solution:
Use explicit type casting

ISSUE: 
Fixes #2047 

TESTS CONDUCTED: 
1. Tested on GCC 14.1.0

### DIFF
--- a/external/RSL_LITE/c_code.c
+++ b/external/RSL_LITE/c_code.c
@@ -484,12 +484,12 @@ RSL_LITE_PACK ( int * Fcomm0, char * buf , int * shw0 ,
 	    MPI_Abort(MPI_COMM_WORLD, 99) ;
           }
           if ( typesize == 8 ) {
-            F_PACK_LINT ( (long *)buf, (long *)p+yp_curs, imemord, &js, &je, &ks, &ke, &is, &ie, 
+            F_PACK_LINT ( (long *)buf, (long *)(p+yp_curs), imemord, &js, &je, &ks, &ke, &is, &ie, 
                                                 &jms,&jme,&kms,&kme,&ims,&ime, &wcount ) ;
             yp_curs += wcount*typesize ;
           }
 	  else if ( typesize == 4 ) {
-            F_PACK_INT ( (int * )buf, (int * )p+yp_curs, imemord, &js, &je, &ks, &ke, &is, &ie,
+            F_PACK_INT ( (int * )buf, (int * )(p+yp_curs), imemord, &js, &je, &ks, &ke, &is, &ie,
                                                &jms,&jme,&kms,&kme,&ims,&ime, &wcount ) ;
             yp_curs += wcount*typesize ;
 	  }
@@ -505,12 +505,12 @@ RSL_LITE_PACK ( int * Fcomm0, char * buf , int * shw0 ,
           ks = kps           ; ke = kpe ;
           is = IMAX(ips-shw) ; ie = IMIN(ipe+shw) ;
           if ( typesize == 8 ) {
-            F_UNPACK_LINT ( (long *)p+yp_curs, (long *)buf, imemord, &js, &je, &ks, &ke, &is, &ie,
+            F_UNPACK_LINT ( (long *)(p+yp_curs), (long *)buf, imemord, &js, &je, &ks, &ke, &is, &ie,
                                                &jms,&jme,&kms,&kme,&ims,&ime, &wcount ) ;
             yp_curs += wcount*typesize ;
           }
 	  else if ( typesize == 4 ) {
-            F_UNPACK_INT ( (int * )p+yp_curs, (int * )buf, imemord, &js, &je, &ks, &ke, &is, &ie,
+            F_UNPACK_INT ( (int * )(p+yp_curs), (int * )buf, imemord, &js, &je, &ks, &ke, &is, &ie,
                                                &jms,&jme,&kms,&kme,&ims,&ime, &wcount ) ;
             yp_curs += wcount*typesize ;
 	  }
@@ -538,12 +538,12 @@ RSL_LITE_PACK ( int * Fcomm0, char * buf , int * shw0 ,
 	    MPI_Abort(MPI_COMM_WORLD, 99) ;
           }
           if ( typesize == 8 ) {
-            F_PACK_LINT ( (long *)buf, (long *)p+ym_curs, imemord, &js, &je, &ks, &ke, &is, &ie,
+            F_PACK_LINT ( (long *)buf, (long *)(p+ym_curs), imemord, &js, &je, &ks, &ke, &is, &ie,
                                                &jms,&jme,&kms,&kme,&ims,&ime, &wcount ) ;
             ym_curs += wcount*typesize ;
           }
 	  else if ( typesize == 4 ) {
-            F_PACK_INT ( (int * )buf, (int * )p+ym_curs, imemord, &js, &je, &ks, &ke, &is, &ie,
+            F_PACK_INT ( (int * )buf, (int * )(p+ym_curs), imemord, &js, &je, &ks, &ke, &is, &ie,
                                                &jms,&jme,&kms,&kme,&ims,&ime, &wcount ) ;
             ym_curs += wcount*typesize ;
   	  }
@@ -559,12 +559,12 @@ RSL_LITE_PACK ( int * Fcomm0, char * buf , int * shw0 ,
           ks = kps           ; ke = kpe ;
           is = IMAX(ips-shw) ; ie = IMIN(ipe+shw) ;
           if ( typesize == 8 ) {
-            F_UNPACK_LINT ( (long *)p+ym_curs, (long *)buf, imemord, &js, &je, &ks, &ke, &is, &ie,
+            F_UNPACK_LINT ( (long *)(p+ym_curs), (long *)buf, imemord, &js, &je, &ks, &ke, &is, &ie,
                                                   &jms,&jme,&kms,&kme,&ims,&ime, &wcount ) ;
             ym_curs += wcount*typesize ;
           }
 	  else if ( typesize == 4 ) {
-            F_UNPACK_INT ( (int * )p+ym_curs, (int * )buf, imemord, &js, &je, &ks, &ke, &is, &ie,
+            F_UNPACK_INT ( (int * )(p+ym_curs), (int * )buf, imemord, &js, &je, &ks, &ke, &is, &ie,
                                                  &jms,&jme,&kms,&kme,&ims,&ime, &wcount ) ;
             ym_curs += wcount*typesize ;
 	  }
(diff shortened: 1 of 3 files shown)


weiwangncar:
The regression test results:

    Failed Simulations are: 
    None
    None

sayedmekhaimer:
The modification to c_code.c, period.c, and rsl_lite.c solved the compilation error with GCC-14.1, but during the execution of wrf.exe using mpirun (dm) causes an error (segmentation fault, reference to memory). The same configuration works well using GCC-13.3.

islas:
@sayedmekhaimer Did you use commit https://github.com/wrf-model/WRF/pull/2062/commits/36d3c64aa3c956c1681fae0be33fd1a334790045 or later? The initial commit in this PR did have that issue which I thought I had fixed (tested locally with gcc14)

sayedmekhaimer:
> @sayedmekhaimer Did you use commit [36d3c64](https://github.com/wrf-model/WRF/commit/36d3c64aa3c956c1681fae0be33fd1a334790045) or later? The initial commit in this PR did have that issue which I thought I had fixed (tested locally with gcc14)

I am sorry that I used previous modification, Now the model is compiled and run in parallel mode with no errors using the the commit  [36d3c64](https://github.com/wrf-model/WRF/commit/36d3c64aa3c956c1681fae0be33fd1a334790045) . Thank you so much

mgduda:
I don't recall whether WRF always uses the PR description as the merge commit message, but if so, perhaps it would be better to include a description of the code changes here rather than simply pointing to #2047? That would make the git log more self-contained.

external/RSL_LITE/c_code.c
```diff
@@ -484,12 +484,12 @@ RSL_LITE_PACK ( int * Fcomm0, char * buf , int * shw0 ,
...
           }
           if ( typesize == 8 ) {
-            F_PACK_LINT ( buf, p+yp_curs, imemord, &js, &je, &ks, &ke, &is, &ie, 
+            F_PACK_LINT ( (long *)buf, (long *)(p+yp_curs), imemord, &js, &je, &ks, &ke, &is, &ie, 
                                                 &jms,&jme,&kms,&kme,&ims,&ime, &wcount ) ;
             yp_curs += wcount*typesize ;
           }
 	  else if ( typesize == 4 ) {
-            F_PACK_INT ( buf, p+yp_curs, imemord, &js, &je, &ks, &ke, &is, &ie,
+            F_PACK_INT ( (int * )buf, (int * )(p+yp_curs), imemord, &js, &je, &ks, &ke, &is, &ie,
```
mgduda, line 492: Is there a reason for using an extra space after the `*` here, but not for the `(long *)` casts?

external/RSL_LITE/period.c
```diff
@@ -169,12 +169,12 @@ void RSL_LITE_PACK_PERIOD ( int* Fcomm0, char * buf , int * shw0 , int * typesiz
 	  MPI_Abort(MPI_COMM_WORLD, 98) ;
         }
         if ( typesize == 8 ) {
-          F_PACK_LINT ( buf, p+xp_curs, imemord, &js, &je, &ks, &ke, &is, &ie,
+          F_PACK_LINT ( (long *)buf, (long *)p+xp_curs, imemord, &js, &je, &ks, &ke, &is, &ie,
```
mgduda, line 172: Although I'm not certain, I suspect we need parentheses here, with
`(long *)p+xp_curs` becoming `(long *)(p+xp_curs)`.

mgduda (changes requested):

external/RSL_LITE/c_code.c
islas: I believe it was so that 
`(long *)` and
`(int * )` 
evaluate to the same length of characters, but it could also just that I added it in the find and replace commands.

external/RSL_LITE/period.c
islas: I think that's correct. I had that issue with the `F_[UN]PACK_INT` commands, but I don't think our tests exercise this section and evidently I forgot to update this section in those particular fixes.

mgduda (approved):

weiwangncar (approved):

weiwangncar:
@islas Maybe you could make a concise message when merging?

islas:
@weiwangncar Do you mean instead of the PR description?

PR #2069, opened 2024-06-14
Break out module_dm external subroutines into separate files

islas:
TYPE: enhancement

KEYWORDS: intel, compilation, llvm, memory

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
The Intel oneAPI compilers (and others like nvhpc) struggle with some of the larger (15k+ lines of code) files within WRF. This causes intense memory usage that is not often available to the average user not in a resource-rich environment. This often limits compilation to single threaded if even possible or to a dedicated environment with enough memory if available. If neither of those is available to a user, they will be unable to use these configurations entirely.

Solution:
This PR focuses on the `module_dm` sections of code to reduce its individual file size to manageable levels. This and its helper subroutines are instead broken out into many smaller files. 

TESTS CONDUCTED: 
Attached to this PR are plots of the respective effects of theses changes. Changes were tested with intel and gcc compilers, but only intel memory usage is shown as it exacerbates the memory usage issue.


islas:
Highlighted is the region during compilation which memory usage spikes that this PR addresses (`module_dm`) before these changes take place :
pre_module_dm

This usage is then dropped when using this PR's edits, splitting across multiple other files :
post_module_dm_split

Zooming in we can now see the external subroutines originally in `module_dm` comprise a number of smaller compilation units :
post_module_dm_split_zoom

weiwangncar:
The regression test results:

    Failed Simulations are: 
    None
    None

weiwangncar:
I tested code before and after this PR, and model produces identical results in my test. Also with 4 processors, the compile time is about 12 minutes!

mgduda (commented):
I haven't yet done any nested runs to test these changes. The only potential issue that I can see so far is related to the use of `.F90` suffixes for files that are compiled based on rules in `configure.wrf`. Specifically, `configure.wrf` has a rule for turning `.F90` files into `.o` files: https://github.com/wrf-model/WRF/blob/v4.6.0/arch/postamble#L206-L212 . For case-insensitive filesystems, the intermediate `.f90` files could be problematic, and so it may be better to use `.F` as the suffix for the new `feedback_domain_em_part1.F90`, etc. files.
Note that other `.F90` files in RSL_LITE (e.g., `f_xpose.F90`) are compiled with RSL_LITE's rules that don't generate intermediate `.f90` files.

mgduda (dismissed):

mgduda:
After cleaning the build, it looks like the following processed source files aren't being removed:
```
	external/RSL_LITE/feedback_domain_em_part1.f90
	external/RSL_LITE/feedback_domain_em_part2.f90
	external/RSL_LITE/force_domain_em_part2.f90
	external/RSL_LITE/interp_domain_em_part1.f90
	external/RSL_LITE/interp_domain_em_part2.f90
	external/RSL_LITE/interp_domain_em_part3.f90
	external/RSL_LITE/interp_domain_em_small.f90
```

islas:
I removed the `*.f90` clean addition in the external/RSL_LITE/makefile, especially as I don't know if on case insensitive systems it would remove actual source files like external/RSL_LITE/f_pack.F90. I think adding back something like `*domain_em*.f90` would suffice but look odd inside the `clean` command

mgduda (approved):

weiwangncar (approved):

PR #2070, opened 2024-06-14
Breakout deallocation calls into simpler smaller files

islas:
TYPE: enhancement

KEYWORDS: intel, compilation, llvm, memory

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
The Intel oneAPI compilers (and others like nvhpc) struggle with some of the larger (15k+ lines of code) files within WRF. This causes intense memory usage that is not often available to the average user not in a resource-rich environment. This often limits compilation to single threaded if even possible or to a dedicated environment with enough memory if available. If neither of those is available to a user, they will be unable to use these configurations entirely.

Solution:
This PR focuses on the `deallocs.inc` sections of code used in `module_domain` to reduce the include size to manageable levels. The include is instead broken out into many smaller files as external subroutines. The files are fully generated source code from the registry, with the calls to the subroutines also being generated as well. This also makes it relatively easy to change the number of files generated from a source code perspective. Build rules would need to be modified accordingly as seen in these changes.

TESTS CONDUCTED: 
Attached to this PR are plots of the respective effects of theses changes. Changes were tested with intel and gcc compilers, but only intel memory usage is shown as it exacerbates the memory usage issue.


islas:
Highlighted is the region during compilation which memory usage spikes that this PR addresses (`module_domain`) before these changes take place :
pre_module_domain

This usage is then dropped when using this PR's edits :
post_module_domain_dealloc_split

Zooming in we can now see that deallocs_ comprise a number of smaller compilation units :
post_module_domain_dealloc_split_zoom

weiwangncar:
@islas This is so cool! Thanks for working on this! Does this affect compile time in any way?

islas:
It only affects compile times as number of threads go up. For typical compilation with `-j 4` things stay more or less the same. For `-j 12` as an example there is a decent improvement, and I suspect we would see a similar trend across most compilers if you're able to use that many threads.

### Using gfortran/gcc 34/MPI-enabled with ALL PR changes (#2070, #2069, #2068)
|Command    | Category |              Before     |   After|
|---|---|---|---|
|time ./compile em_real -j 4|||
||                              real |   7m41.110s  |   7m55.655s   |
||                              user |   15m22.460s |   16m45.250s  |
||                              sys  |   0m28.922s  |   0m31.553s   |
|time ./compile_new -j 4|||
||                              real |   5m19.380s  |   5m21.952s   |
||                              user |   14m44.306s |   16m16.393s  |
||                              sys  |   0m26.598s  |   0m31.263s   |
|time ./compile em_real -j 12|||
||                              real |   7m22.949s  |   6m36.738s   |
||                              user |   20m49.918s |   19m31.420s  |
||                              sys  |   0m42.890s  |   0m39.320s   |
|time ./compile_new -j 12|||
||                              real |   4m25.744s  |   3m41.141s   |
||                              user |   20m33.427s |   22m43.411s  |
||                              sys  |   0m36.085s  |   0m41.981s   |

weiwangncar:
I tested code before and after this PR, and model produces identical results in my test. Also with 4 processors, the compile time is about 12 minutes!

weiwangncar:
The regression test results:

    Failed Simulations are: 
    None
    None

weiwangncar (approved):

tools/gen_allocs.c
```diff
@@ -589,30 +589,97 @@ gen_dealloc1 ( char * dirname )
...
   if ((fp = fopen( fname , "w" )) == NULL ) return(1) ;
   print_warning(fp,fname) ;
-  gen_dealloc2( fp , "grid%", &Domain ) ;
+
+  fprintf(
+          fp,
+          "INTERFACE\n"
+          );
+  if ( dirname == NULL ) return(1) ;
+  for ( ; idx < numFiles; idx++ )
```
mgduda, line 609: Would it be better to initialize `idx = 0` here rather than to assume that no preceding code has modified `idx` since it was initialized at declaration?

tools/gen_allocs.c
```diff
@@ -589,30 +589,97 @@ gen_dealloc1 ( char * dirname )
...
+  {
+    if ( strlen(dirname) > 0 ) { sprintf(fname,"%s/%s%d.F90",dirname,filename_prefix,idx) ; }
```
mgduda, line 611: Will using `F90` suffixes here lead to problems on systems without case-sensitive filesystems? Do the generated `*.F90` files ever get turned into `*.f90` files?

mgduda (changes requested):

tools/gen_allocs.c
islas: These would rely on implicit `.o : .F90` rules, such as the one that was added for the DA build. 

It seems like the problem, as noted in #2057, is usually from clobbering the original source files. As these are auto-generated from the registry overwriting them should not be an issue.

tools/gen_allocs.c
```diff
@@ -589,30 +589,97 @@ gen_dealloc1 ( char * dirname )
   FILE * fp ;
   char  fname[NAMELEN] ;
   char * fn = "deallocs.inc" ;
+  // Open array of deallocs_[n].inc
+  int   numFiles = 12;
+  int    idx      = 0;
+  FILE * fpSub; 
+  char * filename_prefix = "deallocs_" ;
```
mgduda, line 596: It might be nice to use consistent whitespace here between the variable type, variable name, `=` character, and trailing semicolon.

mgduda (changes requested):
Just a minor request to tidy up some whitespace; otherwise, this looks good to me.

weiwangncar (approved):

weiwangncar (approved):

mgduda:
@islas Should we add the generated `inc/deallocs_*.F` files to a `.gitignore` file? Otherwise, these show up as untracked files after building. (Note, too, that we should do the same for the `inc/allocs_*.F` files.)

islas:
I see that there is a `.gitignore` in `inc/`, would `*.F` be better, or maybe just 
```
deallocs_*.F
allocs_*.F
```
to the `inc/.gitignore`?

mgduda:
> I see that there is a `.gitignore` in `inc/`, would `*.F` be better, or maybe just
> 
> ```
> deallocs_*.F
> allocs_*.F
> ```
> 
> to the `inc/.gitignore`?

I think being as restrictive as possible would be good, so I'd favor
```
allocs_*.F
deallocs_*.F
```

mgduda (approved):

weiwangncar (approved):

PR #2086, opened 2024-08-06
CMake Prepare for alternate core selections

islas:
TYPE: enhancement

KEYWORDS: cmake, wrfplus, da

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
The current CMake build only works with the em core / ARW in mind and presents a non-intuitive selection when attempting other cores. While these cores are not yet supported in the CMake build, remove these selections sets up the configuration script to be able to present a familiar set of options.

Solution:
Remove normally unselectable options from `configure_new` after selecting non-ARW cores. Put in place stand-in values for these unselected variables.

PR #2087, opened 2024-08-06
CMake improve configure_reader.py script

islas:
TYPE: enhancement

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
In preparation for alternate core compilation, there is need for additional information from the stanzas. Some of these fields include extraneous information to cmake that must be filtered out. The previous version manually listed fields and values to sanitize, but this is brittle when accounting for all stanza types and configurations.

Solution:
Rather than hard-coding the sanitization of stanza fields to be consumed by cmake, a generalized approach is used to allow full use of stanza fields down the line.


islas:
Requires #2056, #2053, and #2086

amstokely (dismissed):
Looks good to me.

mgduda (approved):

amstokely (approved):

PR #2088, opened 2024-08-06
CMake compilation flags as categorical properties

islas:
TYPE: enhancement

KEYWORDS: cmake, compilation, flags

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
In preparation for alternative core selection, there is a broader range of flags that must be set and controlled throughout the project. This includes being able to set flags on a per-source basis to override certain flags that are normally applied wholly to all source within a CMake target. An example of this is disabling double precision on an individual file when that option is being used.

Solution:
Categorizing the flags first allows for easy application and selection of particular sets of flags for the multiple libraries and binaries created. Second, for the core library, using pseudo-inherited CMake properties between targets and source files allows for granular control of these flag categories.


islas:
Requires #2056, #2053, #2086, and #2087

weiwangncar:
The regression test results:

    Failed Simulations are: 
    None
    None

amstokely (commented):
Looks good to me.

amstokely (approved):

mgduda (approved):

amstokely (approved):

PR #2089, opened 2024-08-06
CMake WRFPLUS

islas:
TYPE: enhancement

KEYWORDS: cmake, wrfplus

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
The current CMake build does not build the WRF PLUS core configuration

Solution:
Add the necessary source files to compile WRF PLUS. WRF binary will still be created, but a symlink to `wrfplus.exe` will also be created for backward compatibility

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -922,6 +922,10 @@ if ( ${ENABLE_HYDRO} )
   add_subdirectory( hydro  )
 endif()
 
+if ( ${WRF_CORE} STREQUAL "PLUS" )
+  add_subdirectory( wrftladj )
+endif()
+
 add_subdirectory( dyn_em   )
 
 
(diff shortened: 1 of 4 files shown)


islas:

amstokely (dismissed):
I will approve these changes, but I’m confused about why the WRF binaries have `.exe` extensions. Typically, `.exe` is used for Windows executables, which could cause confusion for users. Additionally, this could create issues when building WRF in a cross-platform Linux/Windows environment. One of the main advantages of using CMake is its support for cross-platform builds, and this naming convention might undermine that benefit.

Given that this repository will soon move to minimal maintenance, it’s even more important to follow common naming conventions for clarity. Ensuring that executable names are intuitive and platform-appropriate will help reduce user confusion and minimize the need for support as resources become more limited. It will also preserve the long-term maintainability of the project, especially when building across multiple platforms.

amstokely (approved):

mgduda (approved):
Although I have no idea how to use WRFPLUS and therefore can't really test whether it works, compilation does produce a `wrfplus` executable that appears to be identical to the `wrf` executable that's produced by the same build.

PR #2090, opened 2024-08-06
CMake confcheck switch to try_* functions

islas:
TYPE: enhancement

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
The configuration checks done by the CMake build use language-specific calls between various functions. This has lead to a rather complex signature that is also limited in scope.

Solution:
Simplify the `wrf_conf_check` calls to make use of CMake's built-in `try_compile()` and `try_run()` functions, forwarding argument to those as necessary. Aside from minor adjustments to the `wrf_conf_check` calls this should appear as a drop-in replacement that provides the same results as before, i.e. no change to the system configuration options detected.


islas:

amstokely (dismissed):

amstokely (approved):

mgduda (approved):

PR #2091, opened 2024-08-06
CMake check system for underscore requirements between C and Fortran

islas:
TYPE: enhancement

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
The current solution for implicitly matching symbols between C and Fortran relies on user-modification of a stanza for their system. While stanzas can be configured to a specific architecture and options selection, this approach is limited and overly specific to user configuration when it could be programmatically determined. Likewise, if a new stanza is introduced it will likely only be tested by one user on their particular system. Thus, if this stanza could be used elsewhere but needs underscores on C symbols for Fortran usage it will not be evident why the stanza does not work for this new user. 

Solution:
Programmatically determine if underscores are needed in C compilation in the CMake build. Sanitize out any definition of underscore in the stanza so that CMake configuration checks take precedence.


islas:
Requires #2056, #2053, #2086, #2087, #2088, and #2090

amstokely (approved):

islas:
@amstokely @mgduda This is ready to be reviewed again. The plan is to get this and its dependent PRs into the upcoming release. Thanks!

amstokely (approved):

mgduda (approved):

PR #2092, opened 2024-08-06
CMake DA Core

islas:
TYPE: enhancement

KEYWORDS: cmake, da

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
The current CMake build does not support building the DA core

Solution:
Allow the selection and building of the DA core. Some minor changes to source code were made to facilitate the compilation without resorting to in situ file rewrites with `sed`. These changes fix compilation errors with `iargc` intrinsic function in gfortran but still allow the original make / cpp preprocessing to work as originally done.

A notable difference to the original DA build is that no source code is copied or duplicated into the folders. Instead in the respective folders used for shared source code between cores if a DA build is requested only select files are compiled. Otherwise the full set of files in those directories is used.

Additionally, WAVELET, CRTM, and BUFR can be selected as CMake options (not environment variables). All are defaulted to on in this iteration.


islas:
Requires #2056, #2053, #2086, #2087, #2088, #2090, and #2091

amstokely (approved):

liujake (approved):

mgduda (approved):

PR #2095, opened 2024-08-09
Initial testing framework

islas:
TYPE: enhancement

KEYWORDS: testing, regression, test framework

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
The current regression suite code is complex, requires maintenance of multiple alternate repositories, and takes involved effort to add a new test making community contribution limited at best. Likewise, the complexity of the system reduces the likelihood of independent local testing of changes, leading to a development cycle of one-off commits done to reinvoke testing to see if meaningful commits fix the issues.

Solution:
This new proposed regression suite addresses these shortcomings in a number of discrete ways:
1. Modularize the testing framework to an generalized independent repo usable by any repo seeking to set up tests that can run locally, on HPC systems, and within any CI/CD framework
2. Write WRF-specific test scripts _inside_ the WRF repo and in a manner that does not rely on specific layouts/hardware/etc. so long as WRF can compile and run on intended system (i.e. able to be run locally)
3. Write CI/CD tests in a simple and generally CI/CD framework-agnostic method where definitions of these also reside _within the WRF repo_
4. Utilize HPC resources in a safe manner to increase breadth of testing to allow testing of many more compilers and on similar hardware to the general use case of WRF

As a first pass at demonstrating this solution, this PR implements a simple set of compilation tests using GNU x86 configurations testing serial, sm, dm, and sm+dm selections. The CI/CD portion is done via GitHub workflow actions on a specific trigger event. The values and trigger methods are configurable, but this initial implementation will use the `labeled` trigger, which will initiate tests when `compile-tests` or `all-tests` is added as a label to a pull request.

TESTS CONDUCTED:
1. Testing of this github workflow was done in a separate fork also testing on Derecho. Both positive and negative tests were used to demonstrate respective output usefulness.

RELEASE NOTE:
Introduce a modularized testing framework that allows testing locally and natively on HPC systems that lives within the WRF repository

### DIFF
--- a/.ci/env/derecho.sh
+++ b/.ci/env/derecho.sh
@@ -1,0 +1,22 @@
+#!/bin/sh
+
+echo "Setting up derecho environment"
+workingDirectory=$PWD
+. /etc/profile.d/z00_modules.sh
+echo "Loading modules : $*"
+cmd="module purge"
+echo $cmd && eval "${cmd}"
+
+# We should be handed in the modules to load
+while [ $# -gt 0 ]; do 
+  cmd="module load $1"
+  echo $cmd && eval "${cmd}"
+  shift
+done
+
+#  Go back for asinine reasons of HPC config changing your directory on you
+if [ "$workingDirectory" != "$PWD" ]; then
+  echo "derecho module loading changed working directory"
+  echo "  Moving back to $workingDirectory"
+  cd $workingDirectory
+fi
--- a/.ci/env/derecho.sh
+++ b/.ci/env/derecho.sh
@@ -14,7 +14,7 @@ while [ $# -gt 0 ]; do
   shift
 done
 
-#  Go back for asinine reasons of HPC config changing your directory on you
+#  Go back to working directory if for unknown reason HPC config changing your directory on you
 if [ "$workingDirectory" != "$PWD" ]; then
   echo "derecho module loading changed working directory"
   echo "  Moving back to $workingDirectory"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,85 @@
+name: Regression Suite
+run-name : ${{ github.event_name == 'push' && 'CI' || github.event.label.name }} (${{ github.event_name }})
+
+on:
+  push:
+    branches: [ master, develop ]
+  pull_request:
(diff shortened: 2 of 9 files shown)


islas:
I'm using the approach we're using in MPAS to setup testing with a very limited minimal  setup (simple compilation tests) at first to get something started.

The idea would be to then gradually translate the current tests to a usable format by this framework.

weiwangncar:
The regression test results:

    Failed Simulations are: 
    None
    None

.ci/env/derecho.sh
```diff
@@ -0,0 +1,22 @@
...
+echo $cmd && eval "${cmd}"
+
+# We should be handed in the modules to load
+while [ $# -gt 0 ]; do 
+  cmd="module load $1"
+  echo $cmd && eval "${cmd}"
+  shift
+done
+
+#  Go back for asinine reasons of HPC config changing your directory on you
```
mgduda, line 17: Perhaps there's a more diplomatic way of stating this. Regardless, is it always the case that the working directory is changed (in which case, we can unconditionally `cd $workingDirectory`)?

islas: Right, I will change the wording. I left it as a conditional to better detect and output a message if this does occur. It doesn't happen with `qsub` commands with HPC directives or interactive sessions.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,97 @@
...
+#   - key : value
+#     key : value
+#     tests :
+#       - value
+#       - value
+#   - < next test >
+# https://stackoverflow.com/a/68940067
+jobs:
+  buildtests:
+    if : ${{ github.event.label.name == 'compile-tests' || github.event.label.name == 'all-tests' || github.event_name == 'push' }}
```
mgduda, line 33: If not in the PR description, somewhere in the PR discussion it would be good to explain how to trigger tests through GitHub labels.

islas: I've elaborated on the PR solution description, as it is worth being in the history even though the majority of the interactions are solely within the github web view / not the repo.

mgduda: Thanks!

mgduda (approved):

PR #2097, opened 2024-08-21
Bug fix for missed reflectivity calculation in WSM6 scheme

### LINKED ISSUE #2096: WRF 4.6.0 does not generate refl_10cm data in WSM6 microphysics
In version 4.6.0 of WRF, some modifications were made to the WSM6 microphysics. Since then, it is no longer generating data for the refl_10cm variable using the do_radar_ref=1 flag in namelist.input.

To get around this problem, I was forced to reinstall the previous version 4.5.2 so that the refl_10cm data would work again, as it always did.

I think it is necessary to analyze the phys/module_mp_wsm6.F code to see what is going wrong with the modifications that were made in this new version 4.6.0 to solve this problem.

weiwangncar:
TYPE: bug fix

KEYWORDS: radar reflectivity, WSM6

SOURCE: reported by Gabriel Cassol of Sigma Meteorologia, Brazil, internal

DESCRIPTION OF CHANGES:
Problem:
When moving WSM6 to MMM shared physics directory, the calculation of radar reflectivity was left out.

Solution:
The calculation of radar reflectivity is added back in the WSM6 scheme.

Fixes #2096

LIST OF MODIFIED FILES: 
M      phys/module_mp_wsm6.F

TESTS CONDUCTED: 
1. The reflectivity is computed now when namelist do_radar_ref is set to 1.

RELEASE NOTE: Fixed the bug of missing radar reflectivity in WSM6 scheme when do_radar_ref is set to 1.

### DIFF
--- a/phys/module_sf_bem.F
+++ b/phys/module_sf_bem.F
@@ -889,7 +889,7 @@ use module_wrf_error
        uroof=(uout(n+1)**2+vout(n+1)**2)**0.5
       deltat=tpv-tair(n+1)
     hf=2.5*(40./100.*uroof)**(0.5)
-    hc=9.842*abs(deltat)**(1./3.)/(7.283-abs(cos(tiltangle)))
+    hc=9.482*abs(deltat)**(1./3.)/(7.238-abs(cos(tiltangle)))
     hup=sqrt(hc**2.+(hf)**2.)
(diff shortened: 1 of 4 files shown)


weiwangncar:
Not sure what has happened, but this is not properly done.

PR #2108, opened 2024-09-12
CMake Fix split command flags to be correctly populated

islas:
TYPE: bug fix

KEYWORDS: cmake, mpi, compilation

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
The `arch/configure_reader.py` does the job of parsing, organizing, and sanitizing input from configuration stanzas into a CMake toolchain file which can then be used to inform the build about which compilers, flags, and options to use. 

Occasionally, stanzas fields inject flags into a compiler or other command field (like `DM_FC`) so that the actual command is a command + flags. Part of the `arch/configure_reader.py` organization is breaking these up into separable sections automatically. With the example of `DM_FC = mpif90 -f90=gfortran` (`$(SFC)` already expanded) this _should_ be broken into `DM_FC = mpif90` and `DM_FC_FLAGS = -f90=gfortran`. Currently, the `*_FLAGS` field when split out for certain keys in a stanza is not populated due to using the wrong index from the Python `str.partition()` call.

Secondly, when these fields are actually provided to CMake compilation breaks for MPI specifically. Since the MPI "compilers" are wrappers, they are then interrogated for the underlying flags and options meaning further adding the flags back into compilation results in things like `gfortran <all the other flags> -f90=gfortran`. This is incorrect, and instead the flags should be provided to the MPI flags used during wrapper interrogation on a per-language basis. Furthermore, for certain MPI implementations supplying any flags renders the query command (e.g. `-show`, `-showme`, or `-compileinfo`) useless. For instance, OpenMPI `mpif90 -f90=gfortran -show` only outputs `gfortran -f90=gfortran` which is also wrong. 

Solution:
1. The Python call to split the fields should take the actual values
2. MPI flags per language should be supplied to the interrogation flags `MPI_<LANG>_COMPILER_FLAGS` if needed
3. The `FindMPI` module already feeds in the underlying compiler specification for wrappers that support it so flags like `-f90=$(SFC)` should be filtered out from `DM_*_FLAGS` before being written to the `wrf_config.cmake` toolchain file

TESTS CONDUCTED: 
1. Flags in specific command fields are now split and appearing in the `wrf_config.cmake` file
2. MPI compilation works with compiler specification filtered out but correct underlying compiler still selected even when multiple compilers are in the same environment (note: this was the original behavior but now is deliberate)


amstokely (commented):
These changes look good to me.

amstokely (dismissed):

amstokely (dismissed):

mgduda (approved):

amstokely (approved):

mgduda (approved):

PR #2110, opened 2024-09-23
Update how fogdes identifies land use type

WQadam:
Fix fog des (grav_settling) for LCZ dataset 

TYPE: bug fix

KEYWORDS: grav_setting, LCZ

SOURCE: Adam Dury (WeatherQuest)

DESCRIPTION OF CHANGES:
Problem:
WRF crashes when grav_settling is greater than 0 and wudapt_lcz = 1 due to the historical code using hard coded values for land categories to detect which land use type is being used. This isn't the case now so needed to update to use a variable with actual land use category type so the grav_settling code would work when wudapt_lcz was turned on and the nlcat wasnt the same for each land use type.

Solution:
Changed the if statements to use mminlu rather than nlcat so that the grav_settling option knows what land use dataset is being used. Corrected the error message to go into rsl.error files rather than rsl.out files

LIST OF MODIFIED FILES: 
phys/module_sf_fogdes.f
phys/module_surface_driver.f

TESTS CONDUCTED: 
1. Tested with wudapt_lcz on and off with all option of grav_settling.

RELEASE NOTE: Updated grav_settling code to better recognize the land use type so it doesn't crash. Also update the error message if it does crash to go into the rsl.error files rather than rsl.out files.

### DIFF
--- a/phys/module_sf_fogdes.F
+++ b/phys/module_sf_fogdes.F
@@ -105,6 +105,7 @@ CONTAINS
 !JOE-end
 
 ! Local variables
+   CHARACTER*512 :: message
    INTEGER :: i,j
    REAL    :: lad, spcfct, vegh, ftmp1, ftmp2, dp_fog, lwc
    CHARACTER (LEN=25) :: land_use_type, lu_fogdes
@@ -116,7 +117,8 @@ CONTAINS
    ELSEIF (mminlu .eq. 'USGS') THEN ! includes lake category
      land_use_type = 'USGS'
    ELSE
-     PRINT *, 'Unknown landuse category (sf_fogdes.F): num_land_cat=',nlcat
+     write ( message, * ) "Unknown land category for grav_settling (sf_fogdes.f) Landuse type = ",mminlu," and number of land categories= ",nlcat
+     CALL wrf_error_fatal ( message ) 
      STOP
    END IF
 
--- a/phys/module_sf_fogdes.F
+++ b/phys/module_sf_fogdes.F
@@ -117,7 +117,7 @@ CONTAINS
    ELSEIF (mminlu .eq. 'USGS') THEN ! includes lake category
      land_use_type = 'USGS'
    ELSE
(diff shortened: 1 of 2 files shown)


weiwangncar:
The regression test results:

    Number of Builds       : 60           56
    Number of Simulations  : 158           147        0
    Number of Comparisons  : 95           83        0

    Failed Simulations are: 
    None
    None

weiwangncar:
@dudhia Can you review this PR?

phys/module_sf_fogdes.F
```diff
@@ -111,9 +111,9 @@ SUBROUTINE sf_fogdes(&
...
 !------------------------------------------------------------------
 
-   IF     ((nlcat .eq. 20).or.(nlcat .eq. 21)) THEN ! includes lake category
+   IF     (mminlu .eq. 'MODIFIED_IGBP_MODIS_NOAH') THEN ! includes lake category
      land_use_type = 'MODIS'
-   ELSEIF ((nlcat .eq. 24).or.(nlcat .eq. 28)) THEN ! includes lake category
+   ELSEIF (mminlu .eq. 'USGS') THEN ! includes lake category
      land_use_type = 'USGS'
    ELSE
      PRINT *, 'Unknown landuse category (sf_fogdes.F): num_land_cat=',nlcat
```
dudhia, line 119: modify the print to output mminlu

weiwangncar: @WQadam Can you modify the print following your line of change to print mminlu, instead of nlcat? 
@dudhia I assume that is what you suggested?

WQadam: Sorry i have been busy the last couple of days. 
I printed it in my testing but forgot to put it in the github code. I have now made the error go into the rsl.error files instead of the error going into a random rsl.out file. Hopefully it looks okay and will never really be triggered.

dudhia (approved):

weiwangncar (approved):

PR #2116, opened 2024-10-01
ppc64le build

jstrodtb:
Updated build files for modern ppc64le, including undefining the "vector" macro and supporting ibm-clang. Previously, the ppc64le build would fail to compile.

TYPE: bug fix

KEYWORDS: ppc64le, gcc, build

SOURCE: Joshua Strodtbeck (Tachos High Performance Computing)

DESCRIPTION OF CHANGES:
Problem:
The old IBM XL compiler has been superseded by ibm-clang. Moreover, both gcc and ibm-clang define a macro called "vector" which causes the preprocessor to fail. Note that this fix depends on islas/cmake-fixSplitCommandFlags, which must be merged first, or the preprocessor fix won't work.

Solution:
Added ibm-clang and ppc64le gcc to 

N/A

arch/configure.defaults

TESTS CONDUCTED: 
Compiled and ran conus2.5km on my Power9 box.

RELEASE NOTE: The latest IBM compiler, ibm-clang, is now supported.


arch/configure.defaults
```diff
@@ -771,7 +771,54 @@ CC_TOOLS        =      $(SCC)
 NETCDFPAR_BUILD	=      CONFIGURE_NETCDFPAR_BUILD
 
 ###########################################################
-#ARCH    Linux x86_64 ppc64le, gfortran compiler with gcc  #serial smpar dmpar dm+sm
+#ARCH    Linux ppc64le, gfortran compiler with gcc  #serial smpar dmpar dm+sm
```
islas, line 774: There is a comment in `configure.defaults` ( `#insert new stanza here`) showing where new stanzas should be made. While this may not be entirely new and more of a split up of the x86_64 and ppc64le architectures, certain tests assume the position of stanzas with an emphasis on the specific location of x86_64.

This new ppc64le stanza should be moved lower to the appropriate location.

arch/configure.defaults
```diff
@@ -1044,6 +1091,54 @@ RLFLAGS		=	-c
 CC_TOOLS        =      $(SCC) 
 NETCDFPAR_BUILD	=      CONFIGURE_NETCDFPAR_BUILD
 
+###########################################################
```
islas, line 1094: This new stanza should also be moved lower to the appropriate location

islas (changes requested):

PR #2130, opened 2024-11-13
CMake sanitize leading define flag

islas:
TYPE: bug fix

KEYWORDS: compilation, cmake

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
With the addition of #2088, `ARCH_LOCAL` from the stanza is now being fed into compilation with potential `-D` defines. On versions of CMake <3.26 leading `-D` on defines is not removed from certain function calls like `add_compile_definitions()`. This will pass the configuration stage but will fail to compile when using the defined minimum CMake version of the project.  

Solution:
To simplify the logic, all defines fed into the `wrf_config.cmake` file for the configuration step will be sanitized of leading `-D`. This follows the original design intent where stanza sanitization happens before being fed into CMake, thus allowing the CMake code to focus on configuration of options rather than translation of stanza into usable values.

TESTS CONDUCTED: 
1. Tested configuration and compilation on CMake version v3.20.6

RELEASE NOTE: 
Remove leading -D on defines during stanza reading  to allow older versions of CMake to configure properly.


weiwangncar:
The regression test results:

    Failed Simulations are: 
    None
    None

amstokely (approved):

mgduda (approved):

PR #2135, opened 2024-12-02
Fix typo in cmake find netCDF causing failure on blank responses

islas:
TYPE: bug fix

KEYWORDS: netcdf, cmake

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
Certain versions of netCDF with particular configurations return a blank when queried with `nc-config`. To prevent a null argument expansion in CMake we should check for variable emptiness and if empty assume the queried feature is not present. There is a typo copied from the FindnetCDF-Fortran.cmake module causing this check to never get exercised leading to failed configuration due to syntax errors.

Solution:
Correct the typo to reference the netCDF variables rather than the netCDF-Fortran ones.

TESTS CONDUCTED: 
1. Tested against netCDF 4.7.2 without szlib (`--has-szlib ->    `)

RELEASE NOTE: Bug fix in CMake FindnetCDF.cmake for empty --has-* nc-config fields


kkeene44 (approved):

weiwangncar:
The regression test results:

    Failed Simulations are: 
    None
    None

PR #2142, opened 2024-12-12
Use proxy workflow and commit statuses to record testing results

islas:
TYPE: enhancement

KEYWORDS: testing, devops, github, workflow

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
The CI/CD testing framework using github actions is set to trigger on PR label events. However, labels are not just used for testing and will trigger the workflow. While the workflow does have checks in place to skip any labels that aren't meant to trigger testing, this skipped workflow status will override any previous actual test status. This can be confusing if two labels are applied at the same time, one being a test label, and only one status appears showing a skipped workflow. Unique naming of workflow runs does not mitigate this problem as the posted status is tied to the workflow internal id and not the run name.

Solution:
Convert the main workflow that hosts the test sets into a dispatch workflow, meaning it must manually be triggered. This has the intended effect of decoupling the workflow id from any PR and will not normally show up as a status at the bottom of PRs solving the issue of independent labels overriding each other. To solve the workflow no longer appearing within PR statuses, the github REST API is used to create a commit status pointing to its respective workflow run via `target_url` along with the current state of the job. 

As the main workflow must manually be triggered, a new entry point proxy workflow is used to filter test labels and request a test run if needed. This paradigm allows events to be triggered within a PR context, simplifying gathering the data necessary to run the correct PR branch. Furthermore, the entry point will still suffer the initial problem of status override on multiple labels, but this should be acceptable as actual test labels will create their own commit statuses once queued. An added benefit is now there would be a clear path to manually running tests through github actions on any branch, regardless of a PR being opened, with the dispatch workflow. 

TESTS CONDUCTED: 
1. Testing was done in independent fork to demonstrate initial issue and feasibility of this solution.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -118,20 +118,21 @@ jobs:
       statuses: write
     name : Test ${{ matrix.testSet.name }} on ${{ matrix.testSet.host }}
 
-  # In the event that 'all-tests' is used, this final job will be the one to remove
-  # the label from the PR
-  removeAllLabel :
-    if : ${{ !cancelled() && github.event.label.name == 'all-tests' }}
-    name : Remove 'all-tests' label
-    runs-on: ubuntu-latest
-    needs : [ buildtests ] # Put tests here to make this wait for the tests to complete
-    steps: 
-      - name : Remove '${{ github.event.label.name }}' label
-        env:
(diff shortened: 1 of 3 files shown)


islas:
Closing as `workflow_dispatch` works within the context of a repository but is not viable for fork-pull request development

PR #2145, opened 2024-12-18
CMake PnetCDF Find Package 

islas:
TYPE: enhancement

KEYWORDS: cmake, pnetcdf

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
The CMake build does not support PnetCDF detection and thus the `external/io_pnetcdf/` code cannot be built.

Solution:
Write a FindPnetCDF.cmake package that provides the appropriate imported target to link against and does precursory library matching if there is a dependency on netCDF C. Update the cmake code in `external/io_pnetcdf/` to build, with minor adjustments that make the CMake and Make builds compatible.

The package uses the same rules of the netCDF packages, utilizing `pnetcdf-config` in the `PATH` as the source of information, or otherwise found via standard `find_package()` mechanics, e.g. `PnetCDF_ROOT`.

TESTS CONDUCTED: 
1. Tested CMake and Make builds with PnetCDF in environment on local machine and Derecho.

RELEASE NOTE: CMake PnetCDF Find Package


amstokely (approved):

dudhia (dismissed):

external/io_pnetcdf/ext_pnc_put_dom_ti.code
```diff
@@ -38,7 +38,7 @@
   use ext_pnc_support_routines
   implicit none
   include 'wrf_status_codes.h'
-#  include "pnetcdf.inc"
+  include "pnetcdf.inc"
```
mgduda: Here and elsewhere, it might be nice to use single quotes for consistency with other Fortran `include` statements.

mgduda (changes requested):

external/io_pnetcdf/field_routines.F90
```diff
@@ -38,7 +38,7 @@ subroutine ext_pnc_RealFieldIO(Coll,IO,NCID,VarID,VStart,VCount,Data,Status)
   use ext_pnc_support_routines
   implicit none
   include 'wrf_status_codes.h'
-#  include "pnetcdf.inc"
+#  include 'pnetcdf.inc'
```
mgduda: Here and in several other places, it looks like these lines are still pre-processor directives.

mgduda (changes requested):

external/io_pnetcdf/field_routines.F90
islas: Not all need to be changed away from preprocessor directives. `wrf_io.F90` and the `*.code` files needed to since they are `cpp` preprocessed and then `m4` preprocessed and that syntax didn't quite fit with the compilation method of either make or cmake while keeping the C-style preprocessor directives.

That said, I may need to revert all or part of 26bf71a as single quotes are not allowed in the C include directive. Or everything just changed to Fortran include directives regardless. This however starts to get into modifying code _"somewhat"_ unrelated to the changes presented in this PR.

mgduda:
My apologies -- I think I was confused as to what was going on after the second commit. I originally meant to suggest switching to single quotes only for the preprocessing directives that had been changed to Fortran `include` statements.

islas:
@mgduda I believe this should be ready to review again

mgduda (approved):

weiwangncar (approved):

PR #2155, opened 2025-01-17
instruct GitHub to create empty, instead of incomplete, release tarballs (via .gitattributes)

slayoo:
PURPOSE: instruct GitHub to create empty, instead of incomplete, release tarballs (via .gitattributes)

TYPE: enhancement

KEYWORDS: release, gitmodules

SOURCE: @slayoo

DESCRIPTION OF CHANGES:
Problem:
image

Solution:
Make the auto-generated files empty (seems that disabling auto-generation entirely is not supported)

ISSUE: none ?

LIST OF MODIFIED FILES: `.gitattributes` added

TESTS CONDUCTED: 
Tested to work in another project: https://github.com/gnudatalanguage/gdl/blob/master/.gitattributes

RELEASE NOTE: minor change not affecting the proper release files, hence not needed to be mentioned in release notes


.gitattributes
```diff
@@ -0,0 +1,8 @@
+# Instructing GitHub to exlude all files from auto-generated release tar/zip-balls.
+# These auto-generated release files are incomplete - lacking git submodule contents.
+# It is not possible to disable these auto-generated release files, hence at least
+# forcing them to be empty to avoid disseminating incomplete tarballs.
+# Archives with git submodule contents included are generated by CI workflow and are
```
islas, line 5: Could you update the comments to be generalized to only the removal of incomplete archives, as we do not generate release artifacts via CI workflows at this moment.

islas (changes requested):
This looks like a good semi-solution to GitHub's lack of archive control on releases! I think ideally we would want to just disable it, but this should make the error obvious to those that download the wrong archive. We should still keep the bold warning in addition to this.

.gitattributes
slayoo: changed in 5f97b1684c3dadb924e445a46fef77d7b97e1ce6

slayoo:
FWIW, here's a link to github workflow step we use in GDL to upload submodule-aware tarballs to GitHub release pages automatically (when creating a release via GitHub website):
https://github.com/gnudatalanguage/gdl/blob/144aad86a0d5846543c99b77acc193ab02e5adac/.github/workflows/main.yml#L60-L110

weiwangncar:
The results from regression tests:

    Failed Simulations are: 
    None
    None

islas (approved):

weiwangncar:
@islas It seems useful and can avoid confusion. But we may want to do this in 4.8?

PR #2157, opened 2025-01-17
  Determine MPI Data Types in col_on_comm() & dst_on_comm() to prevent displacements overflow. (Fix for #2156)

### LINKED ISSUE #2156: MPI_Gatherv/MPI_Scatterv displacements overflow in frame/collect_on_comm.c
**Describe the bug**

The functions `col_on_comm()` & `dst_on_comm()` in `frame/collect_on_comm.c` use `MPI_CHAR`  as the underlying datatype in `MPI_{Gather,Scattter}v` operations.  This means the required displacements, `displace[]`, are in terms of bytes.  For large problems, and large local communicators, this can cause overflow in the displacement offsets, which manifests in MPI communication failure.  Typically with a very obtuse error message. 

This seems to occur more frequently with large local communicators, typical of high-core-count nodes.

**To Reproduce**

We have boiled this down to a 6-rank example that is available on NSF NCAR/Derecho at `/glade/work/negins/consulting/RC-26919/high-res`, with a PR to be submitted with a proposed fix.

**Expected behavior**

1. These routines should be modified to use proper MPI data types for a given `*typesize` so the displacements are smaller (elements instead of bytes),
2. These routines should perform error checking for overflow.

**Additional context**

Related to #1333 
We think this is also the underlying issue with https://forum.mmm.ucar.edu/threads/cxil_map-write-error-with-real-exe.19321/

benkirk:
Determine MPI Data Types in col_on_comm() & dst_on_comm() to prevent displacements overflow.

TYPE: bug fix

KEYWORDS: prevent displacements overflow in MPI_Gatherv() and MPI_Scatterv() operations

SOURCE: Benjamin Kirk & Negin Sobhani (NSF NCAR / CISL)

DESCRIPTION OF CHANGES:
Problem:
The MPI_Gatherv() and MPI_Scatterv() operations require integer displacements into the communications buffers. Historically everything is passed as an MPI_CHAR, causing these displacements to be larger than otherwise necessary. For large domain sizes this can cause the displace[] offsets to exceed the maximum int, wrapping to negative values.

Solution:
This change introduces additional error checking and then uses the function MPI_Type_match_size() (available since MPI-2.0) to determine a suitable MPI_Datatype given the input *typesize.  The result then is that the displace[] offsets are in terms of data type extents, rather than bytes, and less likely to overflow.

ISSUE: Fixes #2156 

LIST OF MODIFIED FILES: 
M       frame/collect_on_comm.c

TESTS CONDUCTED: 
Failed cases run now.

RELEASE NOTE: 
Determine MPI Data Types in col_on_comm() & dst_on_comm() to prevent displacements overflow.

### DIFF
--- a/frame/collect_on_comm.c
+++ b/frame/collect_on_comm.c
@@ -69,6 +69,7 @@ col_on_comm ( int * Fcomm, int * typesize ,
   int root_task ;
   MPI_Comm *comm, dummy_comm ;
   int ierr ;
+  int dtype = -1;
 
   comm = &dummy_comm ;
   *comm = MPI_Comm_f2c( *Fcomm ) ;
@@ -90,6 +91,18 @@ col_on_comm ( int * Fcomm, int * typesize ,
     for ( p = 1 , displace[0] = 0 , noutbuf_loc = recvcounts[0] ; p < ntasks ; p++ ) {
       displace[p] = displace[p-1]+recvcounts[p-1] ;
       noutbuf_loc = noutbuf_loc + recvcounts[p] ;
+
+      /* check for overflow: displace is the partial sum of sendcounts, which can overflow for large problems. */
+      /* it is sufficient to check displace[p] since it is greater than sendcounts[p] */
+      if ( (displace[p] < 0) || (noutbuf_loc < 0) )
+        {
+#ifndef MS_SUA
+          fprintf(stderr,"%s %d buffer offset overflow!!\n",__FILE__,__LINE__) ;
+          fprintf(stderr," ---> p = %d,\n ---> displace[%d] = %d,\n ---> noutbuf_loc = %d,\n ---> typesize = %d\n",
+                  p, p, displace[p], noutbuf_loc, *typesize);
+#endif
+          MPI_Abort(MPI_COMM_WORLD,1) ;
+        }
     }
 
     if ( noutbuf_loc > * noutbuf )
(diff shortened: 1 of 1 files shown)


benkirk:
Just for awareness, I can't see the output of the failed `WRF-BUILD-2690`; I get a timeout accessing 
https://ncar_jenkins.scalacomputing.com/job/WRF-Feature-Regression-Test/2690/console

dudhia:
Common problem - it will be provided by message later

weiwangncar:
The regression test results:

    Failed Simulations are: 
    None
    None

weiwangncar:
@benkirk Thanks for the fix! I tested it for a few cases we could run before and it is working now.

benkirk:
Thanks for the success report @weiwangncar, happy to help!

dudhia:
I will tag @mgduda and @islas to review this.

islas (approved):
@benkirk Thanks for the fix! This has been a heck of a problem to hunt down.

If my understanding is correct the overflow can still manifest, just with a domain approximately 4x in size than what triggered it before, right? Though the error will now be caught and aborted at the appropriate location.

benkirk:
That's right, the `displace[]` buffer is now in terms of array elements, not bytes. So for `REAL*4` or integer arrays the overflows can occur still occur, but at ~4x larger problem sizes than before. And in these cases the overflow error will be caught in the routine so at least the source of the issue is identified.

(There's some potential fallbacks even in that case I can think of but haven't implemented.  Those fallbacks would require some conditions on the buffers to be sent, like evenly divisible by some integer factor. Such tricks are necessary when trying to send buffers >2GB with MPI.)

weiwangncar (approved):

PR #2161, opened 2025-01-27
CMake Better nf-config libdir detection

islas:
TYPE: enhancement

KEYWORDS: cmake, netcdf, netcdf-fortran, nf-config

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
As nf-config does not provide an equivalent --libdir output we must derive the info ourselves. --flibs provided the info previously, but it may be more practical to use the --prefix output and search suspected suffixes instead as --flibs contains other libraries' link info. 

Solution:
Use output of --prefix from nf-config as the primary source of library dir. To handle the cases where --prefix is empty, we still fall back to --flibs and if that proves insufficient then the as a last resort one directory above the include directory can be used as a substitute for the prefix.

TESTS CONDUCTED: 
1. Tested with netCDF-Fortran v4.6.2, modifying nf-config to emulate fallback if clauses

RELEASE NOTE: Better nf-config libdir detection in CMake build

### DIFF
--- a/cmake/modules/FindnetCDF-Fortran.cmake
+++ b/cmake/modules/FindnetCDF-Fortran.cmake
@@ -81,9 +81,19 @@ else()
   endforeach()
 
 
-  # A bug in previous netcdf-fortran cmake builds, extract from flibs
-  string( REGEX MATCH "^-L([^ ]*)" netCDF-Fortran_LIBRARY_LINK_LOCATION ${netCDF-Fortran_FLIBS} )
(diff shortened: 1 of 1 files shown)


amstokely (approved):

amstokely (approved):

abishekg7 (approved):
Changes look good and check that it configures + builds on Eris.

dudhia (approved):

mgduda (approved):

PR #2171, opened 2025-02-26
Fix the bug: TUV photolysis is inactive if simulation starts on 1st Jan

SeregaOsipov:
TYPE: bug fix

KEYWORDS: TUV, distance to the sun, 1 January, Julian day

SOURCE: Sergey Osipov (KAUST)

DESCRIPTION OF CHANGES:
Problem:
TUV and FTUV fail to initialize the distance to the Sun properly if the simulation starts on 1 Jan. The multiplication factor remains uninitialized at 0, which zeros out photolysis rate calculations (see line 745 @ module_phot_tuv.F. On the next day, the distance is recalculated normally and photolysis rates are calculated properly.

Solution:
Given that the calculation of sun distance is trivial, the if check was disabled

LIST OF MODIFIED FILES: 
chem/module_phot_tuv.F
chem/module_ftuv_driver.F

TESTS CONDUCTED: 
 The Jenkins tests are all passing.

RELEASE NOTE: This PR fixed a bug where TUV and FTUV fail to initialize the distance to the Sun properly if the simulation starts on 1 Jan.

### DIFF
--- a/chem/module_ftuv_driver.F
+++ b/chem/module_ftuv_driver.F
@@ -1308,10 +1308,11 @@ include 'netcdf.inc'
 !-----------------------------------------------------------------------------
 !     set solar distance factor
 !-----------------------------------------------------------------------------
-         if( curjulday /= julday ) then
-            curjulday = julday
-            call sundis( curjulday, esfact )
-         endif
+         ! osipov fix the bug. Always calc the distance. Testing for 0 results in True on 1 Jan, keeps the sfact=0 and shutsdown the photolysis for entire day
+         !if( curjulday /= julday ) then
+         curjulday = julday
+         call sundis( curjulday, esfact )
+         !endif
 
       end subroutine ftuv_timestep_init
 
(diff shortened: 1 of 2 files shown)


weiwangncar:
@jordanschnell Just wondering if you could help review this PR? Thanks!

jordanschnell:
Hi Sergey, thanks for this fix. Instead of commenting out the check such that it is called each time, what about including a check for julday = 0?

SeregaOsipov:
Hi Jordan, good to hear from you.

_Julday_ is zero if the simulation start date is 1st Jan. At least, this is what I saw in the DDT debugger on our supercomputer. 
The value is set in line 346 in chem_driver.F: ijulian=ifix(grid%julian) and then passed to the photolysis_driver.F

An alternative is to initialize the _curjulday_ with -1 (or -999) instead of 0, which is used as a cache for the last calculation.

Overall, the _sundist_ calculation is very trivial and only involves trigonometric functions, meaning that the computatinal overhead is negligeble. IMHO, this is a good trade-off to avoid the hard-to-predict bugs like the 1st January case.

Best, Sergey

jordanschnell (approved):
OK with me. Approved.

jordanschnell (approved):

weiwangncar (approved):

PR #2174, opened 2025-03-01
Upgrade and bug fixes in adios2 (io form=14)

### LINKED ISSUE #2167: ADIOS2 (io_form=14) errors when enabling compression and having more than one (nesting) domains
**Describe the bug**
An error 'Operator Compressor defined twice' occurs when using ADIOS2 to write wrfout* and wrfrs* for more than one domains. The files are all written, but only the wrfout* of the first domain are compressed.

**To Reproduce**
Have the below set in namelist.input
```
&time_control
 io_form_history                     = 14,
 io_form_restart                     = 14,
/

&namelist_adios2
 adios2_compression_enable = .true.,
/

&domains
/
```

**Expected behavior**
The outputted wrfout* and wrfrst* are smaller than what obtained when using 'adios2_compression_enable = .false.', and with no errors.

**Screenshots**
Image
Image
Image

**Attachments**
[namelist_input.txt](https://github.com/user-attachments/files/18782956/namelist_input.txt)

**Additional notes**
The net file size becomes larger when using more processors (MPI tasks) to run exactly the same case, i.e., namelist.input, wrfinput*, wrfbdy*. Just FYI here, I will search and ask about this behavior in the ADIOS2 github.

SKanoksi:
Upgrade and bug fixes in ADIOS2 (io_form=14)

TYPE: new feature, bug fixes, enhancement

KEYWORDS: ADIOS2, compression, blosc compressor, blosc clevel, operator defined twice, IO declare twice, lib64

SOURCE: Somrath Kanoksirirath (NCR, ThaiSC)

DESCRIPTION OF CHANGES:
Problem:
1. 'Compressor operator defined twice' exception is thrown when having more than one domains, causing no compression in subsequent files. (#2167)
2. Segmentation fault occurs when running wrf.exe with 'adios2_compressoin_enable = .true.' when WRF is compiled using Intel oneAPI (ifx) or Cray Fortan compiler, for example. Note: this error is not found when using ifort.
3. New/future blosc compressor (input for 'adios2_blosc_compressor') cannot be used since the accepted list is hard coded.
4. The compression level of the blosc compressor cannot be adjusted.
5. 'IO wrfbdy_d01 declared twice' exception is thrown when running wrf.exe with 'io_form_boundary = 14', causing the run to stop.
6. Error/Warning messages do not state which adios2 function is causing the problem
7. '-ladios2_fortran_mpi' and/or '-ladios2_fortran' library are not found when building WRF if ADIOS2 installation has both /lib and /lib64 directories.

Solution: 
(in order)
1. Use 'adios2_inquire_operator' to check and reuse the existing 'Compressor' operator. Because ADIOS2 operator needs to have an unique name, so declaring a new one using exactly the same name will throw an exception.
2. Remove 'if (DH%compress_operator%name .ne. 'Compressor')' condition which accessing the name before the operator is defined in later lines. (Note: the fix mentioned in #2167 was not taken account of this issue.)
3. Remove the hard-coded check. When the blosc compressor is not supported or not available, let the exception from ADIOS2 propagates up and return with a meaningful error message instead.
4. Add 'adios2_blosc_clevel' to &namelist_adios2 in Register/registry.io_boilerplate. Use 'adios2_set_operation_parameter' to set blosc compression level, in external/io_adios2/wrf_io.F90. Update doc/README.adios2.
5. Find and use a vacant IO name by iterating 'filename_<n>' and provide a heads-up to user. (Note: ADIOS2 IO needs to have an unique name under the same ADIOS2 program-wide interface.)
6. Edit error/warning messages in external/io_adios2/wrf_io.F90 to make them more meaningful.
7. Edit arch/Config.pl to make it prefer /lib64 (where the shared libraries are resided if the directory exists) over /lib (where the ADIOS2 Python API, /pythonXX/site-packages/adios2, is resided when both /lib64 and /lib exist).

ISSUE: Fixes #2167 and more

LIST OF MODIFIED FILES:
M Registry/registry.io_boilerplate
M arch/Config.pl
M doc/README.adios2
M external/io_adios2/wrf_io.F90

TESTS CONDUCTED:
1. Similar tests as in #2167 were conducted using Intel Classic (ifort), Intel oneAPI (ifx) and Cray Fortran (ftn) compilers (using DM+SM option) on a HPE Cray EX system. The programs run and complete successfully. The compressed outputs are reduced in size by around 60%. The sampling results are the same.
Screenshot from 2025-03-01 19-00-53
2. Additional tests, where ADIOS2 is also used for wrfbdy and wrfinput to run real.exe and then wrf.exe, were conducted. The programs run and complete successfully but with some warnings/exceptions from ADIOS2. The RSL logs, with step checking in external/io_adios2/wrf_io.F90 uncomment, are attached. This behavior is also mentioned in the new README.adios2. [ADIOS2_bdy14_log.zip](https://github.com/user-attachments/files/19042941/ADIOS2_bdy14_log.zip)
4. The restart run works properly.

RELEASE NOTE:
Upgrade and bug fixes in ADIOS2 (io_form=14)


SKanoksi:
Don't know why the WRF regression test failed, even when only the doc/README.adios2 is changed. Will try another one step by step.

PR #2175, opened 2025-03-04
Upgrade and bug fixes in ADIOS2 (IO=14)

### LINKED ISSUE #2167: ADIOS2 (io_form=14) errors when enabling compression and having more than one (nesting) domains

SKanoksi:
(Previously #2174 ) Upgrade and bug fixes in ADIOS2 (io_form=14)

TYPE: bug fix, enhancement

KEYWORDS: ADIOS2, compression, blosc compressor, blosc clevel, operator defined twice, IO declare twice, lib64

SOURCE: Somrath Kanoksirirath (NCR, ThaiSC)

DESCRIPTION OF CHANGES:
Problem:
1. 'Compressor operator defined twice' exception thrown when having more than one domains, causing no compression in subsequent files. (#2167)
2. Segmentation fault when running wrf.exe with 'adios2_compressoin_enable = .true.' when WRF is compiled using Intel oneAPI (ifx) or Cray Fortan compiler, for example. Note: this error is not found when using ifort.
3. New/future blosc compressor (for 'adios2_blosc_compressor') cannot be used since the accepted list is hard coded.
4. Add 'adios2_blosc_clevel' in &namelist_adios2 for adjusting blosc compression level
5. 'IO wrfbdy_d01 declared twice' exception thrown when running wrf.exe using 'io_form_boundary = 14'
6. Error/Warning messages do not state which adios2 function is causing the problem
7. '-ladios2_fortran_mpi' and/or '-ladios2_fortran' library not found when ADIOS2 having both /lib and /lib64

Solution:
1. Use 'adios2_inquire_operator' to check and reuse the existing 'Compressor' operator. Note: ADIOS2 operator needs to have an unique name under the same ADIOS2 IO.
2. Remove 'if (DH%compress_operator%name .ne. 'Compressor') then' condition which accessing the name before the operator is defined in subsequent lines. Note: the fix mentioned in #2167 was not taken account of this issue.
3. Remove the hard coded check. When the compressor is not supported, let the exception from ADIOS2 propagates up and return with a more meaningful error message.
4. Add 'adios2_blosc_clevel' to Register/registry.io_boilerplate. Use 'adios2_set_operation_parameter' to add this feature to the compress operator. Update doc/README.adios2.
5. Find and use a vacant IO name by iterating 'filename_<n>' and provide a heads-up to user. Note: ADIOS2 IO needs to have an unique name under the same ADIOS2 program-wide interface.
6. Edit error/warning messages making them more meaningful.
7. Edit arch/Config.pl to make it use /lib64, where the shared libraries are reside, if the directory exists before use /lib

ISSUE: Fixes #2167 and more

LIST OF MODIFIED FILES:
M Registry/registry.io_boilerplate
M arch/Config.pl
M doc/README.adios2
M external/io_adios2/wrf_io.F90

TESTS CONDUCTED:
1. Similar tests as in #2167 were conducted using Intel classic (ifort), Intel oneAPI (ifx) and Cray Fortran (ftn) compilers (using DM+SM option) on a HPE Cray EX system.
2. Additional tests, where ADIOS2 is also used for wrfbdy and wrfinput to run real.exe and wrf.exe, were conducted. The program run and complete successfully but with some warnings from ADIOS2 as mentioned in README.adios2.
3. The restart run  also works properly.

RELEASE NOTE:
Upgrade and bug fixes in ADIOS2 (io_form=14)

### DIFF
--- a/arch/Config.pl
+++ b/arch/Config.pl
@@ -732,13 +732,13 @@ while ( <CONFIGURE_DEFAULTS> )
     if ( $sw_adios2_path ) 
       { $_ =~ s/CONFIGURE_WRFIO_ADIOS2/wrfio_adios2/g ; 
         $_ =~ s:CONFIGURE_ADIOS2_FLAG:-DADIOS2: ;
-        if ( -d "$sw_adios2_path/lib" )
-          {
-            $adios2_libdir = "$sw_adios2_path/lib" ;
-          }
-        elsif ( -d "$sw_adios2_path/lib64" )
+        if ( -d "$sw_adios2_path/lib64" )
           {
             $adios2_libdir = "$sw_adios2_path/lib64" ;
+          }
+        elsif ( -d "$sw_adios2_path/lib" )
+          {
+            $adios2_libdir = "$sw_adios2_path/lib" ;
           }
         else
           {
@@ -1120,13 +1120,13 @@ while ( <ARCH_PREAMBLE> )
     if ( $sw_adios2_path ) 
       { $_ =~ s/CONFIGURE_WRFIO_ADIOS2/wrfio_adios2/g ; 
         $_ =~ s:CONFIGURE_ADIOS2_FLAG:-DADIOS2: ;
-        if ( -d "$sw_adios2_path/lib" )
-          {
-            $adios2_libdir = "$sw_adios2_path/lib" ;
-          }
-        elsif ( -d "$sw_adios2_path/lib64" )
+        if ( -d "$sw_adios2_path/lib64" )
           {
             $adios2_libdir = "$sw_adios2_path/lib64" ;
+          }
+        elsif ( -d "$sw_adios2_path/lib" )
+          {
+            $adios2_libdir = "$sw_adios2_path/lib" ;
           }
         else
           {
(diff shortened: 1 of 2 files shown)


SKanoksi:
I don't know why the WRF regression test always fails even when only the doc/README.adios2 is changed, since #2174. Furthermore, I never got the emails even after following the [guideline](https://github.com/wrf-model/WRF/wiki/How-to-Get-Jenkins-Emails).

I give up, since there are other IO options and the community mostly uses NetCDF anyway. 

For anyone wanting to try ADIOS2 (IO=14) with data compression for nest domain and others, either visit #2167 or my [fork](https://github.com/SKanoksi/WRF/tree/Upgrade-and-bug-fixes-in-ADIOS2-(io_form%3D14)), then edit the above mentioned files manually. Best regards.

PR #2181, opened 2025-03-11
Move netcdfpar build to postamble

islas:
TYPE: bug fix

KEYWORDS: netcdfpar, configure, make

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
Control of whether the external/io_netcdfpar folder is controlled via the presence of `NETCDFPAR_BUILD = CONFIGURE_NETCDFPAR_BUILD` which is replaced by a substitution in arch/Config.pl. However, currently this relies on all stanzas to include this value verbatim, otherwise that stanza will be unable to skip building the netCDF parallel IO code.    

Solution:
Since this is required regardless of stanza selected this should be isolated from the arch/configure.defaults and relocated to a common location for all stanzas to automatically use. Following current substitution variable layout, this is placed in arch/postamble.

TESTS CONDUCTED: 
1. Tested to ensure CMake and make builds of common stanzas are unaffected.

RELEASE NOTE: 
Fix uncontrollable building of external/io_netcdfpar folder for all stanzas


islas:
See forum post https://forum.mmm.ucar.edu/threads/netcdfparpath.20997/ for more detail on originating issue.

mgduda (approved):

smileMchen (approved):

PR #2185, opened 2025-03-26
Removed unnecessary factor in computation of dry air density in modul…

saneku:
Removed unnecessary factor in computation of dry air density in module_chem_utilities.F

TYPE: bug fix

KEYWORDS: WRF-Chem, dry air density

SOURCE: NOAA GSL, Alexander Ukhov (KAUST)

DESCRIPTION OF CHANGES:
Problem:
It was found that dry air density was miscalculated.

Solution:
Removed unnecessary factor. Simulations before and after did not show any significant difference, as expected.

before_and_after

LIST OF MODIFIED FILES:
M chem/module_chem_utilities.F

TESTS CONDUCTED: 

RELEASE NOTE: Fixed calculation of dry air density in module_chem_utilities.F. The bug had a very minor effect.

### DIFF
--- a/chem/module_chem_utilities.F
+++ b/chem/module_chem_utilities.F
@@ -78,7 +78,7 @@ CONTAINS
 
       p_phy(i,k,j) = p(i,k,j) + pb(i,k,j)
       t_phy(i,k,j) = (t(i,k,j)+t0)*(p_phy(i,k,j)/p1000mb)**rcp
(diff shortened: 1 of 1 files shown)


jordanschnell (approved):
Thanks for this bug fix!

weiwangncar (approved):

PR #2190, opened 2025-04-01
CMake README documentation on <PackageName>_ROOT variables

islas:
TYPE: text only

KEYWORDS: cmake, documentation

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
There has been some confusion on what variables are available within the CMake build to control where dependencies can be found. This has been noted a few times within the forum where users want fine tuned control on the location of these dependencies. The current build support this through native CMake support, however the current documentation left it as an exercise to the reader to construct the variables necessary for these features.

Solution:
Add explicit documentation of dependency search variables and how a user would make use of them. Additionally point to the exact external documentation to reference for the CMake `find_package()` call

### DIFF
--- a/doc/README.cmake_build
+++ b/doc/README.cmake_build
@@ -152,6 +152,35 @@ Advanced Configuration
   WRF_CASE option. Note that the value used is the actual name of the value, not
   the numeric shorthand used during interactive dialog.
 
+  Note: When using the cmake `*_ROOT` package variables to control where certain
+        depenendencies are found, please refer to the top-level CMakeLists.txt 
+        (i.e. <your WRF dir>/CMakeLists.txt) for all instances of `find_package()`
+        for the latest accurate list of all possible values. The construction of the
+        respective `*_ROOT` variable for a package uses the name in the call to
+        `find_package( <PackageName> )` verbatim and *is case sensitive* resulting
+        in `<PackageName>_ROOT`.
+        
+        The current list of possible `*_ROOT` is as follows :
+          Dependecies always used :
+            * netCDF_ROOT
+            * netCDF-Fortran_ROOT
+            * ZLIB_ROOT
+          
+          Conditional depenendencies based on configuration:
+            * MPI_ROOT
+            * OpenMP_ROOT
+            * HDF5_ROOT
+            * Jasper_ROOT
+            * RPC_ROOT
+            * BISON_ROOT
+            * FLEX_ROOT
+          
+        One can make use of passing in `CMAKE_PREFIX_PATH` to provide a lower priority
+        general search location if multiple depenendencies reside in the same directory.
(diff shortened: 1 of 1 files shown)


doc/README.cmake_build
```diff
@@ -152,6 +152,35 @@ Advanced Configuration
...
+            * RPC_ROOT
+            * BISON_ROOT
+            * FLEX_ROOT
+          
+        One can make use of passing in `CMAKE_PREFIX_PATH` to provide a lower priority
+        general search location if multiple depenendencies reside in the same directory.
+        Please refer to https://cmake.org/cmake/help/v3.20/command/find_package.html
+        for further documentation on using `*_ROOT` variables or other control methods
+        of resolving depenendency locations. Please note the version specification
+        as WRF uses the logic of CMake v3.20 even if you have a newer version of cmake installed.
```
mgduda, line 183: I've read it a few times, and I'm still having trouble understanding what this last sentence means. Is this referring to something that a person should do when specifying, e.g., `netCDF_ROOT`?

islas: This portion right?
> Please note the version specification
        as WRF uses the logic of CMake v3.20 even if you have a newer version of cmake installed.

This is in reference to the version specified in the URL, and that the logic used will follow that particular version regardless of the user's current CMake version. 

Perhaps a clearer phasing would be:
>Please note the version specification of the CMake documentation as seen in the URL. WRF uses the minimum CMake version v3.20, which is followed as close as possible for subsequent built-in calls. This means even if a user has a newer version of cmake installed, the package finding logic will [more or less] use what was outlined in CMake v3.20.

This is to address discrepancies with newer documentation and policies as setting a version requirement in the project implicitly changes policies to match as close as possible to that lowest version. https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html#policy-settings

mgduda: The alternative you suggest helps make it clear that the note about versions is in reference to the version of the documentation (and its agreement with the required CMake version for WRF). Just as another option to consider, how about something like this:
```
Note that the minimum require CMake version for WRF is specified as 3.20, and
so the package finding logic for WRF will follow as closely as possible the
sematics of find_package in verson 3.20, even if a newer version of CMake is
being used to build WRF.
```
?

islas: I like that! Though, I think `behavior` instead of `semantics` carries the point across better.
So:
>Note that the minimum require CMake version for WRF is specified as v3.20, and so the package finding logic for WRF will follow as closely as possible the behavior of find_package() in version v3.20, even if a newer version of CMake is being used to build WRF.

mgduda: Sounds good to me!

mgduda (approved):

weiwangncar (approved):

PR #2192, opened 2025-04-01
Fix aarch64 GCC build when DM configuration selected

islas:
TYPE: bug fix

KEYWORDS: make, compilation, dm

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
When a DM (distributed memory) configuration is selected, the configuration phase of the make build will extract the corresponding stanza from `arch/configure.defaults` and modify various variable lines with substitutions or comment removals.

Of particular note is the normally defined `DMPARALLEL = # 1` which is changed to `DMPARALLEL = 1` when DM configurations are selected. This definition  is then used to inform the make build about which pieces of code to compile for DM capabilities. The `Linux aarch64, GCC` stanza does not have this comment (`# 1`) and so the variable ends up blank when DM is selected causing compilation failure.

This issue possibly also exists for the `Linux aarch64, armclang` stanza, however no bug report has been noted for this at the time of this PR. 

Solution:
Add `# 1` to the `DMPARALLEL =` line in the `Linux aarch64, GCC` stanza definition. Likewise, perform the same change in the `Linux aarch64, armclang` stanza definition as well.

TESTS CONDUCTED: 
1. Confirmed with user on forum that this resolved their build issues for the `Linux aarch64, GCC` stanza.

### DIFF
--- a/arch/configure.defaults
+++ b/arch/configure.defaults
@@ -2174,7 +2174,7 @@ NETCDFPAR_BUILD	=      CONFIGURE_NETCDFPAR_BUILD
 #ARCH    Linux aarch64, armclang compiler OpenMPI # serial smpar dmpar dm+sm
 #
 DESCRIPTION     =      armclang ($SFC/$SCC): Aarch64
-DMPARALLEL      =
+DMPARALLEL      =      # 1
 OMPCPP          =      -fopenmp
 OMP             =      -fopenmp
 OMPCC           =      -fopenmp
--- a/arch/configure.defaults
+++ b/arch/configure.defaults
@@ -2214,7 +2214,7 @@ CC_TOOLS        =      $(SCC) -Wno-implicit-function-declaration -Wno-int-conver
 #ARCH    Linux aarch64, GCC compiler OpenMPI # serial smpar dmpar dm+sm
 #
 DESCRIPTION     =      GCC ($SFC/$SCC): Aarch64
-DMPARALLEL      =
+DMPARALLEL      =      # 1
 OMPCPP          =      -fopenmp
 OMP             =      -fopenmp
 OMPCC           =      -fopenmp


islas:
See partial thread https://forum.mmm.ucar.edu/threads/netcdfparpath.20997/post-51053 for more info on originating issue

arch/configure.defaults
```diff
@@ -2214,7 +2214,7 @@ CC_TOOLS        =      $(SCC) -Wno-implicit-function-declaration -Wno-int-conver
 #ARCH    Linux aarch64, GCC compiler OpenMPI # serial smpar dmpar dm+sm
 #
 DESCRIPTION     =      GCC ($SFC/$SCC): Aarch64
-DMPARALLEL      =
+DMPARALLEL      =      # 1
```
mgduda, line 2217: Though it would be a preemptive fix that we can't easily test ourselves, should a similar change also be made on line 2177?

islas: I did see that. I was a bit hesitant to make that change without testing to confirm a fix.

mgduda: I've been going back and forth on this. If we make a change without testing it, there's a risk that we actually break something; on the other hand, if we don't fix what looks like an obvious error, sooner or later it will impact someone. In my estimation, the likelihood is greater that changing line 2177 will actually fix a bug rather than introduce one, and so I'd support the application of the fix for line 2217 to 2177 also.

islas: Sounds good. I'll go ahead and include that change and note it in the PR description as well.

mgduda: Thanks!

mgduda (approved):

weiwangncar (approved):

PR #2195, opened 2025-04-09
NSSL-mp bug fix for (obsolete) droplet nucleation

MicroTed:
Quick bug fix for NSSL microphysics for old default droplet nucleation option

TYPE: bug fix

KEYWORDS: NSSL microphysics scheme

SOURCE: Ted Mansell (NOAA/NSSL) (@MicroTed )

DESCRIPTION OF CHANGES:
Problem:
Other updates resulted in the old droplet nucleation scheme (irenuc=2) producing too few cloud droplets. This option is no longer the default, but ought to work the same as it did before in case anybody wanted to use it.

Solution:
Restored use of background value for this particular option.

ISSUE: N/A

LIST OF MODIFIED FILES: 
phys/module_mp_nssl_2mom.F

TESTS CONDUCTED: 
2. Are the Jenkins tests all passing? TBD

RELEASE NOTE: Nothing to add to main NSSL update

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -124,7 +124,7 @@ wrf : framework_only
 	   echo "NoahMP submodule files populating WRF directories" ; \
 	   echo "------------------------------------------------------------------------------" ; \
 	fi
-	if [ \( ! -f phys/module_bl_mynnedmf.F \) -o \
+	@if [ \( ! -f phys/module_bl_mynnedmf.F \) -o \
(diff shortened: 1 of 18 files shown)


dudhia:
Should be safe. I will wait for the checks to pass.

dudhia (approved):
only affects non-default option

MicroTed:
Thanks!

PR #2199, opened 2025-04-14
Remove inaccurate error message regarding ghg_input

kkeene44:
TYPE: bug fix

KEYWORDS: ghg_input, idealized, no radiation

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
Most idealized cases have radiation options turned off, which causes the following error message to inaccurately print to the rsl files when ideal.exe is run:

ERROR: ghg_input available only for these radiation schemes: CAM, RRTM, RRTMG, RRTMG_fast
           OK = CAM LW with CAM SW

Solution:
In share/module_check_a_mundo.F, add code to ensure that when radiation is turned off, not only does the message not print, but the namelist value for ghg_input is set to 0, as well. 

LIST OF MODIFIED FILES: 

TESTS CONDUCTED: 
1. Code compiles okay and no longer prints the message for idealized cases (tested with em_quarter_ss).

RELEASE NOTE: This PR fixes an erroneous print for using ghg_input when no radiation option is selected, mostly from idealized cases.

### DIFF
--- a/share/module_check_a_mundo.F
+++ b/share/module_check_a_mundo.F
@@ -984,6 +984,8 @@
                    ( model_config_rec % ra_sw_physics(i) .EQ. RRTMG_SWSCHEME       )   .OR.  &
                    ( model_config_rec % ra_sw_physics(i) .EQ. RRTMG_SWSCHEME_FAST  ) ) ) ) THEN
                ! This is OK, no way would a negation have been understandable!
+            ELSE IF ( model_config_rec % ra_sw_physics(i) == 0 ) THEN
+                    model_config_rec%ghg_input=0
             ELSE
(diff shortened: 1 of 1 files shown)


weiwangncar:
The regression test results:

    Failed Simulations are: 
    None
    None

weiwangncar:
@kkeene44 Thanks. I assume you've tested this, and it behaves as expected.

weiwangncar:
@smileMchen Can you help review this PR? Thanks!

kkeene44:
> @kkeene44 Thanks. Please see if the added release note is ok with you.

I'm fine with that!

smileMchen (approved):

PR #2221, opened 2025-05-05
Add dependent files to custom command call to enable recompilation timestamp check

islas:
TYPE: bug fix

KEYWORDS: cmake, compilation

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
With the new CMake build, if a file that gets manually preprocessed prior to compilation is edited it will not trigger preprocessing again and thus skip recompilation for those edits.

Solution:
Add the original source file to the list of `DEPENDS` files when adding the custom command. This will add the appropriate make checks to compare timestamps between the original source file and the output results, triggering recompilation.

TESTS CONDUCTED: 
1. Tested recompilation of `wrf_io.F90` as a file that undergoes C and M4 preprocessing.

### DIFF
--- a/cmake/c_preproc.cmake
+++ b/cmake/c_preproc.cmake
@@ -29,6 +29,7 @@ macro( wrf_c_preproc_fortran )
   # Generate compile command and file outputs
   set( WRF_PP_F_OUTPUT   )
   set( WRF_PP_F_COMMANDS )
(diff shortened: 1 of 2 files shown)


weiwangncar:
The regression test results:

    Failed Simulations are: 
    None
    None

mgduda (approved):

Issue #2216, opened 2025-04-30
WRFCHEM fails to build with Intel OneAPI LLVM using old configure file


HathewayWill:
**Describe the bug**
using the old compiler method and intel llvm I attempted to build WRFCHEM w/KPP enabled.  It works on 4.6.1 so it should work with 4.7.0 without any issues.  However, there are some modules that are not linking correctly that is causing the model to fail.

**To Reproduce**
choose option 78 and basic nesting 1 after downloading 4.7.0.tar.gz
compile once and then attempt to compile a second time to fix the linkers based on past reccomendations

**Attachments**

**Additional context**
Add any other context about the problem here, such as:
used to work on 4.6.1 without a problem.

Interesting enough WRFCHEMDA built with intel llvm 2025.1.0 without a problem.
log file for wrfchemda: 
[wrfchemda_sucessful.zip](https://github.com/user-attachments/files/19971663/wrfchemda_sucessful.zip)

log file for wrfchem  without kpp (sucessful):
[wrfchem_no_kpp_success.zip](https://github.com/user-attachments/files/19971863/wrfchem_no_kpp_success.zip)

log files for wrchem with kpp (fail)
[wrf chem failed.zip](https://github.com/user-attachments/files/19969837/wrf.chem.failed.zip)

@mgduda @islas any thoughts?

HathewayWill:
`cb05_sorg_aq_jacobiansp_mp_lu_crow_'

seems to be an undefined reference to that ^

islas:
Please try #2222 to see if that fixes the problem. I found that this is not limited to the Intel compilers though it is beyond me how this passed testing...

HathewayWill:
@islas So the good news is #2222 seems to work where it doesn't fail  at the same spot but for some reason I think it is stuck in an infinite loop

![Image](https://github.com/user-attachments/assets/d0ebd507-c114-45a3-9004-b802d6b70003)

I can't upload the log file because it is over 24gb

I used these commands:

	source /opt/intel/oneapi/setvars.sh --force


```
	cd "${WRF_FOLDER}"
	git clone https://github.com/wrf-model/WRF.git

	cd WRF
	git submodule update --init --recursive
	gh pr checkout 2222
```

another thing that is odd is that even though I used 

`./compile -j 4 em_real 2>&1 | tee compile.wrf1.log`
my computer used every available core

[...]

PR #2222, opened 2025-05-05
Fix in KPP compilation during integration edit to use correct make variable

islas:
TYPE: bug fix

KEYWORDS: make, compilation

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
PR #2018 introduced CMake compilation of the chem and KPP code. In order to do that certain helper programs created during the compilation process were reworked. The KPP mechanism integration edit program, `integr_edit.exe` was changed to take in explicit arguments for the files to read and write. However, the edits to the makefile that control this call use an invalid variable `$(MODEL)` where `$(MECH)` should be used instead. 

Solution:
Use the appropriate `$(MECH)` variable to construct command line arguments to `intergr_edit.exe`

ISSUE: 
#2216

TESTS CONDUCTED: 
1. Recreated the issues noted in #2216, though not limited to the intel compilers. Diagnosed as segfault due to reading of nonexistent file from null expansion of bad make variable. Fixed and recompiled successfully.

### DIFF
--- a/chem/KPP/util/write_decomp/Makefile
+++ b/chem/KPP/util/write_decomp/Makefile
@@ -28,7 +28,7 @@ all:
 	$(MAKE) comp
 	./write_decomp.exe
 	$(MAKE) integr_edit
-	./integr_edit.exe $(MECH) module_kpp_$(MECH)_Integr.F decomp_$(MODEL).inc $(MODEL)_new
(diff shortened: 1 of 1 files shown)


weiwangncar:
The regression test results:

    Failed Simulations are: 
    None
    None

weiwangncar (approved):

Issue #2225, opened 2025-05-13
wrf.exe aborts after upgrading to WRF v4.7.0.


nalssi89:
When running wrf.exe, the simulation aborts with a fatal MPI error. This error was not present in WRF v4.6.1 but started occurring after upgrading to WRF v4.7.0.

The error message indicates a failure in PMPI_Type_match_size due to an "Invalid argument." Specifically, it states that "No MPI datatype available for typeclass MPI_TYPECLASS_REAL and size 1."

This change in behavior with WRF v4.7.0 suggests that modifications introduced in this version, particularly those utilizing MPI_Type_match_size in MPI communication routines (such as in frame/collect_on_comm.c, potentially related to PR #2157 which was included in v4.7.0 to improve handling of large data transfers), are now encountering a new situation. The error implies these updated routines are attempting to find an MPI data type for a real number that is 1 byte in size.

Standard MPI libraries do not typically define a 1-byte real data type. The core problem appears to be that the typesize argument, passed to these C communication routines from the WRF Fortran code, is unexpectedly set to 1 during the execution of wrf.exe in WRF v4.7.0.

Error Log:

Abort(604644108) on node 47 (rank 47 in comm 0): Fatal error in PMPI_Type_match_size: Invalid argument, error stack:
PMPI_Type_match_size(199): MPI_Type_match_size(typeclass=1, size=1, datatype=0x7ffea7915780) failed
PMPI_Type_match_size(178): No MPI datatype available for typeclass MPI_TYPECLASS_REAL and size 1

nalssi89:
WRF v4.7.0 aborts with PMPI_Type_match_size when using quilt I/O (nio_tasks_per_group=4, nio_groups=8).

At domain edges, patches shrink to a single grid point after halo trimming (size=1 byte), and MPI lacks a 1-byte REAL type.

Disabling quilting (nio_tasks_per_group=0, nio_groups=1) allows the run to complete successfully.

PR #2231, opened 2025-05-30
Use a temp pass through errhandler for trivial errors causing abort

islas:
TYPE: bug fix

KEYWORDS: mpi, quilting, comm

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
PR #2157 added changes to match an appropriate `MPI_Datatype` to a specific `typesize` during `col_on_comm()` and `dst_on_comm()`. This relies on `MPI_Type_match_size()` to query MPI about the equivalent MPI definition for a particular datatype size. There are safety checks to query `MPI_TYPECLASS_INTEGER` if `MPI_TYPECLASS_REAL` fails. 

However, when given a datatype size that does not match a possible `MPI_TYPECLASS_REAL` value (e.g. 1 byte where no real exists for single byte) instead of getting a failure via return code the query is treated as a critical failure and fully aborts the program. As the query does not rely on critical process handling _and_ since there already exists adequate checks to abort if no sufficient value is found, this preemptive abort is unnecessary. 

Solution:
Temporarily install a pass through errhandler that does not modify the return code but also does not abort. Allow the if statements of finding a correct `MPI_Datatype` to abort if deemed necessary. Additionally, once the checks are complete, reinstate any previous errhandler and free our pass through handle.

ISSUE: 
#2225 

TESTS CONDUCTED: 
1. Tested the stability of this call to handle correct and incorrect types various times while constantly replacing the error handler.

RELEASE NOTE: 
In collect_on_comm.c, use a temporary pass through errhandler to allow MPI_Type_match_size to fail correctly with error code rather than fully abort the program.

### DIFF
--- a/frame/collect_on_comm.c
+++ b/frame/collect_on_comm.c
@@ -37,6 +37,47 @@
 #endif
 
 
+#if defined( DM_PARALLEL ) && ! defined( STUBMPI )
+void
+temp_errhandler( MPI_Comm *comm_ptr, int *errcode, ... )
+{
+  // do nothing...
+}
+
+MPI_Datatype
+get_datatype_from_typesize( int *typesize )
+{
+  MPI_Errhandler errhandler;
+  MPI_Errhandler previous;
+  MPI_Datatype dtype;
+  int ierr = -1;
+
+  // MPI critically aborts on trivial errors caused by this call so replace error
+  // handler temporarily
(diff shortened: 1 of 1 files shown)


weiwangncar:
I passed the routine to another user who has problem with io_quilting on Derecho and it worked for him!

weiwangncar (approved):

PR #2249, opened 2025-09-24
Fix floating point exception in phys/module_mp_radar.F when debugging with -D

### LINKED ISSUE #2248: Floating point exception in phys/module_mp_radar.F with debugging option -D
**Describe the bug**
Line 225 in phys/module_mp_radar.F (`xoamh = 1./xam_h`) triggers a floating point exception when debugging with -D and using any microphysics scheme which calls `radar_init` except the UDM scheme.

It seems this is because `xam_h` is only given a value in the UDM scheme, so any other scheme which calls `radar_init` will naturally lead to a floating point exception.

It's not a major issue because the variable `xam_h` is only used by the UDM scheme in which it is initialised, but it means you cannot debug with option -D when using some microphysics schemes.

**To Reproduce**
1. Do `./configure -D`, then compile
2. Try to run with a microphysics scheme except UDM that calls `radar_init`, e.g. Thompson

This is my first time raising an issue, so I hope I'm doing it right!

jackbartlett20:
Prevent floating point exception in phys/module_mp_radar.F when debugging with -D related to new variables in UDM scheme.

TYPE: bug fix

KEYWORDS: debug, radar, microphysics, UDM, -D

SOURCE: Jack Bartlett (Imperial College London)

DESCRIPTION OF CHANGES:
Problem:
Lines 225 and 226 in phys/module_mp_radar.F trigger a floating point exception when debugging with option -D and using any microphysics scheme which calls `radar_init` except the UDM scheme because `xam_h` and `xbm_h` are not initialised in any other microphysics scheme.

Solution:
Initialise `xam_h` and `xbm_h` at 0 and check if they have been given a value greater than 0 before finding their reciprocals.

Fixes #2248

phys/module_mp_radar.F

TESTS CONDUCTED: 
Only tested by trying `mp_physics=8` with debugging option -D and the new fixes (which previously triggered a fpe).

### DIFF
--- a/phys/module_mp_radar.F
+++ b/phys/module_mp_radar.F
@@ -49,8 +49,10 @@ MODULE module_mp_radar
       REAL, PUBLIC:: xam_r, xbm_r, xmu_r, xobmr
       REAL, PUBLIC:: xam_s, xbm_s, xmu_s, xoams, xobms, xocms
       REAL, PUBLIC:: xam_g, xbm_g, xmu_g, xoamg, xobmg, xocmg
-      REAL, PUBLIC:: xam_h, xbm_h, xmu_h, xoamh, xobmh, xocmh
+      REAL, PUBLIC:: xmu_h, xoamh, xobmh, xocmh
       REAL, PUBLIC:: xorg2, xosg2, xogg2, xohg2
+      REAL, PUBLIC:: xam_h = 0. ! Prevents fpe in debugging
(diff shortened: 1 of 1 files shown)


islas (commented):
Do the set of `xo*mg` variables using `xam_g` / `xbm_g` need to be protected in the same manner?

dudhia (approved):

PR #2275, opened 2026-01-15
Update GWDO and SHINHONG PBL

Songyou184:
TYPE: Enhancement  

KEYWORDS: Revised GWDO and SHINHONG PBL schemes, as in the updates in MPAS 

SOURCE: Songyou Hong (NCAR)

DESCRIPTION OF CHANGES:
- YSU GWDO was revied following the study of Hong et al. (2025, WAV), now is icalled as KIM GWDO
 KIM GWDO requires additional maximum orography file, "ELVMAX", in WPS
 Non-hydrostatic effect of Xu et al. (2024, JAS) was added (gwd_if_nonhyd=.true., default)
 Factor for the effective grid spacing was added (gwd_dx_factor=2.0, default)
 Tutbulent form drag (TOFD) was added in computing surface drag (kim_tofd=.true., default)
 Factor for TOFD was added (tofd_factor=.003, default)

- SHINHONG PBL was updated

Problem: SHINHONG PBL occasionally was crashed in global models

Solution: Entrainment ratio and depth  at the PBL top were reformulated. Topdown mixing, orographic drag, and 
building canopy moduels in the YSU PBL were revised and incorporated into revised the SHINHONG scheme. In the revised scheme, one can select nonlocal mixing flux term: either counter-gradient (YSU type) or eddy-mass flux (SHINHONG type). In the namelist,
 - shinhong_nonlocal_flux=.false., default for YSU nonlocal flux, .true. for SHINHONG nonlocal flux
 - shinhong_scu_mixng = .false., default, for stratocumulus downward mixing (revised TOPDOWN mixing in the   
                      YSU scheme
 - shinhong_ke_dissipation=.false., defalult, for additional haeting due to kinetic turbulent mixing
 - 
Remarks : PHYSICS_MMM is pointed to a personal site, https://github.com/Songyou184/WRF.git -b GWDO_SH

	modified:   arch/Externals.cfg
	modified:   main/depend.common
	modified:   phys/CMakeLists.txt
	modified:   phys/Makefile
	modified:   phys/module_bl_gwdo.F
	modified:   phys/module_bl_shinhong.F
	modified:   phys/module_pbl_driver.F
	modified:   phys/module_physics_init.F
	modified:   phys/module_sf_sfclayrev.F
	modified:   phys/module_surface_driver.F


Songyou184:
Evaluation of the two GWDO and PBL updates could be made in order.. 
WIth the default namelist file in WRF and MPAS (mmm suite)

1. Updated GWDO will be evaluated ==> Expect reduced winds in lower and upper troposphere and in its above. Surface wind in unstable conditions are reduecd due to a surface layer based form drag activation  (kim_tofd=.true., default), which is absent in the previous versions. Note that this GWDO update is not compatible with the previous WPS since additional mountain data, elvmax, is required in the new update.

2. WIth the SHINHONG PBL option (option = 11 in WRF), the difference of SHINHONG PBL over the YSU PBL will be evaluated. With the default SHINHONG option,  the result would be identical or very similar to that from YSU scheme with TOPDOWN mixing =0, CTOPO = off , for the grid spacing greatern than 5 km.  shinhong_nonlocal_flux=.false, default; results in the same nonlocal flux in the YSU scheme.

3. Additional issues are :
-- shinhong_scu_mixing=.false. default ; can be changed to .true. to follow the strato comulus mixing in the YSU PBL with TOPDOWN mixng = 1.
--  shinhong_nonlocal_flux=.false, default; can be chaned to .true. to follow the mass-flux type 
nonlocal mixing in the SHIN HONG scheme.

arch/Externals.cfg
```diff
@@ -1,8 +1,8 @@
 [MMM-physics]
-local_path = ./phys/physics_mmm
+local_path = ./physics_mmm
 protocol = git
 repo_url = https://github.com/NCAR/MMM-physics.git
-tag = 20240626-MPASv8.2
+branch = testing-branch
```
dudhia: @islas @Songyou184 Can this be converted to a submodule call as a test, even though this is not the final version for the PR?

islas: Yes, I can do that change.

dudhia:
Looks like this has not passed tests but I am not seeing the submodule implementation yet.

islas:
The submodule change is there (for me it shows as the last diff on the "Files changed" page) but since I merged `develop` it won't look like a new change since the submodule usage is already in. 

I think we'll have to wait for @weiwangncar to see the error log since I don't get emails for PRs that aren't mine.

dudhia:
OK, I saw that Externals.cfg was still in the Makefile, but maybe I am not seeing what was tested.

weiwangncar:
This PR has been replaced by three separate ones: 2284, 2285 and 2286.

PR #2269, opened 2026-01-09
Replace GF with GFL submodule for WRF4.8 release

haiqinli:
The Grell-Freitas (GF) cumulus convection is replaced with the Grell-Freitas-Li (GFL) submodule, which is consistent in WRF, MPAS and CCPP.

TYPE: new feature

KEYWORDS: submodule,  cumulus, parameterization, scale-aware, aerosol-aware

SOURCE: Haiqin Li (Haiqin.Li@noaa.gov, CIRES/CU Boulder at NOAA/GSL)

DESCRIPTION OF CHANGES: The Grell-Freitas (GF) cumulus convection is replaced with the Grell-Freitas-Li (GFL) submodule, which is consistent in WRF, MPAS and CCPP. It will be easy to access the GFL updates via the hash pointer of submodule.

LIST OF MODIFIED FILES: 
M     .gitmodules
M     Registry/Registry.EM_COMMON
M     dyn_em/module_big_step_utilities_em.F
M     dyn_em/module_convtrans_prep.F
M     dyn_em/module_em.F
M     phys/Makefile
M     phys/module_cumulus_driver.F
M     phys/module_lightning_driver.F
M     phys/module_physics_addtendc.F
M     phys/module_physics_init.F
M     share/module_check_a_mundo.F
M     wrftladj/module_cumulus_driver_ad.F
M     wrftladj/module_cumulus_driver_tl.F
D      module_cu_gf_ctrans.F
D      module_cu_gf_deep.F
D      module_cu_gf_sh.F
D      module_cu_gf_wrfdrv.F
M   chem/chem_driver.F
M    chem/KPP/compile_wkc

TESTS CONDUCTED: 
1. It was tested with a NWP case and a WRF_Chem case.
2. It passed regression tests.

RELEASE NOTE: The Grell–Freitas (GF) cumulus convection scheme has been further developed as the Grell–Freitas–Li (GFL) submodule, which is implemented consistently across WRF, MPAS, and CCPP. Major updates include improved aerosol–convection interactions, several simplified approaches for cold-pool parameterization, and a new scale-awareness feature with subsidence spreading.

References:
Li, H., G. Grell, and Freitas, S. (2026): Evauating the impact of aerosol interactions in a convective parameterization on medium range precipitation forecasts in the Unified Forecast System, to be submitted.

Grell, G., S., Freitas, and Li, H. (2026): A New Scale Aware Technique for Convective Parameterizations, to be submitted.

PR #2290, opened 2026-03-03
Updating MYNN-SFC

joeolson42:
The purpose for this PR is to update the refactored MYNN surface layer scheme, which has now been more fully debugged and has been tested in both WRF and MPAS. 

TYPE: enhancement

KEYWORDS: MYNN-SFC version 6.0.0 refactored HRRR

SOURCE: Joseph B. Olson

DESCRIPTION OF CHANGES:
The Main Problem:
This scheme was refactored starting from the CCPP version, which has a few distinctions from WRF/MPAS that required further work to fully integrate into both WRF and MPAS.

Solutions:

1. Corrected the conversions of mixing ratio/specific humidity.
2. added output of variables qgh and cqs, which are needed in some mode frameworks for some configurations
3. Some tuning was done for sf_mynn_sfcflux_land = 1 to work better when initializing off of RAP/HRRR/RRFS and using RUC LSM. If using Noah LSM, it will default to the Chen et al. (variable Czil) approach.
4. Added the infrastructure for bathymetry-dependent z0, which is expected to be compared with other approaches (possibly from WFIP3 data) and possibly tweaked in the near future.
5. Added automated CI testing in "MYNN-SFC/tests", which further help check for code errors for every PR to the MYNN-SFC repository. This helped find a few bugs in relatively unused configurations.

LIST OF MODIFIED FILES: list of changed files (use `git diff --name-status master` to get formatted list)
M       phys/MYNN-SFC
M       phys/module_surface_driver.F

TESTS CONDUCTED: 
1. Do mods fix problem? Yes, case study testing with debug flags active.
2. Are the Jenkins tests all passing?

PR #2297, opened 2026-03-05
Fix GFL build issues

islas:
TYPE: bugfix

KEYWORDS: build, make, cmake

SOURCE: internal

DESCRIPTION OF CHANGES:
Problem:
PR #2269 introduced the GFL submodule replacing the older GF implementation. In doing so, the related files' dependencies and additions to both builds was not done completely.

Solution:
Add object dependencies for make build. Remove old file reference and update with new files in cmake build.

### DIFF
--- a/main/depend.common
+++ b/main/depend.common
@@ -2569,6 +2569,21 @@ module_cu_gf_ctrans.o: \
 	../frame/module_state_description.o 
 
 
+module_cu_gfl_common.o: \
+	ccpp_kind_types.o
+
+
+module_cu_gfl_sh.o: \
+	module_cu_gfl_deep.o \
+	module_cu_gfl_common.o
+
+
+module_cu_gfl_wrf.o: \
+	module_cu_gfl_deep.o \
+	module_cu_gfl_sh.o \
+	ccpp_kind_types.o
+
+
 module_cumulus_driver.o: \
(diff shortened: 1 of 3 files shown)


dudhia:
There seems to be some gf_* removal that can also be done in depend.common.

weiwangncar (approved):

dudhia (dismissed):
Conditional on regression tests completing.

weiwangncar (dismissed):

islas:
@weiwangncar Do you know if either of these files triggers the Scala tests? It seems like it won't

weiwangncar:
@islas It doesn't look like the changes triggered a re-test. Let's force it for a test by adding 'Retest' label. I see that you added 'Retest' yesterday. But I didn't receive any output from the test.

islas:
@weiwangncar I did that yesterday, but no luck

weiwangncar:
@islas Maybe changes in depend.common is excluded from the test? But in the meantime, you may need to add a trivial change in one of the .F files to trigger a test.

dudhia (approved):

islas:
@weiwangncar Could you take a look at this? I'd prefer this go in before #2290

weiwangncar (approved):

PR #2349, opened 2026-06-07
Limit aerosol optics range warnings to active bins

saneku:
Limit aerosol optics range warnings to active bins

TYPE: bug fix

KEYWORDS: Mie, refractive index, diagnostics, aerosols

SOURCE: Alexander Ukhov (KAUST)

DESCRIPTION OF CHANGES:
Problem:
The aerosol optics Mie calculation clamps refractive indices to lookup-table bounds before interpolation. For empty
or nearly empty optical bins, fallback refractive indices such as (1.5,0.0) can trigger repeated range-warning
messages, especially when zero imaginary refractive index values are compared against the table's small negative
upper bound. This produces excessive log output without indicating a meaningful optical impact.

Solution:
The existing refractive-index clamping behavior is preserved. The diagnostic warning messages are now printed only when the particle number concentration in the optical bin is significant. For upper-bound imaginary-index warnings, diagnostics are also suppressed for effectively zero absorption using abs(refi).gt.1.e-12. This keeps the interpolation guard in place while avoiding warning spam from empty or trace aerosol bins. Warnings are still retained for active aerosol bins whose refractive indices fall outside the lookup-table range.

LIST OF MODIFIED FILES:
M	chem/module_optical_averaging.F

TESTS CONDUCTED:
Test simulations were conducted. The simulation results are unchanged, and the excessive refractive-index range warnings no longer appear in the model output.

RELEASE NOTE:
Suppresses repeated WRF-Chem aerosol optics refractive-index range warnings from empty or trace optical bins while
preserving existing Mie lookup-table clamping behavior.


weiwangncar:
The regression test results:

    Failed Simulations are: 
    None
    None

jordanschnell (dismissed):

PR #2369, opened 2026-07-27
Fix uninitialized ilnb read on the thin-snow path in the RUC LSM

### LINKED ISSUE #2368: RUC LSM: `ilnb` is read uninitialised in `SNOWTEMP` and `SNOWSEAICE` on thin snow packs, corrupting `TSNAV`
### Summary

In `phys/module_sf_ruclsm.F`, `ilnb` (number of snow layers) is read on a path where it is never assigned. The read selects between two formulas for `TSNAV`, and on that path the selected formula is degenerate, producing errors that grow without bound as the pack thins — 14 K at 3 cm of snow, 32 K at 1 cm, 60 K at 1 mm.

I do not believe this measurably affects forecast fields (see Impact below), but it is undefined behaviour in operational physics, it makes restarts non-reproducible, and the fix appears to be one line.

### The defect

The clearest statement is a standard-conformance one, which needs no assumption about compilers or stack reuse:

`SNOWTEMP` declares `ilnb` as **`intent(out)`** at `:4994`. Under the Fortran standard an `intent(out)` dummy argument is undefined on entry, regardless of what the caller holds. `SNOWTEMP` then assigns `ilnb` only inside `if(snhei.ge.snth)`:

```fortran
5118:       if(snhei.ge.snth) then
5124:          ilnb=1
      ...
5148:          ilnb=2
```

and reads it at `:5716`:

```fortran
5715:        if(snhei.gt.0.) then
5716:          if(ilnb.gt.1) then
```

On the thin-pack path — `if(snhei.lt.snth.and.snhei.gt.0.)` at `:5174` — neither assignment executes, but `snhei > 0.` is true, so the read happens.

There is a strong indication this is unintended rather than a deliberate formulation: **the thin block already computes the correct value at `:5191`**

```fortran
5191:          tsnav=0.5*(soilt+tso(1)) - 273.15
```

**and the `ilnb`-dependent block at `:5715-5722` then overwrites it.**

The caller side corroborates: `SFCTMP` declares `integer :: k,ilnb` at `:1385` and never assigns it, passing it to `SNOWSOIL` at `:1928` and `SNOWSEAICE` at `:1956`. `SNOWSOIL` (`intent(inout)`, `:3329`) never assigns it either — it only forwards it at `:3600`.

`SNOWSEAICE` has the identical structure at `:3896` (declaration), `:4035` (guard), `:4086` (thin block) and `:4410` (read). There `ilnb` is `intent(inout)`, so it is a stale-value bug rather than a formally undefined one.

Why the degenerate branch is wrong on this path: `deltsn = 5 * snth` by construction (`:3387-3388`), so with `snhei < snth` the two-layer weights `deltsn/snhei` and `(snhei-deltsn)/snhei` are an extrapolation across a negative layer thickness. The error is analytically `0.5*(deltsn/snhei - 1)*(soilt - tso(1))`, and it places the pack-mean temperature outside the two temperatures bounding the pack.

### Reproduction

No source edits required. Compile the unmodified module and drive `SFCTMP` with a single thin-snow column (`snhei = 0.02 m`, `rhosn = 300`, giving `snth = 0.0333 m`), argument list copied from the call site at `:944`. Build twice:

```
gfortran ... -finit-integer=1     ->  TSNAV = -5.72 C
gfortran ... -finit-integer=2     ->  TSNAV = -26.08 C
```

With no init flag, at `-O0`, the value carried in depends on the preceding column: a thin column preceded by a deep column gives -26.08, preceded by a medium column gives -5.72. Branch selection also varied with optimisation level (`-O1` selected the degenerate branch with no preceding snow column at all) and changed when unrelated code was added to the driver — consistent with the by-reference carry-through described above.

Tested with gfortran 13.3.0 against v4.6.1 at `d66e442`.

### Impact

I want to be careful not to overstate this.

`TSNAV` has exactly one read site in the model: the Koren compaction term at `:1495`. That path is gated at `:1493` requiring SWE > 8.1 mm, while the thin-pack path requires SWE < 10.0 mm, so corrupted values can only reach prognostic state in a ~1.9 mm SWE band — and the gate at `:1496` closes even that unless the timestep is long.

Integrating a column through the unmodified module, `SNOWH`, `SNWE`, `RHOSN`, `SOILT` and `TSO(1)` were **bit-identical** at dt = 20 s and dt = 60 s; divergence appeared only at dt = 180 s and above, and then at ~0.5 mm of SNOWH out of ~97 mm. `TSNAV` itself carries the restart-stream flag only in `Registry.EM_COMMON:1977`, so it is not in `wrfout` by default.

So the practical consequences look like: (1) a restart-stream field that is wrong whenever the pack is thin, and (2) **restart non-reproducibility**, since a run restarted mid-forecast may select a different branch than the continuous run. I could not confirm the timestep-dependence against an operational configuration, and a maintainer with access to one is better placed to check whether the compaction gate is genuinely closed there.

### Suggested fix

Add a default before the branch in each routine:

```fortran
! SNOWTEMP, immediately before the guard at :5118
      ilnb = 1

! SNOWSEAICE, immediately before the guard at :4035
      ilnb = 1
```

One layer is the physically correct choice for a sub-`snth` pack — the block's own comment says the snow "is too thin to be treated separately, therefore it is combined with the first soil layer" — and it reproduces exactly what `:5191` already computes before the overwrite. It changes nothing on the thick path, where the existing assignments still run.

(Since `SFCTMP` never references `ilnb` after the calls, it could also be demoted to a local in each callee, but that is cleanup rather than the fix.)

### Secondary observations

- The comments at `:3375-3379` and `:3939-3943` describe `snth` as using `snwe = 0.016 m` ("4 cm at 400 kg/m³") and `deltsn` as `0.03 m` ("7.5 cm"), but the code at `:3387-3388` uses 0.01 and 0.05, giving 2.5 cm and 12.5 cm. Off by 1.6x in the place a reader would go to reason about the layer scheme.
- `gfortran -O2 -Wall` on the unmodified module also flags `cotsn`, `rhtsn`, `d1sn` and `d2sn` as possibly-uninitialised in the same snow blocks. I have not investigated those and offer them only as a lead.

### Version information

- Present in v4.6.1, v4.7.0, v4.7.1, v4.8.0, `master` and `develop`.
- `phys/module_sf_ruclsm.F` is byte-identical across all of them (md5 `aa014adb857cf76d819c6951e51000ea`), last modified by #2028.
- The same code appears in `NCAR/ccpp-physics`. CCPP clamps both branches with `min(zero, ...)`, which does not mitigate this, since the degenerate branch typically errs cold.
- I found no existing issue mentioning `ilnb` or `tsnav` in this repo, `NCAR/ccpp-physics`, or `ufs-weather-model`.

FahrenheitResearch:
TYPE: bug fix

KEYWORDS: RUC LSM, snow, uninitialized variable, ilnb, TSNAV, thin snow, restart reproducibility

SOURCE: FahrenheitResearch (independent)

DESCRIPTION OF CHANGES:
Problem:
As described in #2368: in `SNOWTEMP` and `SNOWSEAICE`, `ilnb` is assigned only inside `if(snhei.ge.snth)`, but the TSNAV recalculation at the end of each routine (`:5716` / `:4410`) reads it whenever `snhei > 0.`. On the thin-snow path (`0 < snhei < snth`) nothing has assigned it — `ilnb` is `intent(out)` in `SNOWTEMP` (undefined on entry under the Fortran standard) and carries stale caller memory in `SNOWSEAICE` — and when the leftover value is greater than 1 the two-layer formula runs with `deltsn > snhei`, an extrapolation across a negative layer thickness whose error grows without bound as the pack thins (14 K at 3 cm of snow, 60 K at 1 mm). It also makes the branch selection, and therefore TSNAV in the restart stream, dependent on compiler, optimization level, and the preceding column, so restarted runs need not reproduce continuous ones.

On @weiwangncar's suggestion to set `ilnb` in `LSMRUC` before the call to `SFCTMP`: I tried that form first, but `ilnb` is a local variable of `SFCTMP` (declared under `!--- local variables` at `:1385`), not an argument passed in from `LSMRUC`, so there is nothing in `LSMRUC` to set. And because `SNOWTEMP` declares it `intent(out)`, an initialization upstream of the call chain would still be formally undefined on entry to `SNOWTEMP` — it would work with gfortran's by-reference scalar passing, but by the same accident that makes the current code usually work. Setting the default inside each routine that reads it is the smallest change that is well-defined on every path; happy to rework it into the caller-side form plus an intent change if the committee prefers.

Solution:
Default `ilnb=1` immediately before the layer-count branch in both `SNOWTEMP` and `SNOWSEAICE`. One layer is what the thin-snow block itself assumes (its comment: the snow "is too thin to be treated separately, therefore it is combined with the first soil layer"), and it reproduces exactly the TSNAV that block already computes (`:5191` / `:4103`) before the overwrite. Thick-path behavior is unchanged: both existing assignments (`ilnb=1` / `ilnb=2`) still execute and take precedence.

ISSUE: Fixes #2368

LIST OF MODIFIED FILES:
M       phys/module_sf_ruclsm.F

TESTS CONDUCTED:
1. Yes. A standalone harness compiles the module from this branch exactly as shipped (gfortran 13.3.0, `-O0`) and drives `SFCTMP` with a 33-case, 160-column fixture covering land and sea ice, mosaic and non-mosaic, thick, thin, melting, and snow-free packs:
   - **Unpatched, driver stack pre-filled 0.0 vs 12345.0:** outputs differ in exactly 45 cells, all of them TSNAV, all on thin-pack mosaic cases (one case reports a pack-mean temperature of +20.12 C against the correct -0.875 C) — the uninitialized read firing.
   - **Unpatched, `-finit-integer=1` vs `-finit-integer=2`:** the same 45 TSNAV cells differ — the reproduction from the issue, confirmed on `release-v4.8.0`.
   - **Patched:** both controls go quiet — 0 differing cells under stack pre-fill, 0 under `-finit-integer=1/2`, so the output no longer depends on what memory or the compiler left in `ilnb`.
   - **Patched vs unpatched with a zeroed stack** (where `ilnb` reads 0 and already selects the one-layer formula): bit-identical across all 33 cases — no output changes on any thick-snow or snow-free path.
2. Jenkins tests have passed.

RELEASE NOTE: Fixed an uninitialized read of the snow-layer count in the RUC LSM: on thin snow packs the pack-average snow temperature TSNAV could be computed with the two-layer formula, whose weights are degenerate there, producing errors that grow without bound as the pack thins and making restarts non-reproducible. The layer count now defaults to the single-layer form that the thin-snow path itself computes.

### DIFF
--- a/phys/module_sf_ruclsm.F
+++ b/phys/module_sf_ruclsm.F
@@ -4032,6 +4032,10 @@ print *, 'tso before calling snowtemp: ', tso
    33  continue
 !--- the nzs element in cotso and rhtso will be for snow
 !--- there will be 2 layers in snow if it is deeper than deltsn+snth
+!--- default to one snow layer: on the thin-snow path (snhei < snth)
+!--- neither branch below is taken, and ilnb is read for the tsnav
+!--- recalculation at the end of this routine
+       ilnb=1
        if(snhei.ge.snth) then
         if(snhei.le.deltsn+snth) then
 !-- 1-layer snow model
@@ -5115,6 +5119,10 @@ print *, 'snowtemp: snhei,snth,soilt1: ',snhei,snth,soilt1,soilt
    33  continue
 !--- the nzs element in cotso and rhtso will be for snow
 !--- there will be 2 layers in snow if it is deeper than deltsn+snth
+!--- default to one snow layer: on the thin-snow path (snhei < snth)
+!--- neither branch below is taken, and ilnb is read for the tsnav
+!--- recalculation at the end of this routine
+       ilnb=1
        if(snhei.ge.snth) then
         if(snhei.le.deltsn+snth) then
 !-- 1-layer snow model


weiwangncar:
The regression test results:

    Failed Simulations are: 
    None
    None

dudhia (approved):

weiwangncar (approved):

PR #2353, opened 2026-06-14
Fixed the volc diags bounds crash

saneku:
Fixed volcanic diagnostics bounds crash and minor volcanic transport fixes.

TYPE: bug fix

KEYWORDS: volcanic run, chem_opt=402, volcanic diagnostics, wet scavenging, emissions

SOURCE: Alexander Ukhov (KAUST)

DESCRIPTION OF CHANGES:

Problem:
Several volcanic diagnostic explicit interfaces declared volc_diags with a hard-coded third dimension of 1:9. The Registry-generated volcanic diagnostics package is sized by num_volc_diags, and p_ash_fall resolves to index 10. Debug builds with bounds checking therefore crashed in module_vash_settling when writing volc_diags(i,j,p_ash_fall). Optimized builds did not expose the bounds violation consistently.

Solution:
Replaced hard-coded volc_diags(...,1:9) declarations with volc_diags(...,num_volc_diags), and imported num_volc_diags where needed.

Additional minor fixes:
- Updated volcanic emission cell-area calculation to use map factors per active eruption grid cell.
- Ensured convective wet scavenging density is initialized before use in module_ctrans_grell.

LIST OF MODIFIED FILES:
M chem/dry_dep_driver.F
M chem/emissions_driver.F
M chem/mechanism_driver.F
M chem/module_ctrans_grell.F
M chem/module_vash_settling.F
M chem/module_volc_chem.F

TESTS CONDUCTED:
- Ran volcanic WRF-Chem simulation with chem_opt=402 using a debug build with bounds checking. The previous p_ash_fall / volc_diags bounds crash no longer occurs.
- Ran volcanic WRF-Chem simulation with chem_opt=402 using a non-debug optimized build. The simulation also runs successfully.

RELEASE NOTE:
Fixed a volcanic diagnostics bounds crash for chem_opt=402 and included minor volcanic emissions / scavenging
fixes.

PR #2387, opened 2026-08-28
chem: add scalar guards for p_hcl and p_ch3cl in add_anthropogenics t…

AlphaBetaGammaChi:
Fixes a runtime SIGSEGV crash during `emissions_driver` when running chemical mechanisms that do not include anthropogenic chlorine (such as CRI-MECH with `emiss_opt = 19` or `emiss_opt = 20`).

Cause
In `chem/module_emissions_anthropogenics.F`, assignments for `p_hcl` and `p_ch3cl` were placed in the generic anthropogenic emissions loop without `param_first_scalar` bounds checks. 
For mechanisms where `e_hcl` and `e_ch3cl` are not present in the package, referencing `chem(..., p_hcl)` and `emis_ant(..., p_e_hcl)` causes an invalid memory reference and crashes on timestep 1:
```text
Program received signal SIGSEGV: Segmentation fault - invalid memory reference.
Backtrace:
__module_emissions_driver_MOD_emissions_driver
chem_driver_
solve_interface_

### CODE REVIEW

**Comment by weiwangncar** [comment]
The regression results:

    Failed Simulations are: 
    None
    None

File: chem/emissions_driver.F [hunk 1]
```diff
@@ -423,9 +423,8 @@ subroutine emissions_driver(id,ktau,dtstep,DX,DY,                      &
 
       ivolcano=0
       
-      !area=dx*dx
-      ! A. Ukhov 12 Oct 2024
-      area=(dx/mapfac_mx(i,j))*(dy/mapfac_my(i,j))
+      
+      area=dx*dx
```
dudhia, line 427: This reverts to an approximate grid size. Do we need this?

weiwangncar:
@dudhia Without this, the model can abort as it has been shown - using array indices outside do loops. If it needs to be that accurate, there are other ways to fix it.

saneku:
I introduced 

**area=(dx/mapfac_mx(i,j))*(dy/mapfac_my(i,j))**

And I could achievemass-balancee closure with it.

Using
**area=dx*dx**

does not provide mass-balance closure, as it is a less accurate area calculation.

weiwangncar:
@saneku Can you move the area calculation to a do loop where the accurate area information is needed? For example area can be defined inside the i,j do loop that begins on line 455, for example. Or you can declare a local 2D array to store the area array to be used later?

saneku:
I agree with @weiwangncar. The map-factor-aware area calculation should be preserved, but moved inside the i,j loops, where the indices are defined:

  ! For volcanic ash transport, vash variables are in mixing ratio here.
  do j = jts, jte
    do i = its, ite
      if (erup_end(i,j).gt.0) then
        area = (dx/mapfac_mx(i,j)) * (dy/mapfac_my(i,j))

        so2_mass = 1.5e4*3600.*1.e9/64./area

The area argument should then be removed from the preceding diagnostic message. Its format should also be changed from 2f15.3 to f15.3:
  write(message,'(" ADJUSTED ASH HEIGHT: ",f15.3)') emiss_ash_height

weiwangncar:
@AlphaBetaGammaChi Would you like to make this change in this PR?

saneku:
@weiwangncar and @AlphaBetaGammaChi i have noticed that this correction is already implemented in 
https://github.com/wrf-model/WRF/pull/2353

weiwangncar:
@saneku Thanks for the alert. We will review PR-2353 and close this one afterwards.

AlphaBetaGammaChi:
Hi @weiwangncar,

Thank you! Just to flag: PR #2353 addresses the `emissions_driver.F` area calculation, but the other half of this PR — the scalar guards for `p_hcl` and `p_ch3cl` in `chem/module_emissions_anthropogenics.F` — is a separate, independent bug fix that is not covered by #2353 I believe.

Without those guards, any CRI-MECH run (`chem_opt=601` with `emiss_opt=19` or `emiss_opt=20`) will still crash I believe with a SIGSEGV on timestep 1 due to unguarded array references for species not present in the mechanism.

Would it be possible to either:
1. Keep this PR open for the `module_emissions_anthropogenics.F` fix (I can revert the `emissions_driver.F` change so it doesn't conflict with #2353), or
2. Incorporate the scalar guards into #2353?

Happy to proceed either way!

weiwangncar:
@AlphaBetaGammaChi I suggest to go with your proposal #1. Thanks for the reminder for the other fix.

